PR #4, opened 2016-06-02
check for the existence of an imports folder

pbuttigieg:
prevents dir not found error


pbuttigieg:
Needs to be `git add`ed, not sure if that should be done here or in the seed-my-ontology-repo.pl

cmungall:
I believe with git you can just add the files, you can't add an empty dir

pbuttigieg:
True, but if you try to add files to a dir that doesn't exist you'll get an error. That's the intention here at least.

PR #6, opened 2016-06-02
cp release files to the release dir

pbuttigieg:
Currently copied to the src dir


pbuttigieg:
I think there's an extra space in the code that would mess up the cp target:
`cp $^ $RELEASEDIR && cp imports/* $RELEASEDIR imports`

It should be:
`cp $^ $RELEASEDIR && cp imports/* $RELEASEDIR/imports`

PR #45, opened 2018-04-22
[DO NOT MERGE YET] Use robot when making obo version of imports

cmungall:
This requires the changes in https://github.com/ontodev/robot/issues/64 which is not in the latest release yet

For discussion see https://github.com/owlcs/owlapi/issues/752


bjonnh:
CI failed because the line should be:
$(ROBOT) convert --check false -i $<  -f obo -o $@.tmp && mv $@.tmp $@

Not sure if it is normal, but it spits errors with:
http://bjonnh.net/share/clo_import.owl

cmungall:
Thanks!

Looks like the build fail now is to do with no globbing on some unixen

PR #49, opened 2018-05-10
Fix shell portability issues

kimrutherford:
We had problems with seed-my-ontology-repo.pl on Debian and Ubuntu because of some Bash-isms that aren't supported by /bin/sh.  Constructs like "subsets/*{obo,owl}" were failing.


cmungall:
Thanks!

Because I am so weirdly attached to my lazy bashisms I wrote an expander in the perl. Abstraction just to save a few repetitive lines not sure that's wise..

My fixes are in a branch that will be released soon

kimrutherford:
Great.  I've tested the current master branch and it works for me on Debian.

PR #62, opened 2018-07-22
Dosdp support

dosumis:


dosumis:
@matentzn - updated README on this branch.  Please give it a test drive.

dosumis:
@matentzn - Fixed a couple more issues on this branch.  Should now be working as per updated README.md. Please test.

matentzn:
@cmungall seems to me the reason why this fails is because robot does not recognise the --catalog parameters. Ideas?

cmungall:
Currently the docker container is built from robot master. Last release to
docker hub a week ago. Probably preceded addition of catalog. We can make a
new build but would like to start tracking formal robot releases soon

matentzn:
Thanks! Is this build something I can trigger, or will you do it?

matentzn:
I rebuilt it oskfull locally, and everything seems to works fine now. In order for the ODK to work properly, someone needs to rebuilt and push the docker oskfull to dockerhub. :) Thanks!

cmungall:
Hmm, you show up as an owner here 
https://hub.docker.com/u/obolibrary/dashboard/teams/?team=owners

matentzn:
Mea culpa. Pushing now.

matentzn:
After updating the oskfull on dockerhub, the travis job completes again.

template/src/ontology/foobar-edit.owl
```diff
@@ -7,6 +7,7 @@ Prefix(rdfs:=<http://www.w3.org/2000/01/rdf-schema#>)
 
 Ontology(<http://purl.obolibrary.org/obo/foobar.owl>
 Import(<http://purl.obolibrary.org/obo/foobar/imports/MY-IMPORTED_import.owl>)
+Import(<http://purl.obolibrary.org/obo/foobar/patterns/definitions.owl>)
```
matentzn: Not too sure whether this is the best way to do this. We should use a robot command to add the import, or only merge the definitions.owl into the final release. There will still be users of the ODK that do not want to use patterns.

dosumis: OK. I'll remove.  We can add a note to the README about adding the import manually if using patterns.

dosumis:
If no objections, I'll merge this pull request once i've:
1. Switched the default dosdp.yaml & tsv to something trivial
2. Removed the definitions.owl import from the editor's file.

PR #72, opened 2018-07-31
Make docker commands workable on windows (DONT MERGE)

matentzn:
I added a bat file for the seeding of a new repo and successfully generated a new ontology on windows (with only docker installed). 

However, there was one snippet in the seed script (perl) that was not compatible with the windows file system, I dont know why: listing all files in the template directory using finddepth() did not descent into subdirectories (listing only files in the top level dir). I changed the traversal method now. Locally I generated two repos, one with the old traversal, one with the new, and compared the log files plus diff of the resulting releases. I think they are identical. Before we merge: Is there any test we should run that makes sure the new directory listing is exactly the same as the old?

### DIFF
--- a/seed-my-ontology-repo.pl
+++ b/seed-my-ontology-repo.pl
@@ -110,12 +110,33 @@ else {
 my $TEMPLATEDIR = 'template';
 
 my @files;
-finddepth(sub {
-    return if($_ eq '.' || $_ eq '..');
-    push @files, $File::Find::name;
-},
-$TEMPLATEDIR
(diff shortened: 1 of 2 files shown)


dosumis:
@matentzn - Sounds like this is now fixed but you're being paranoid about tests?  Think this is probably urgent enough that we should go without as long as now gives same result on *nix (inc mac) & windows.

cmungall:
I looked at the code and the travis output. I'm confident to merge this

PR #76, opened 2018-08-03
DO NOT MERGE: Experiment with osklite

cmungall:


cmungall:
There are some advantages to using lite (100-200mb). However, currently python is required for all dosdp checks, this seems to inflate size too much: #78

matentzn:
Closing this without merging, we can revisit this one day. I think rather than having on ODK lite, its better to try and shrink the current image..

PR #78, opened 2018-08-03
DO NOT MERGE: Experimenting with adding python to osklite - this inflates image to …

cmungall:
…500MB, which means it is no longer lite...


matentzn:
Closing this for now. Maybe better to work on shrinking the main ODK image..

PR #85, opened 2018-08-07
MERGE: robot lightweight container

cmungall:
Used to make alpha release on obolibrary dockerhub to support
https://github.com/rctauber/robot-tutorial/pull/1

Not clear if this belongs here or in robot repo
Note current robot Dockerfile builds from source which
makes a larger image

cc @rctauber @jamesaoverton


jamesaoverton:
Now that we have proper ROBOT releases, we could just package those.

I haven't seen a huge benefit to Docker for ROBOT, since you have to mount the file system and stuff. It was handy for the ROBOT tutorial, though.

rctauber:
Hey @matentzn ! Thanks for pointing me to this issue. 

We set up the lightweight container for the ICBO tutorial last year and it would be great to have the container updated to `v1.4.1` for this year's tutorial (as I'm updating the tutorial itself for this version). We request everyone to install ROBOT in advance, but it's nice to have the option to get it quickly via Docker.

matentzn:
I have updated it and it seems to work;

I also pushed a new version on dockerhub.

https://hub.docker.com/r/obolibrary/robot/tags

Becky, this PR includes also a file run.sh which you can place in any directory from which you want to be running robot; it wraps some of the verbose docker boilerplate. So now if you want to use it, dropt the run.sh somewhere and run:

```
sh run.sh robot --version
```

Let me know how this goes! Usable even without merging, but I think this can be merged @cmungall

rctauber:
This is great, thanks so much @matentzn !

PR #115, opened 2018-10-05
Robot version in ODK docker as ENV

matentzn:
Updated to new ROBOT release (alpha) to be able to do the remove commands.


matentzn:
@cmungall Can this be merged? :)

cmungall:
Thanks! We should make sure this doesn't point to an alpha when we make the next ODK release.

matentzn:
Unfortunately we rely on the alpha at least to create the base product
(look at odk-base branch, it’s not merged yet). We need the robot remove
method :D

PR #117, opened 2018-10-08
Change ontology-starter-kit references to ontology-development-kit

marijane:
A very minor PR.  I noticed some stray references to the the old project name in comments and READMEs and such.  They were bugging me.


matentzn:
Thank you so much :)

PR #121, opened 2018-10-30
DO NOT MERGE: make sure definitions.owl is properly released

matentzn:


matentzn:
Whether to merge this PR comes down to the question of whether we want definitions.owl to be copied to the top level by default (thereby potentially annoying people that do not want to use dosdp).

PR #122, opened 2018-10-30
ODK base release product

matentzn:
(requires robot alpha because of remove command)

This requests needs a more thorough look at to be sure it is according to the new @balhoff specs

### DIFF
--- a/template/src/ontology/Makefile
+++ b/template/src/ontology/Makefile
@@ -27,10 +27,10 @@ DOSDPT=dosdp-tools
 # Top-level targets
 # ----------------------------------------
 
-all: all_imports sparql_test patterns $(ONT).owl $(ONT).obo
+all: all_imports sparql_test patterns $(ONT).owl $(ONT).obo $(ONT)-base.owl
 test: sparql_test all
 prepare_release: all
-	cp $(ONT).owl $(ONT).obo $(RELEASEDIR) &&\
+	cp $(ONT).owl $(ONT).obo $(ONT)-base.owl $(RELEASEDIR) &&\
 	mkdir -p $(RELEASEDIR)/imports &&\
 	cp imports/*.owl $(RELEASEDIR)/imports &&\
 	cp imports/*.obo $(RELEASEDIR)/imports &&\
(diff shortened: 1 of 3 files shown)


template/src/ontology/Makefile
```diff
@@ -49,7 +49,11 @@ $(ONT).owl: $(SRC)
 	$(ROBOT)  reason -i $< -r ELK -e none relax reduce -r ELK annotate -V $(BASE)/releases/`date +%Y-%m-%d`/$(ONT).owl -o $@
 $(ONT).obo: $(ONT).owl
 	$(ROBOT) convert --check false -i $< -f obo -o $(ONT).obo.tmp && mv $(ONT).obo.tmp $@
-
+	
+$(ONT)-base.owl: $(SRC) ../patterns/definitions.owl
+	$(ROBOT) remove --input $< --select imports --output $@ &&\
+	$(ROBOT) merge -i $@ -i ../patterns/definitions.owl -o $@ &&\
```
balhoff: @matentzn I think you should add an `annotate` here for the ont IRI and version IRI.

matentzn: Like:
```
annotate -V $(BASE)/releases/`date +%Y-%m-%d`/$(ONT)-base.owl
```

and

```
annotate --ontology-iri $(BASE)/$(ONT)-base.owl
```

balhoff: Yep

matentzn:
@balhoff hope that works. 

Build fails because of outdated OSKFULL version on dockerhub. Will fix this tomorrow! So if you are down with the changes proposed here, I will merge them tomorrow when the new ODK is on dockerhub.

template/src/ontology/Makefile
```diff
@@ -49,7 +49,11 @@ $(ONT).owl: $(SRC)
 	$(ROBOT)  reason -i $< -r ELK -e none relax reduce -r ELK annotate -V $(BASE)/releases/`date +%Y-%m-%d`/$(ONT).owl -o $@
 $(ONT).obo: $(ONT).owl
 	$(ROBOT) convert --check false -i $< -f obo -o $(ONT).obo.tmp && mv $(ONT).obo.tmp $@
-
+	
+$(ONT)-base.owl: $(SRC) ../patterns/definitions.owl
+	$(ROBOT) remove --input $< --select imports --output $@ &&\
+	$(ROBOT) merge -i $@ -i ../patterns/definitions.owl annotate -V $(BASE)/releases/`date +%Y-%m-%d`/$(ONT)-base.owl annotate --ontology-iri $(BASE)/$(ONT)-base.owl -o $@ &&\
```
balhoff: I think you can do this in one `annotate`, but doesn't matter much.

PR #126, opened 2018-11-10
DO NOT MERGE: Yaml driven config

cmungall:
First pass at yaml-driven Makefile generation
Intended as incremental progression beyond the current v static one.
Eventually the entire workflow could be driven by the yaml

See https://github.com/ontodev/robot/issues/37


cmungall:
Needs careful thought.

The spreading of logic through python, Jinja2 and the Makefile could get confusing.

The main thing to focus on is the project configuration specification. If we get this right, we can have any choice of engines for either generating workflows or executing directly.

template/src/ontology/Makefile.jinja2
```diff
@@ -0,0 +1,236 @@
...
+VCHECKS = equivalent-classes trailing-whitespace owldef-self-reference xref-syntax nolabels
+
+# Only needed until robot supports catalogs
+
+
+
+#  run all violation checks
+VQUERIES = $(foreach V,$(VCHECKS),$(SPARQLDIR)/$V-violation.sparql)
+sparql_test: $(SRC)
+	robot verify  --catalog catalog-v001.xml -i $< --queries $(VQUERIES) -O reports/
```
jamesaoverton: $(ROBOT)

template/src/ontology/Makefile.jinja2
```diff
@@ -0,0 +1,236 @@
...
+
+# ----------------------------------------
+# Sparql queries: Reports
+ # ----------------------------------------
+
+REPORTS = basic-report class-count-by-prefix edges xrefs obsoletes synonyms
+REPORT_ARGS = $(foreach V,$(REPORTS),-s $(SPARQLDIR)/$V.sparql reports/$V.tsv)
+all_reports: $(SRC)
+	robot query -f tsv -i $< $(REPORT_ARGS)
```
jamesaoverton: $(ROBOT)

template/src/ontology/Makefile.jinja2
```diff
@@ -0,0 +1,236 @@
...
+# ----------------------------------------
+
+# by default we use {{ project.reasoner }} to perform a reason-relax-reduce chain
+# after that we annotate the ontology with the release versionInfo
+$(ONT).owl: $(SRC)
+	$(ROBOT)  reason -i $< -r {{ project.reasoner }} relax reduce -r {{ project.reasoner }} annotate -V $(BASE)/releases/`date +%Y-%m-%d`/$(ONT).owl -o $@
+$(ONT).obo: $(ONT).owl
+	$(ROBOT) convert -i $< -f $(OBO_FORMAT_OPTIONS) obo -o $(ONT).obo.tmp && mv $(ONT).obo.tmp $@
+$(ONT).json: $(ONT).owl
+	$(ROBOT) convert -i $< -f json -o $(ONT).json.tmp && mv $(ONT).obo.tmp $@
```
jamesaoverton: `.obo.tmp` should be `.json.tmp`

jamesaoverton:
I think that using a template system to build a Makefile is a good first step. The Python OO wrapper is kind of gross but it's readable. (I don't like OO in general, and in Python I find it ruins a perfectly good imperative language, but that's just my biases. I'd prefer a declarative schema of some sort, but I admit that if you squint this looks pretty close to a schema.)

I made some comments on small things. In the OBI Makefile we use long option names and split ROBOT commands across lines, for readability: https://github.com/obi-ontology/obi/blob/master/Makefile

Overall, this Makefile seems bigger than it needs to be. I guess it's doing a bunch of different stuff.

template/src/ontology/Makefile.jinja2
cmungall: done

template/src/ontology/Makefile.jinja2
cmungall: done

cmungall:
I knew you wouldn't like the OO code :-)

Yes, the OO classes should not be a substitute for having a nice json-schema or similar.

Makefile.jinja2 was created from template/src/ontology/Makefile, and then some light weight templatization. So it was possibly too large to begin with.

The derived Makefile is currently bigger because right now it makes a separate target for every ground mirror target, rather than a generic rule. We're not necessarily committed to this, we could have the default be a wildcard rule and only write a ground target if the config has specific rules for that.

jamesaoverton:
I'm fine with the Python OO code. But just out of curiosity: My preferred approach would be code that reads the schema then "coerces" the config file into a nested data structure with all the right types. There are several ways to do this in Clojure. It's very useful when handling JSON. Do you know of anything like this in Python?

cmungall:
This uses the dacite module which takes the untyped nested dict from parsing the yaml and populates the strongly typed OO/dataclasses model. Is this what you mean?

There is also some procedural code to populate some missing values, this should be discussed.

cmungall:
Is there an easy way to switch from echo to no-echo? Echo can be useful when looking at travis logs for bugs

cmungall:
UPDATE:

This needs a bit more user friendly docs before merging into master, but I think we should switch to developing on this branch rather than making more changes on master that will be hard to switch over.

We still support doing everything via command line params; e.g. from the test in the Makefile

```
odk.py seed -c -d pato -d cl -d ro -t my-ontology3 myont
```

this essentially builds the yaml structure on the fly

but you can also do via an explicit yaml file:

```
odk.py seed -C examples/triffo/project.yaml
```

You can also choose to only create the Makefile:

```
odk.py create_makefile -C examples/triffo/project.yaml > Makefile
```

It should be relatively easy to figure out how things work, the template folder has files *.jinja2 that are expanded in the predicted way. But in cases where the filename is dynamic see `template/_dynamic_files.jinja2` which expands to a kind of tar file that then unrolls to multiple files whose names can be dynamic. @dosumis and @matentzn I think this will be particularly useful for the DPs.

cmungall:
Because I am rubbish at the gits, this PR is closed and superseded by #138, which is ready to merge!

PR #144, opened 2019-01-23
updating robot version and installing rsync

dougli1sqrd:


dougli1sqrd:
Looks like there's a syntax error in the python, in odk.py?
```
./odk/odk.py seed  -c -d pato -t my-ontology1 myont
  File "./odk/odk.py", line 41
    memory_gb : Optional[int] = None
              ^
SyntaxError: invalid syntax
Makefile:15: recipe for target 'test1' failed
make: *** [test1] Error 1
```

dougli1sqrd:
Looks like the code is expecting the python3 version to be 3.6 or higher as per: https://mypy.readthedocs.io/en/latest/cheat_sheet_py3.html
But when I run `python3 --version` I see that we are using 3.5.
```
docker run -v $PWD/:/work -w /work --rm -ti obolibrary/odkfull python3 --version
Python 3.5.3
```

cmungall:
Yes, I have been using 3.6 and 3.7 for a while now (but that doesn't have anything to do with this PR does it? Do you want to make a separate PR for the README documenting requirements for the seed command)

matentzn:
superceded by https://github.com/INCATools/ontology-development-kit/pull/152

Issue #145, opened 2019-01-23
Python version in Dockerfile is different than what odk.py expects causing tests to fail


dougli1sqrd:
The Dockerfile in `ontology-development-kit` installs python3 `RUN apk add --no-cache python3`. This installs python version 3.5:

```
docker run -v $PWD/:/work -w /work --rm -ti obolibrary/odkfull python3 --version
Python 3.5.3
```

Travis, runs `make test`, which at a certain point runs `odk.py`. This file uses the mypy type annotation style for variables only added in python 3.6: `memory_gb : Optional[int] = None` (from line 41 in odk.py).

This document here: https://mypy.readthedocs.io/en/latest/cheat_sheet_py3.html indicates that this syntax feature is only available in python 3.6. 

This is causing unrelated odk changes to fail the travis build. How this passed travis builds in the first place is an unknown, but this should be fixed so new PRs can pass Travis checks.

cmungall:
It's possible the yaml PR was failing originally - is there an easy way to use python3.6 in alpine?

matentzn:
Was fixed by the above.

PR #152, opened 2019-01-31
Updated dockerfile, rsync, pandas, version

matentzn:
(also added the hp example config, for reference)

Solution through updating alpine to latest version (3.9)

Ran the usual tests (make docker-tests), tried a few release pipelines. All fine.

supersedes https://github.com/INCATools/ontology-development-kit/pull/144 (NOTE rsync stuff etc incorporated here, please close!) 

addresses https://github.com/INCATools/ontology-development-kit/issues/145

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 ### From https://stackoverflow.com/questions/51121875/how-to-run-docker-with-python-and-java
 ### 1. Get Linux
-FROM alpine:3.7
+FROM alpine:3.9
 
 ### 2. Get Python, PIP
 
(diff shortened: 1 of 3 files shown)


matentzn:
Note that I am currently uploading the new ODK to dockerhub. Will take an hour or more from where I am. I will restart travis once its up.

dougli1sqrd:
Thanks! It looks like you added rsync twice? Line 29 and then on 58?

matentzn:
> subversion?

unfortunately needed to install numpy with pip3.

PR #155, opened 2019-02-01
Making definitions.owl functional for easier review

matentzn:


matentzn:
Oh, I thought -f only works for robot convert; check whether this suggestion is acceptable (using ofn instead of tmp).

PR #159, opened 2019-02-01
Git global email hack

matentzn:


matentzn:
I don't get why it fails. Everything was working fine before; Now the same error happens that we had yesterday (the syntax error with Optional [int], why we moved up to python 3.6.8). - That passed no problems a few commits back.

matentzn:
Anyone have an idea?
https://travis-ci.org/INCATools/ontology-development-kit/jobs/487585420

matentzn:
oh man i must be chaining the commants in the wrong way.. these will be executed outside of the odk.. oops. sec

matentzn:
All good, if you want to check the single line I added into travis.yml, I think it can be merged. @cmungall

Issue #158, opened 2019-02-01
Making sure ROBOT query for seeds includes imports


matentzn:
We make a lot of use of ROBOT query (Makefile). Now I had a lot of problems when using base-imports, because query does not take into account imports by default. I suggest to replace the current incantations by:

```
$(ROBOT) query --use-graphs true --catalog catalog-v001.xml -f csv -i $< --query ../sparql/terms.sparql $@
```
essentially adding

```
--use-graphs true --catalog catalog-v001.xml
```

Does this make sense @rctauber @cmungall

rctauber:
You shouldn't need to specify the catalog as ROBOT is pretty good about guessing that, unless you have a catalog in a different folder or with a different name. Does it work correctly if you *only* add `--use-graphs true`?

matentzn:
I think it should! Thanks for the clarification :)

PR #160, opened 2019-02-01
Add use-graphs true parameter to robot seed queries

matentzn:
See diff. Just added the parameter to make sure imported ontologies are taken into account. Please review carefully @cmungall @rctauber


matentzn:
Adresses https://github.com/INCATools/ontology-development-kit/issues/158

matentzn:
I removed two of the commands of which I was not sure, but I will make a ticket to document.

PR #163, opened 2019-02-05
Critical: Fixing various issues with the initial import seeds and the pattern creation

matentzn:
This PR contains two fixes:

1) obvious one to the pattern generation system (a wrong path in one of the goals)
2) A few issues with the default import system. I still dont like it: I would prefer ROBOT to [allow empty seeds](https://github.com/ontodev/robot/issues/434), resulting in empty ontologies being extracted. This is algorithmically well-defined behavior, and precludes the need of hacking some kind of default seed for the initial repo seeding process.

### DIFF
--- a/template/_dynamic_files.jinja2
+++ b/template/_dynamic_files.jinja2
@@ -162,28 +162,32 @@ Datatype: rdf:PlainLiteral
 
 #}
 ^^^ src/ontology/imports/{{ imp.id }}_terms.txt
-{{ imp.id|upper }}:0000001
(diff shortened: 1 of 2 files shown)


cmungall:
1. good
 2. agreed, but should we not keep my hack until we have a new robot release with the force option formally released?

matentzn:
@cmungall Yes, it is all still there. I just expanded it to work with [some more ontologies](https://github.com/INCATools/ontology-development-kit/pull/163/files)..

The comment was only for announcing a future intention. :P Can this PR be merged?

PR #166, opened 2019-02-09
New Pattern pipelines in ODK

matentzn:
This took me a long time, too long, and I am sure there are some bugs (in particular in terms of makefile redundancy - just looking at the execution order makes me queasy now). But: It seems to work. All tests still seem to pass (I also pushed a new ODK up, I know it is somewhat circular, but the tests cant pass without). If this PR gets rejected, I will undo the latest odk release. 

The main things to look out for are:
1) Changes to the odk.py and to the jinja Makefile. There are some things that could be done better; in particular more effective ROBOT chaining (ont-base did not work properly when I chained it, but might have been another issue).
2) I know it is a nightmare to review, but my biggest worry is makefile execution order. For example, maybe the goal: 'prepare_patterns' is causing some problems (I dont think so, but there is a nagging sense in me). It seems to do something similar to your trigger goals for the imports. Maybe you have an idea to do it better.

This will be the last major ODK update for a while; with that, I can now start working properly. So I promise, after this one, I will only make small, easy to review PRs :P I have added a pattern-test to the Makefile to make sure pattern pipeline generation is tested on travis as well.


template/src/ontology/Makefile.jinja2
```diff
@@ -162,21 +167,21 @@ ANNOTATE_VERSION_IRI = annotate -V $(ONTBASE)/releases/`date +%Y-%m-%d`/$@.owl
 
 # by default we use {{ project.reasoner }} to perform a reason-relax-reduce chain
 # after that we annotate the ontology with the release versionInfo
-$(ONT).owl: $(SRC)
+$(ONT).owl: $(SRC) {% if project.use_dosdps %}$(PATTERNDIR)/definitions.owl{% endif %}
```
cmungall: I have a very mild preference for this pattern

```
{% if project.use_dosdps %}
OTHER_SRC = $(PATTERNDIR)/definitions.owl
{% endif %}
...
$(ONT).owl: $(SRC) $(OTHER_SRC)
```

template/src/ontology/Makefile.jinja2
```diff
@@ -133,8 +134,12 @@ show_assets:
 
 # This should be executed by the release manager whenever time comes to make a release.
 # It will ensure that all assets/files are fresh, and will copy to release folder
-prepare_release: $(ASSETS)
-	rsync -R $^ $(RELEASEDIR) &&\
+prepare_release: $(ASSETS) $(PATTERN_RELEASE_FILES)
```
cmungall: I might be inclined to include the pattern release file in ASSETS, but we can do this later.

matentzn: The reason why I didn't (I tried :)) was that that the rsync target dir did not match to the required dir for definitions.owl. definitions.owl needs to be in ./patterns/definitions.owl, but I cant get it there without changing the structure of the rsync! :P But yeah, in future, we can fix it.

PR #180, opened 2019-02-26
changes to .bat file to make this run on windows

kaiiam:
Here is an attempted change to the .bat file to try and run on windows. Cutting out the pl script and replacing the python.


matentzn:
@kaiiam Did you test this on a Windows machine? Does it run?

Issue #179, opened 2019-02-26
Need an odkfull docker release


balhoff:
Is there any reason not to make a docker release for odkfull from master? We need to get the latest owltools for GO.

matentzn:
It just adds an extra layer of security; I quite like it know. But I am open, if everyone votes for dynamic built, I can change that as well. Slight tendency to the manual approach now.

I just built and pushed a new version up, but you can always do this yourself by simply running `make docker-publish` in the root directory of the odk.

balhoff:
Thanks @matentzn! I think I was unclear: I didn't mean to push for a totally dynamic build; just wanted to make sure there wasn't anything in progress on master that shouldn't be released. For some reason the newest owltools didn't get into this image. I will try to figure it out.

matentzn:
Oh! Yeah let me know what you find! That is important! The build log said that something related to dosdp and owl tools was downloaded.

cmungall:
At the moment the dockerfile uses the owltools built on Jenkins maybe that
didn't build?

balhoff:
I ran the Jenkins build yesterday. I downloaded it directly a few minutes ago to check the version. Seems okay.

matentzn:
well the only thing I can think off is that docker-build skips the step where the owltools is installed. have you tried running make docker-build locally and then test the odk?

balhoff:
Doing it now!

matentzn:
And?

balhoff:
I rebuilt and published and the new owltools seems to have gotten in.

PR #187, opened 2019-02-27
Add no-cache to docker build. Ensure tools are downloaded each time.

balhoff:
@matentzn @cmungall I needed this to ensure that the new owltools was downloaded. Otherwise it was using a cache from a previous download. Should we do this every time?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,7 +37,7 @@ VERSION = "v1.2.5"
 IM=obolibrary/odkfull
 
 docker-build:
-	@docker build -t $(IM):$(VERSION) . \
+	@docker build --no-cache -t $(IM):$(VERSION) . \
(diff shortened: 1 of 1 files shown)


balhoff:
See https://github.com/INCATools/ontology-development-kit/issues/179

matentzn:
Yeah, I think this should be done, even if it is painful for me (as ODK dev, I need to rebuild the image up to 20 times in a single day). I guess I can remove the -no-cache manually on an ODK dev day. The [alternatives](https://stackoverflow.com/questions/31782220/how-can-i-prevent-a-dockerfile-instruction-from-being-cached) are just hacks (selectively caching). I agree to the change!

matentzn:
on second thought, maybe a seperate Make targe for docker-build-cache would be appreciated :P

PR #190, opened 2019-03-02
DO NOT MERGE: Envo support

cmungall:


template/src/ontology/Makefile.jinja2
```diff
@@ -26,8 +26,11 @@ EDIT_FORMAT=                {{ project.edit_format|default('owl') }}
 SRC =                       $(ONT)-edit.$(EDIT_FORMAT)
 ROBOT=                      robot
 RELEASEDIR=                 ../..
-PATTERNDIR=                 ../patterns
-REPORTDIR=                 reports
+{% if project.pattern_pipelines_group %}
+PATTERNDIR=                 {{ project.pattern_pipelines_group.pattern_directory }}
```
matentzn: I dont think this will work; at least not in its current form. The current pipeline assumes 1 pattern directory, which is in PATTERNDIR/dosdp-patterns, and an arbitrary number of pipeline directories, which can be found under patterndir/data/. There is currently no support for having a seperate pattern directory for every single pipeline. Is that every necessary? Do you want to be able to use two different patterns with the same name?

matentzn: So as an example, we currently have
- ontology
- patterns
  - dosdp-patterns [abnormal.yaml, b.yaml, c.yaml]
  - data
    - pipeline 1 [abnormal.tsv]
    - pipeline 2 [b.tsv, c.tsv]
    - pipeline 3 [b.tsv]

template/src/ontology/Makefile.jinja2
```diff
@@ -101,9 +104,9 @@ PATTERNS = {% for x in project.pattern_group.products %} {{ x.id }}{% endfor %}
 PATTERNS = 
 {% endif %}
 PATTERN_ROOTS = $(patsubst %, $(PATTERNDIR)/%, $(PATTERNS))
-PATTERN_FILES = $(foreach n,$(PATTERN_ROOTS), $(n).owl)
+PATTERN_OWL_FILES = $(foreach n,$(PATTERN_ROOTS), $(n).owl)
 
-all_patterns: $(PATTERN_FILES)
```
matentzn: This goal, and the two preceeding lines, do not do anything currently (they are not called from anywhere). I think this should be deleted! (check the patterns: goal much later (somewhere around line 326), that is responsible for actually compiling the patterns.

template/src/ontology/Makefile.jinja2
cmungall: This is to support ontologies that already place the patterns dir somewhere else, or have patterns and tsvs in separate folders.E.g. envo.  If it's too complex they could simply git mv

matentzn: Patterns and tsvs are in different folders right now! So it is only about the exact location of the patterns directory? I kind of like this to be standard, no? or do you think it is an effort for people to move their patterns to this patterns/dosdp-patterns directory?

matentzn:
Closing this for good, too many changes happens since.

PR #193, opened 2019-03-06
DO NOT MERGE (but please start checking, complaining): ODK releasefile overhaul

matentzn:
This PR is a bit massive, unfortunately, making sure of the following:
- export formats in the configuration files are not hard coded any more and can be changed
- we have now the option of 7 release artefacts, all of which are configurable, except for base and full, which are mandatory.
- all release artefacts are exported to all formats as specified
- some minor fixes to the pattern pipeline as well

Please do not merge until I have vetted the release artefacts with @dosumis 
Also, avoid changing the Makefile until we have this sorted, as the changes are extensive and conflicts likely!


template/src/ontology/Makefile.jinja2
```diff
@@ -219,15 +325,24 @@ imports/%_terms_combined.txt: seed.txt imports/%_terms.txt
...
-		 annotate --ontology-iri $(ONTBASE)/$@ --version-iri $(ONTBASE)/releases/$(TODAY)/$@ --output $@
+		annotate --ontology-iri $(ONTBASE)/$@ --version-iri $(ONTBASE)/releases/$(TODAY)/$@ --output $@
 .PRECIOUS: imports/%_import.owl
 
+{% if 'obo' in project.export_formats -%}
 # convert imports to obo.
 # this can be useful for spot-checks and diffs.
 # we set strict mode to false by default. For discussion see https://github.com/owlcs/owlapi/issues/752
 imports/%_import.obo: imports/%_import.owl
-	$(ROBOT) convert --check false -i $< -f obo -o $@.tmp && mv $@.tmp $@
```
cmungall: Is this deliberate? This is dangerous as convert can fail leaving an empty file. I always use the $@.tmp paradigm, even if it is ugly. (For consistency now I do `$@.tmp.obo` since ROBOT uses suffix as default)

matentzn: THATS why you had this in!!! Ahhh! Thanks for pointing this out, I put it back now! :)

PR #210, opened 2019-04-23
Run ODK from docker (Fixed)

matentzn:


matentzn:
(all local tests have passed)

Issue #214, opened 2019-05-09
OBO validator into ODK


matentzn:
Transferring an email thread into a ticket.

OBO files are typically 'terminals' in our release chains - i.e. final release products. It is very common to **hack** OBO files into shape after they are produced by the OWL API based tools such as ROBOT or owltools. For example:

- owl-axioms are stripped out (all ODK ontologies). Example:
```
$(ONT)-base.obo: $(ONT)-base.owl
	$(ROBOT) convert --input $< --check false -f obo $(OBO_FORMAT_OPTIONS) -o $@.tmp.obo && grep -v ^owl-axioms $@.tmp.obo > $@ && rm $@.tmp.obo
```
- date annotations are added manually (Disease Ontology):

```
annotate --version-iri "$(OBO)doid/releases/$(DATE)/$(notdir $@)"\
	 --output $(basename $@)-temp.obo && \
	grep -v ^owl-axioms $(basename $@)-temp.obo | \
	grep -v ^date | \
	perl -lpe 'print "date: $(TS)" if $$. == 3'  > $@ && \
	rm $(basename $@)-temp.obo && echo "Created $@"
```

and many more.  Moreover, some interactions between ROBOT remove and specific annotation properties (such as subset and synonym) can cause problems as well. Since after serialisation, no tools are touch these files until ingest downstream, we dont catch when we accidentally break them; which causes annoyances in our stakeholders (MODs, canto, intermine, AGR ingest and more), which we should and can avoid. 

The preliminary cause of action is this:

1. Include the [FastOBO](https://github.com/althonos/fastobo/) validator into the ODK (@althonos + @matentzn)
2. Make sure that the FastOBO validation process catches the most common errors that are flagged up on ingest (date/xref formatting issues, missing annotation defs, missing ontology annotations etc)
3. Add another round of ROBOT qc on the FINAL (post-hacked) OBO artefact
4. We will share the additional QC with pipelines that are not using the latest ODK (@rctauber DO, ECO; GENO; SEPIO etc)

While we are working on this in the next weeks, @sbello has agreed to keep us updated:

Whenever an OBO ingest on the AGR side fails, add a comment here with the reason of failure, affected ontology and ideally the tool affected. Everyone else who has issues like that is of course invited to add their ingest failures to this issue as well! Much appreciated! @dosumis @cmungall @chris-grove

althonos:
[...]

PR #215, opened 2019-05-16
Add `fastobo-validator` to Dockerfile

althonos:
I added the latest stable `fastobo-validator` to the Dockerfile as discussed in #214 . Usage is as follow:
```make
x.obo: x.owl
  $(ROBOT) convert --check false -i $< -f obo -o $@.tmp.obo \ 
  && fastobo-validator $@.tmp.obo \
  && mv $@.tmp.obo $@; 
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -48,6 +48,9 @@ RUN wget https://github.com/ontodev/robot/releases/download/$ROBOT/robot.jar -O 
     
 ENV PATH "/tools/:$PATH"
 
+# Setup fastobo-validator
+RUN wget https://dl.bintray.com/fastobo/fastobo-validator/stable/fastobo_validator-x86_64-linux-musl.tar.gz -O- | tar xzC /tools
+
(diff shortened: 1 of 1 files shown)


althonos:
By the way, the final image could be made slimmer of about 100MB simply by removing `gcc` and other build dependencies after `pandas` is built :wink:

matentzn:
This is all amazing! I will test it tomorrow. If you want to add some cleanup stuff that you think are not necessary (like removing gcc) feel free to do so any time! Thanks for the suggestion!

matentzn:
I had to roll back the uninstall dependency command because unfortunately, it uninstalled some stuff that was needed elsewhere, in particular `make` for some reason... :P I will take it under advisement though and one day, when we launch another effort to shrink the image, we will try it again.

althonos:
Maybe you can just uninstall `gcc` when you're done, its the largest build dependency and probably the one you don't really need :wink:

Issue #232, opened 2019-06-24
Date in odk docker container is not the same as outside


cmungall:
Noticed by @kltm and @dougli1sqrd 

Note we make use of the current date to set the versionIRI:

```
TODAY   := $(shell date +%Y-%m-%d)
```

one thing to be aware of (this is from west coast)

```
~/repos/go-ontology/src/ontology(master) $ date
Mon Jun 24 16:13:53 PDT 2019
~/repos/go-ontology/src/ontology(master) $ docker run --rm -ti obolibrary/odkfull date
Mon Jun 24 23:13:58 UTC 2019
```

This is not necessarily wrong or particularly unexpected, but release managers should be aware of this. If I was to start making a release at 5pm I may be surprised to see tomorrows date. Or if I started 4.55pm, abandon part way, resume at 5pm, my versionIRI tags for different artefacts will be off.

This was noticed on the go pipeline, which is executed automatically rather than by a release manager.

Maybe we should encourage clients to set TODAY from outside so they have total control

PR #233, opened 2019-06-25
Issue 232 external date

dougli1sqrd:
This let's the project yaml have a property called `use_external_date`, which defaults to False. 

If True, when an ontology project is created and templated, the `run.sh` will set the docker environment with 
```
TODAY=`date +%Y-%m-%d`
```
when running commands.

`TODAY` is then picked up by the Makefile in `src/ontology` to properly set the date and Version URI.

### DIFF
--- a/odk/odk.py
+++ b/odk/odk.py
@@ -6,7 +6,7 @@ See https://github.com/ontodev/robot/issues/37
 
 For help on command line usage:
 
-odk.py
+odk.py 
 
 """
 from typing import Optional, Set, List, Union, Dict, Any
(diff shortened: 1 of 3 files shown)


dougli1sqrd:
For https://github.com/INCATools/ontology-development-kit/issues/232

odk/odk.py
```diff
@@ -50,20 +50,20 @@ class Product(JsonSchemaMixin):
 
     Here, a product is something that is produced by an ontology workflow.
     A product can be manifested in different formats.
-
```
cmungall: It would be good to avoid spurious diffs. Do you want to write guidelines and add a linter to travis? How do other projects do this?

cmungall:
It's hard to evaluate due to the spurious diffs but I think this is makes sense. Also spurious diffs can cause merge conflicts, not sure if Nico has any open. How hard is it to redo PR with only meaningful changes?

odk/odk.py
dougli1sqrd: Ah yeah, sorry about that. I can try and fix my editor from changing whitespace like that. I have no opinion really as to "correct" white space. This is just something my editor does to everyone's annoyance. I don't think we need a linter or anything.

dougli1sqrd:
I can try and redo it. Sorry, that's just my editor being weird. I don't know why it does it, but I'm sure it's a setting I can turn off.

dougli1sqrd:
okay white space should be fixed.

matentzn:
Some whitespaces are still there. And I learned something new ?= in the makefile! Awesome.

PR #241, opened 2019-07-08
Including Konclude

matentzn:
In order to have any hope of doing OWL DL reasoning beyond EL on some ontologies, I wanted to play a bit with the possibility of including Konclude.. Does not work yet, but if you are against this, let me know @cmungall

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -43,9 +43,9 @@ RUN wget http://build.berkeleybop.org/userContent/owltools/owltools -O /tools/ow
     wget http://build.berkeleybop.org/userContent/owltools/owltools-oort-all.jar -O /tools/owltools-oort-all.jar 
 
 # Installing Konclude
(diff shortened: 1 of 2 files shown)


matentzn:
@cmungall 

Andreas (Konclude dev) has compiled Konclude for us so it can be used with Alpine.

This is super cool. For a mere 25 MB in size difference, we get really nice and fast OWL 2 DL reasoning. This will allow me to classify stuff with disjunction and universals reliably where HermiT would fail (DPO with Hermit: 60 min, DPO with Konclude: 5 sec). I have tested this in the context of DPO development; I think it would be great to have this in ODK. What do you think?

balhoff:
This sounds great!

cmungall:
+1 pressed for time do you mind fixing the conflicts @matentzn

PR #244, opened 2019-07-16
Odk robot param

matentzn:
Make it possible to overwrite ROBOT jar parameter


matentzn:
@cmungall the idea is here to be able to substitute the ROBOT jar on the fly so I can quickly build a test ODK with a ROBOT test build (see [here](https://github.com/monarch-ebi-dev/robot_test/blob/master/README.md) for example use).

Do you think there is a problem with doing it like that? Seems to work fine!

PR #246, opened 2019-07-16
Robot catalog option

matentzn:
- Added ROBOT catalog option as option to config file;
- ROBOT now explicitly refers to the catalog


matentzn:
Please only look at odk/odk.py  and template/src/ontology/Makefile.jinja2

Issue #245, opened 2019-07-16
Seeding incomplete when Ontology contains components


matentzn:
We should make components a real thing. A component is a subset of the ontology that is managed outside of the edit file; you may want to, for example, curate some specific kinds of annotations like xrefs in a seperate .owl file. Example: c-elegans-phenotype ontology contains the EQ component (all non-patternised EQ definitions) and fbcv contains the DPO component (the simple version of DPO).

I believe to extract the correct seeds, we need this:

```
$(ROBOT) merge --input $< $(patsubst %, -i %, $(OTHER_SRC)) --collapse-import-closure false \
	query --use-graphs false -f csv --query ../sparql/simple-seed.sparql $@.tmp
```

i.e.: we first merge the src with its components, and then extract the seed using --use-graphs false.

balhoff:
Are the components imported? If so why not `--use-graphs true`?

matentzn:
That is what it is now; but how will you get your CHEBI seed if imports are imported and components are imported?

balhoff:
Oh yeah I see—makes sense!

matentzn:
Started a component extension here, which works:

https://github.com/INCATools/ontology-development-kit/pull/247

What still needs to happen:

- [ ] Make sure that components are taken into accounts when seeds are created

matentzn:
That has been fixed a while back.

PR #247, opened 2019-07-16
First draft at creating components

matentzn:


matentzn:
This is does not fully solve #245 but its a start. Seems safe but look through diff to eyeball potential issues.

PR #262, opened 2019-08-02
Fixes subset injection problem

### LINKED ISSUE #261: Some negative interactions between module extraction and OBO
I think the problem is this:

1. Extract an arbitrary module from RO which contains at least one class belonging to a subset:
```
  <!-- http://purl.obolibrary.org/obo/RO_0002218 -->

    <owl:ObjectProperty rdf:about="http://purl.obolibrary.org/obo/RO_0002218">
        <rdfs:subPropertyOf rdf:resource="http://purl.obolibrary.org/obo/RO_0000057"/>
        <obo:IAO_0000112>&apos;heart development&apos; has active participant some Shh protein</obo:IAO_0000112>
        <obo:IAO_0000114 rdf:resource="http://purl.obolibrary.org/obo/IAO_0000125"/>
        <obo:IAO_0000115>x has participant y if and only if x realizes some active role that inheres in y</obo:IAO_0000115>
        <obo:IAO_0000116>This may be obsoleted and replaced by the original &apos;has agent&apos; relation</obo:IAO_0000116>
        <obo:IAO_0000117>Chris Mungall</obo:IAO_0000117>
        <obo:IAO_0000118>has agent</obo:IAO_0000118>
        <oboInOwl:inSubset rdf:resource="http://purl.obolibrary.org/obo/ro/subsets#ro-eco"/>
        <rdfs:label rdf:datatype="http://www.w3.org/2001/XMLSchema#string">obsolete has active participant</rdfs:label>
        <owl:deprecated rdf:datatype="http://www.w3.org/2001/XMLSchema#boolean">true</owl:deprecated>
    </owl:ObjectProperty>
```

2. The module extraction misses the crucial axiom (module extraction is defined only for logical axioms, not annotations):

```
AnnotationProperty: 'eco subset'

    Annotations: 
        rdfs:label "eco subset"
    
    SubPropertyOf: 
        subset_property
```

3. OBO requires this axiom to be present to understand it needs the `subsetdef:` declaration. 

This causes problems such as this one:
https://github.com/FlyBase/drosophila-developmental-ontology/issues/5

Any idea for a solid solition, I shall implement it.

matentzn:
Fixes #261


matentzn:
@cmungall review carefully as this tempers with the critical import functionality. Please let @rctauber give her opinion before merging this in; if you are both fine we merge. Becky, my question to you is: I get some minor diff inconsistencies when running a sparql update like the one proposed here; in particular, data-type declarations for strings get removed; is this intended/an artefact of the rdf spec that says that string is the default data type?

matentzn:
(btw this solution solves the problem reported in #261 )

matentzn:
@rctauber should i use .ru file extension instead of sparql?

jamesaoverton:
I'm pretty sure that the differences you're seeing are because your SPARQL CONSTRUCT is run by Jena, and Jena uses the RDF 1.1 standard which omits `xsd:string` data types, while OWLAPI uses the older RDF 1.0 behaviour. This is a known issue, and we don't have a better solution. See the [ROBOT 1.2.0 release notes](https://github.com/ontodev/robot/releases/tag/v1.2.0)

matentzn:
I also cant see why this fails (4 tests run successfully, and for some reason the 5th one fails when generating the import); Locally, all tests pass just fine; explicitly invoking the test that is failing here on Jenkins does not cause a problem on my machine.

Anyone an idea how I can get odk.py to print a few lines of the error message? @dougli1sqrd @althonos @cmungall 

I guess its somewhere around this line: https://github.com/INCATools/ontology-development-kit/blob/b45bae030a93a9a30aec70649426fc61addd6ad3/odk/odk.py#L719

And it must be some weird interaction with make and stderr; through odk.py, I only get a very nondescript error. If I cd in the directory and run the ODK from in there, I get a normal error message.. Dont know whats going on!

matentzn:
Thanks @jamesaoverton I think I can live with that!

beckyjackson:
The SPARQL UPDATE docs recommend that the file extension is `.ru` - so it might be good to change it just to follow that. ROBOT doesn't care what the extension is as long as the syntax in the file is correct 🙂

matentzn:
@beckyjackson sorry I only now realised your GH handle has changed :P 

Can you sanity check the pull request? Does it otherwise look ok (changed the file extension as well)

template/src/ontology/Makefile.jinja2
```diff
@@ -249,7 +249,7 @@ OTHER_SRC = {% if project.use_dosdps -%} $(PATTERNDIR)/definitions.owl{% endif -
 
 
 $(ONTOLOGYTERMS): $(SRC) $(OTHER_SRC)
-	$(ROBOT) query --use-graphs true -f csv -i $< --query ../sparql/{{ project.id }}_terms.sparql $@
+	$(ROBOT) query --use-graphs true -f csv -i $< --query ../sparql/{{ project.id }}_terms.ru $@
```
beckyjackson: I don't think this line was meant to be changed to `.ru`

template/src/ontology/Makefile.jinja2
```diff
@@ -375,6 +375,7 @@ imports/%_terms_combined.txt: seed.txt imports/%_terms.txt
 # This pattern uses ROBOT to generate an import module
 imports/%_import.owl: mirror/%.owl imports/%_terms_combined.txt
 	@if [ $(IMP) = true ]; then $(ROBOT) extract -i $< -T imports/$*_terms_combined.txt --force true --method BOT \
+		query --update ../sparql/inject-subset-declaration.sparql \
```
beckyjackson: This is the `.ru` file

beckyjackson:
I just left two minor comments, I think the wrong file was changed. Other than that, the query looks good.

template/src/ontology/Makefile.jinja2
matentzn: Lol ooooops. Thanks

matentzn:
Thanks Becky, perfect then! Now its up to @cmungall to approve

althonos:
@matentzn : seems like `odk.py` [only prints STDERR on error](https://github.com/INCATools/ontology-development-kit/blob/b45bae030a93a9a30aec70649426fc61addd6ad3/odk/odk.py#L723), so if some tools are printing to STDOUT instead you won't see it.

matentzn:
@cmungall little poke for review

PR #266, opened 2019-08-04
Print stdout when error occurs during seeding

matentzn:
If an error occurs during the seed command, it was sometimes impossible to tell what has caused it. Printing stdout as well as stderr might help in this case!

@althonos Can you also check this makes sense?

### DIFF
--- a/odk/odk.py
+++ b/odk/odk.py
@@ -720,6 +720,7 @@ def runcmd(cmd):
     (out, err) = p.communicate()
     logging.info('OUT: {}')
     if err:
+        logging.error(out)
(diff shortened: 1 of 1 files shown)


odk/odk.py
```diff
@@ -720,6 +720,7 @@ def runcmd(cmd):
     (out, err) = p.communicate()
     logging.info('OUT: {}')
```
althonos: I think this should be `logging.info('OUT: {}'.format(out))` instead!

matentzn: Thanks, that looks more sensible now!

PR #268, opened 2019-08-05
Creating releases on GitHub

matentzn:
This is the first attempt to do this. It seems to work:

1) Try to retrieve a release for the current date;
2) If there add assets to that release
3) If not there, create the release, then add assets to that release

This solution is realised using the GitHub REST API (implemented with curl);  

I would like to merge this sooner rather than later to get some people to test it (I flagged it as "highly experimental" and did not connect it with the other release commands). One thing I would like to do if someone knows how to easily do it (I could find out, but dont have the bandwidth): at the moment, each call to the API requires the user to authenticate; I would like to authenticate only once, obtaining some kind of an access token, and then use that for the remaining calls. If anyone can tell me exactly how this is done, I can implement, else I am happy to move forward as is for now.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,5 @@ venv/
 template/src/patterns/catalog-v001.xml
 .idea
 upgrade_repos.sh
+examples/phenotype-ontologies/zeco-odk-nico-curl.yaml
+examples/phenotype-ontologies/zeco-odk-nico.yaml
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ CMD = ./odk/odk.py seed
 
 EMAIL_ARGS=
 
-test: test1 test2 test3 test4 test-go-mini test-patterns test-release1 test-release2 test-release3
+test: test1 test2 test3 test4 test-go-mini test-patterns test-release1 test-release2 test-release3 test-github-release
 
 test1:
 	$(CMD) $(EMAIL_ARGS) -c -d pato -t my-ontology1 myont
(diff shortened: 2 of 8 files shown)


matentzn:
Problem is on order to test this I need to roll this out as an experimental feature; any objections @cmungall

cmungall:
The advantage of doing a specific piece of code is it's easier to have options like `-f` which I found to be useful

matentzn:
@cmungall So as you can see, I fiddled your approach in as an - untested - alternative.

- the odk config file now can be configured with either "github_curl" or "github_python" (potentially extensible to gitlab_curl etc)
- Your [make-release-assets.py](https://github.com/monarch-initiative/mondo/blob/master/src/utils/make-release-assets.py) script is now part of odk proper and can be called simply by calling `make-release-assets.py xyz`.
- The goal that is generated with the `github_python` option is simply:
```
# GitHub release (HIGHLY experimental)

RELEASEFILES= zeco.owl
TAGNAME=v$(TODAY)
GITHUB_RELEASE_PYTHON=make-release-assets.py

public_release:
	ls -alt $(ASSETS)
	$(GITHUB_RELEASE_PYTHON) --release $(TAGNAME) $(RELEASEFILES)
```

I have not tried this, nor have I looked at the python script; but consider this option perhaps before you edit the mondo version again, to have everything in one place.

I have a feeling we will be stuck with github for a while; the impossibility to refer to a file in the latest release page of Zenodo basically forces us to. Once this possible, we could write a zenode_api release goal as well!

PR #269, opened 2019-08-05
Updated documentation to abstract away from GitHub when appropriate

matentzn:
First pass. Is that what you meant @cmungall


cmungall:
travis fail due to a network glitch?

PR #288, opened 2019-09-24
More roboust way to download patterns from external file

### LINKED ISSUE #260: Make sure that patterns that have JUST been added to external.txt compile
Right now it takes to run of the odk to pick up new patterns.

### LINKED ISSUE #271: Fix Docker 123 errors when external.txt contains CRLF line endings
Solution:

```
$(PATTERNDIR)/dosdp-patterns: .FORCE
    sed -i 's/\r\n/\n/g' $(PATTERNDIR)/dosdp-patterns/external.txt
    cat $(PATTERNDIR)/dosdp-patterns/external.txt | sed 's!.*/!!' | sed 's! !!g' |  xargs -I{} rm -f $(PATTERNDIR)/dosdp-patterns/{}
    cat $(PATTERNDIR)/dosdp-patterns/external.txt | sed 's! !!g' | xargs -I{} wget -q {} -P $(PATTERNDIR)/dosdp-patterns
```

Probably better to do something along these lines: https://help.github.com/en/articles/configuring-git-to-handle-line-endings#per-repository-settings

But as a quick hack, this will suffice. Thans @jseager7

matentzn:
The old solution gave us a lot of trouble because it was sensitive to CRLF line endings. Rather than dealing with those, I decided to go another route. The new solution make a slightly hacky assumption (There is no file with the extension `.yaml.1` in the dosdp patterns directory, but it otherwise just works better. 

EDIT: Else fixes issue that caused newly added patterns to not be added before tsvs are compiled.

Fixes #271
Fixes #260

### DIFF
--- a/template/src/ontology/Makefile.jinja2
+++ b/template/src/ontology/Makefile.jinja2
@@ -536,8 +536,8 @@ pattern_schema_checks: $(PATTERNDIR)/dosdp-patterns
 
 #This command is a workaround for the absence of -N and -i in wget of alpine (the one ODK depend on now). It downloads all patterns specified in external.txt
 $(PATTERNDIR)/dosdp-patterns: .FORCE
-	cat $(PATTERNDIR)/dosdp-patterns/external.txt | sed 's!.*/!!' | sed 's! !!g' |  xargs -I{} rm -f $(PATTERNDIR)/dosdp-patterns/{}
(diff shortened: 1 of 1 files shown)


matentzn:
Also includes fix for issue that newly added patterns were downloaded to late in the process

Fixes #260

Issue #302, opened 2019-12-20
Migrate docker base image to avoid DNS issues


cmungall:
From @kltm:

```
The terminal generated command seems to be:

curl -L -o mirror/uberon.owl.tmp.owl
http://purl.obolibrary.org/obo/uberon.owl

Failing with:

curl: (6) Could not resolve host: ontologies.berkeleybop.org

Playing more, it is actually semi-random: can get the same problem with
ping:

/tmp/mammalian-phenotype-ontology/src/ontology # ping
ontologies.berkeleybop.org
ping: bad address 'ontologies.berkeleybop.org'
/tmp/mammalian-phenotype-ontology/src/ontology # ping
ontologies.berkeleybop.org
PING ontologies.berkeleybop.org (13.35.127.20): 56 data bytes
64 bytes from 13.35.127.20: seq=0 ttl=245 time=2.393 ms

I'd note here that the alpine image that the odk is based on makes some
kinds of debugging quite hard. I'd recommend basing on something
minimal, but with a few more features. Playing around some:

This does not happen on my local machine.
This does not happen in the ODK image on my local machine.
This does not happen on the Jenkins machine.
This does not happen on the Jenkins machine with an Ubuntu docker image.
This does not (so far seem to) happen on the Jenkins machine with
obolibrary/odkfull:v1.1.7.
This only seems to happen on the Jenkins machine with the current
obolibrary/odkfull:latest.
```

It seems there may be a known issue with DNS on alpine:

 - https://github.com/gliderlabs/docker-alpine/issues/476
 - https://unix.stackexchange.com/questions/441664/alpine-linux-sometimes-dns-is-not-resolved
 - https://forums.docker.com/t/resolved-service-name-resolution-broken-on-alpine-and-docker-1-11-1-cs1/19307/22

kltm:
While there are known issues (I looked at a few of them yesterday and walked through the Dockerfile history), the fact that I can consistently cause the problem to occur on one machine but I cannot on the other means that at least part of the problem is the interaction between the current image (likely at the base), docker, and the host system. As I've never run into something quite like that before, and it's likely something pretty deep, I think the fastest way forward would be to switch base images, at least away from what is currently there and ideally to something more "robust".

cmungall:
Could it be the version of Docker that is used on the problematic machine?

kltm:
[...]

PR #307, opened 2020-01-23
WIP: migrate docker to ubuntu:18.04

matentzn:
- We are faced with this daunting issue:
https://github.com/INCATools/ontology-development-kit/issues/302

Our hope is that swapping to the ubunti base will solve the issue.


balhoff:
Would be awesome to get ammonite in before the release.

PR #339, opened 2020-05-11
fix variable typo - FORMATS not FORMAT

cmungall:


matentzn:
```
ERROR:root:b"make: *** No rule to make target 'subsets/goslim_metagenomics.json', needed by 'all_subsets'.  Stop.\n"
```

Your fix revealed some deeper routed issue with the formats + extensions I think.. the mini-go test case is the only one doing subsets, so that's when it failed.

matentzn:
Superseded by recent updates to odk

Issue #338, opened 2020-05-09
Ontologies should be stamped with versionInfo


cmungall:
See https://github.com/ncbo/bioportal-project/issues/168

jonquet:
I would complement the issue by: 
- ontologies version information should be stamped with the  owl:versionInfo property
- ontologies version IRI or URI should be stamped with the owl:versionIRI property 

The first property can be used to describe the version with a string e.g., "3.0 (beta)" and the second is an IRI.

matentzn:
@jonquet Not sure I exactly understand what you are suggesting; can you provide an example? You mean an annotation assertion on an annotation assertion?

jonquet:
I was suggesting to use 2 properties one to descrive versionIRI and the other one to describe version. 
For instance: 
`
<http://purl.myorganization.org/myonto> 
	rdf:type 			owl:Ontology ;
	dct:title 			"My super cool ontology"@en;
	owl:ontologyIRI	http://purl.myorganization.org/myonto ;
	owl:versionIRI 		http://purl.myorganization.org/myonto-v2.3/
	dct:identifier		http://doi.org/anexternalidentifier ;
	owl:versionInfo	"2.3" ;
	omv:status		"beta" ;
	owl:deprecated	false ;
	owl:priorVersion	http://purl.myorganization.org/myonto-v2.2/
.
`
I allow myself to include a few other metadata annotation properties that I think are also useful to track versioning of ontologies.

balhoff:
I like the idea of using `owl:priorVersion`.

matentzn:
Ok ODK currently encourages (add templates for) these:

```
<owl:Ontology rdf:about="http://purl.obolibrary.org/obo/hp/hp-base.owl">
        <owl:versionIRI rdf:resource="http://purl.obolibrary.org/obo/hp/releases/2020-03-27/hp-base.owl"/>
        <dc:description>The Human Phenotype Ontology (HPO) provides a standardized vocabulary of phenotypic abnormalities and clinical features encountered in human disease.</dc:description>
        <dc:title>Human Phenotype Ontology</dc:title>
        <terms:license>https://hpo.jax.org/app/license</terms:license>
    </owl:Ontology>
```

So we can think about the rest:

[...]

PR #341, opened 2020-05-14
Add version info

matentzn:
This PR:
- makes VERSION a variable that is by OBO default the $(TODAY), but can be overwritten to be anything (like "2.3").
- adds owl:versionInfo to all release files
- moves version info definition into a single extensible variable (ANNOTATE_ONTOLOGY_VERSION)

#338

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -72,7 +72,7 @@ docker-publish: docker-build
 	@docker push $(IM):$(VERSION) \
 	&& docker push $(IM):latest
 
-docker-test: docker-build-use-cache
+docker-test: docker-build
 	docker images | grep odkfull &&\
 	make test CMD=./seed-via-docker.sh
 
--- a/template/src/ontology/Makefile.jinja2
+++ b/template/src/ontology/Makefile.jinja2
@@ -39,6 +39,8 @@ ODK_VERSION =               {% if env is defined %}{{env['ODK_VERSION'] or "Unkn
 
 TODAY ?= $(shell date +%Y-%m-%d)
 OBODATE ?= $(shell date +'%d:%m:%Y %H:%M')
+VERSION=$(TODAY)
+ANNOTATE_ONTOLOGY_VERSION = annotate -V $(ONTBASE)/releases/$(VERSION)/$@.owl --annotation owl:versionInfo $(VERSION)
 
 {%- if project.use_dosdps %}
 PATTERNDIR=                 ../patterns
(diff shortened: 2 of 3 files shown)


matentzn:
Cheers Jim, thanks for letting me know.. Looking at this, owl:versionInfo does not seem to be mapped:

http://owlcollab.github.io/oboformat/doc/obo-syntax.html#5.8

Why it however gets forced into the obo namespace.. Strange! This problem replicates with ROBOT:

Starting from test_version_info.owl:

```
<owl:Ontology rdf:about="http://www.semanticweb.org/matentzn/ontologies/2020/5/untitled-ontology-1235">
        <owl:versionInfo rdf:datatype="http://www.w3.org/2001/XMLSchema#decimal">1.4</owl:versionInfo>
    </owl:Ontology>
```

Converting to obo (`robot convert -i test_version_info.owl -f obo -o test_version_info.obo`)

```
format-version: 1.2
ontology: http://www.semanticweb.org/matentzn/ontologies/2020/5/untitled-ontology-1235
property_value: owl:versionInfo "1.4" xsd:decimal
```

Then running:

```
robot convert -i test_version_info.obo -o test_version_info_from_obo.owl
```

test_version_info_from_obo.owl:
```
<owl:Ontology rdf:about="http://www.semanticweb.org/matentzn/ontologies/2020/5/untitled-ontology-1235">
        <obo:owl_versionInfo rdf:datatype="http://www.w3.org/2001/XMLSchema#decimal">1.4</obo:owl_versionInfo>
        <oboInOwl:hasOBOFormatVersion rdf:datatype="http://www.w3.org/2001/XMLSchema#string">1.2</oboInOwl:hasOBOFormatVersion>
    </owl:Ontology>
```

Then:
```
odk robot convert -i test_version_info_from_obo.owl -f obo -o test_version_info_from_obo_from_owl.obo
```

Resulting back in:

```
format-version: 1.2
ontology: http://www.semanticweb.org/matentzn/ontologies/2020/5/untitled-ontology-1235
property_value: owl:versionInfo "1.4" xsd:decimal
```

So it seems that the OBO serialiser/converter does not handle the owl namespace correctly! So this feature will only work for two cases:
1. OBO format is the end product
2. OWL is the edit format and there is no OBO intermediate

Very unfortunate!

balhoff:
@matentzn I opened an issue: https://github.com/owlcs/owlapi/issues/947

matentzn:
Thats great @balhoff thanks! Although it will take a few months for any such fix to go through ROBOT etc (especially if we are waiting for Protege).. I guess we will simply live with it? Or hold of on merging until this is sorted?

balhoff:
I guess if the ODK workflow doesn't read OBO, it's not a big problem.

matentzn:
The default workflow only writes obo, but of course the edit file could be OBO; I would say though that this is independent of this here issue, because this PR really just tags the final release artefact.

cmungall:
Exactly - should not be a problem, no need to keep the versionInfo in the
edit file, and for release files it can be stamped

PR #348, opened 2020-07-01
added java opts ENV to run.sh

matentzn:
This is so that the memory for dosdp can be controlled as well. I chose to use the same configuration for this as the robot_java_args - Its hard to envision a case where anyone would want to restrict the one but not the other..

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -72,7 +72,7 @@ docker-publish: docker-build
 	@docker push $(IM):$(VERSION) \
 	&& docker push $(IM):latest
 
-docker-test: docker-build-use-cache
+docker-test: docker-build
 	docker images | grep odkfull &&\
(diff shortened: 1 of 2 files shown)


balhoff:
Would it be hard to change the setting to something more general like `project.java_memory_settings`?

matentzn:
The problem is all those already existing odk yaml files! a change like this would not be downward compatible.. I see this solution lacks a bit of artistic grace, but since its unlikely most of our users would ever use dosdp entirely without robot, I think its good enough - and we can always separate it out again.. Do you think its a problem? Would you rather we added `java_memory_settings` and kept that separate from robot memory settings?

balhoff:
That makes sense, I think it's fine to leave it how you have it.

cmungall:
whatever you think best!

PR #354, opened 2020-08-07
Use proper versions for `fastobo-validator` in Docker image

althonos:
Hi! 

I recently updated `fastobo-validator` but I can't get Bintray to have the latest version it in the `stable` release, which meant the URL in the `Dockerfile` was downloading an outdated tarball. Since all the other tools in the Docker image are versioned, I also made `fastobo-validator` versioned and made sure to only download that one version.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -63,13 +63,14 @@ RUN wget $ROBOT_JAR -O /tools/robot.jar && \
     chmod +x /tools/robot && \
     chmod +x /tools/robot.jar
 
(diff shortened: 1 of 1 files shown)


matentzn:
@althonos I merged it and patched yesterdays ODK release, so the new version should also now be available in the image.

althonos:
Great news! New version in particular will be fine with full-line, comments, which was not the case before, albeit them being syntactically valid.

PR #362, opened 2020-10-21
Always label in reports

cmungall:


matentzn:
I will take care of that in a different (configurable) way elsewhere. thank you!

PR #369, opened 2020-10-26
Update ROBOT version

balhoff:
This ROBOT includes whelk.


matentzn:
Is there an immediate use case that requires a patch release? I am working on a few features in ODK, but it could take a few weeks..

balhoff:
We have a use for whelk in GO, but just sometime in the near future would be good.

matentzn:
Already done as part of another PR stay tuned. Thank you!

PR #372, opened 2020-11-03
added pandas to requirements

### LINKED ISSUE #371: Add Python pandas to docker
Useful for scripted generation of robot templates.

Looking at [Dockfile](https://github.com/INCATools/ontology-development-kit/blob/master/Dockerfile) it seems that it should be sufficient to extend [requirements.txt](https://github.com/INCATools/ontology-development-kit/blob/master/requirements.txt)

dosumis:
Fixes #371

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,3 +6,4 @@ dataclasses >= 0.0
 dataclasses-json >= 0.0
 dataclasses-jsonschema >= 0.0
 PyGithub >= 0.0
+pandas


matentzn:
Pandas is already in ODK:

https://github.com/INCATools/ontology-development-kit/blob/c5b94f783885715e00df68339fc3713b84cbcfff/Dockerfile#L33

But i will merge this anyways, because all the python dependencies should be in the requirements txt as well. Thanks you!

matentzn (approved):

PR #379, opened 2020-12-01
Bump `fastobo-validator` to v0.4.0

althonos:
The new release:

* Adds support for line comments (i.e. comments that span on their own line, as opposed to comments at the end of a clause line, which was already supported).
* Adds a additional check (disabled by default) to ensure obsoletion clauses (`replaced_by` and `consider`) only appear in obsolete entities (`-O` flag to the CLI).


matentzn (approved):
Thanks a ton @althonos

Issue #403, opened 2021-03-25
Add relation-graph and steps to make an ubergraph-style triplestore


cmungall:
This is very useful https://github.com/balhoff/relation-graph

I'm not sure if it should be added to odk or directly to robot, cc @balhoff 

it is incredibly useful to get the inferred relational graph. I may even consider distributing this with some of my ontologies (though it can be large).

if RG is added then we can add some of the steps from here https://github.com/NCATS-Tangerine/ubergraph/blob/master/Makefile

This will give every ontology developer a useful triplestore stack with entailed inferences. This is especially nice for making importer/application ontologies using ODK.

Broader context:

https://docs.google.com/document/d/1Nr6mon6FEZAyvkAKC5OFmTLk5uANWzhoXUjxZVmOEJk/edit#

matentzn:
- [X] Add relationgraph #529 

Lets make more differentiated tickets for the triple store question.. Feels like a mild degree of scope creep.

PR #410, opened 2021-04-06
WIP: relation-graph #403

cmungall:
@balhoff is the path I am on correct? Should we try and inject UG configs into ODK, or have UG be something that sits above?


matentzn:
If you get this to work that would be amazing

cmungall:
@balhoff @matentzn should we try and build RG with sbt as part of the Dockerfile, or would it make more sense to build the executables as part of RG releases?

balhoff:
I publish a pre-built executable at the RG repo: https://github.com/balhoff/relation-graph/releases/tag/v1.1

matentzn:
Isnt sbt already fully in ODK? we could install it there, I don't think this is a problem!

matentzn:
Obviously pre-built should be included no matter what.

matentzn:
Superceded by #529

Issue #413, opened 2021-04-21
M1 compatible images?


matentzn:
We have not tried to diagnose the problem, but there seem to be issues running ODK on M1's. @gouttegd is trying to build ODK for M1 now and running it - if that works we will push an m1 release to dockerhub for those in need.

gouttegd:
So there's (at least) 3 different issues to build ODK for the M1 architecture.

### matplotlib
Matplotlib cannot be installed because some of its compile-time dependencies (namely `libfreetype6` and `libpng`) are missing. I guess the problem only occurs when targeting M1 because for other architectures, `pip` is able to download a pre-compiled binary package from PYPI -- but there's no pre-compiled binary package for arm64.

This is easily solved by manually installing the required dependencies.

### The obolibrary/robot image
This image cannot be built at all because it is based on the `openjdk:8-slim` image, which apparently is not available for arm64.

Unless/until someone creates a `openjdk:8-slim` image targeting arm64, an easy fix that seems to work is to base the image on `ubuntu:18.04` and manually install the OpenJDK (as is already done for the main `obolibrary/odkfull` image).

### SWI-Prolog
The `swipl` step fails with the following error:

```
ERROR: -g Opts=[interactive(false)],pack_install(dcgutils,Opts),pack_install(obo_metadata,Opts),pack_install(index_util,Opts),pack_install(regex,Opts),pack_install(typedef,Opts),halt: url `'https://www.swi-prolog.org/pack/query'' does not exist (status(500,Internal Server Error))
```

This is a [known error](https://github.com/SWI-Prolog/swipl-devel/issues/482) caused by the use of an old version of SWI-Prolog (< 8.0.3). Indeed, the version of SWI-Prolog inside the container is older than what we want:

```
swipl --version
SWI-Prolog version 7.6.4 for arm64
```

This is because the PPA repository for SWI-Prolog does not provide arm64 packages. Therefore, when we install `switch-prolog`, we get the package provided by Ubuntu, which is stuck to an older version.

I don't think there is an easy fix for that one. I am currently trying to build swipl from source from within the container, but this is obviously a pretty intrusive change (there's a bunch of dependencies to add just to compile it).

matentzn:
@gouttegd Thanks a lot for the detailed analysis! Awesome!

gouttegd:
With the release of ODK 1.2.29, M1-powered machines are now fully and officially supported. Closing here.

PR #414, opened 2021-04-21
Fix build for arm64 architecture

gouttegd:
Build a M1-compatible image is currently not possible due to several issues (detailed in #413) -- mostly, the fact that some packages are not currently available as pre-compiled binary packages for the arm64 architecture.

This PR addresses those issues by:
* ensuring `matplotlib` can be compiled from source if a binary package is not available;
* changing the `obolibrary/robot` image so that it does not derive from the `openjdk:8-slim` image, which is not available for arm64;
* building SWI-Prolog from source if we're building for arm64.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -51,45 +51,51 @@ RUN apt-get update &&\
   	&& rm -r /root/.cache
 
 
-### 3. Install SWI-Prolog from source
-# First install the build-time dependencies
-RUN apt-get install -y \
-    cmake \
-    ncurses-dev \
-    libreadline-dev \
-    libedit-dev \
-    libgoogle-perftools-dev \
-    libunwind-dev \
-    libgmp-dev \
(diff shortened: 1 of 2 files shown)


matentzn:
@gouttegd This is pretty amazing, thank you. I am running some tests now to see whether the image behaves ok on intel. Did you try the FBBT build? Did it work with that image?

gouttegd:
@matentzn I tried the `make docker-test`, which worked fine.

As for the FBBT build, it needed some further tweaking on our side, because at some point we run a Python script that requires `psycopg2`. We have a custom step in our Makefile that install `psycopg2` through `pip`, but that does not work on M1 because, again, there is no binary package on PYPI for arm64. So I tweaked our Makefile to have it install `psycopg2` through `apt-get` instead. And with that... it worked!

cmungall (approved):

Issue #421, opened 2021-05-06
Experiment with Multiarch images


matentzn:
@gouttegd recommended this blog post: https://www.docker.com/blog/multi-arch-build-and-images-the-simple-way

I think its worth exploring `buildx` and multi-arch builds. @gouttegd should you ever be interested to try this, I would be interested how this works. 

So all images, for all platforms exist at the same tag, but somehow docker fetches the right one depending on the infrastructure of the client? Is this also how you understand it?

gouttegd:
> So all images, for all platforms exist at the same tag, but somehow docker fetches the right one depending on the infrastructure of the client? Is this also how you understand it?

Yes. For an example, look at [this image for ubuntu/nginx](https://hub.docker.com/r/ubuntu/nginx/tags?page=1&ordering=last_updated). For each tag, there are several versions of the image for 3 or 4 different platforms (`amd64`, `arm64`, `ppc64`...). If I `pull ubuntu/nginx` from my M1-powered machine, I will get the `arm64` image, whereas someone pulling from a x86_64 machine will get the `amd64` image.

From my understanding of the blog post I mentioned, this is what `buildx` should allow us to do.

gouttegd:
So I started playing with `buildx`, and here are a few observations:

1. Making it work to build multi-arch images requires a bit of tweaking of one's Docker installation. Nothing very complicated (creating a *builder* instance and using it as the default builder), but it’s probably reason enough not to use multi-arch build by default. That is, `make build` should still default to building an image for the local platform.

2. My understanding is that even after a multi-arch image has been built, `make docker-test` will only test the image for the local architecture, and I don’t think there is anything we can do about that. `buildx` makes it possible to build a arm64 image on a x86_64 machine, but the generated arm64 image cannot be *run* on the x86_64 machine and therefore cannot be tested (and the other way around: I can build a x86_64 image on my M1 but I cannot run it).

This creates the possibility that we release an image that works fine on one platform, but is broken on another, because whoever does the release and pushes the image on Docker Hub will only have been able to test for one platform. If we want to avoid that, before any release someone should take care of building the image on arm64, run the tests, and report/fix any issue. I would be happy to do that of course.

3. The `buildx build` command can take care of pushing the multi-arch to Docker Hub after generating it, and this seems the easier way to do it (I still need to explore ways). In that case we would need to change our `docker-publish` target.

matentzn:
[...]

PR #422, opened 2021-05-13
Overhaul the Makefiles and support multi-arch builds

gouttegd:
This is a proposal to add support for building multi-arch images as discussed in #421.

This PR adds a `publish-multiarch` rule to build multi-arch images (for x86_64 and arm64, for now) for all products (*odkfull*, *odklite*, *robot*) and push them on Docker Hub. A section in the README gives some details about the preliminary steps that are required on a developer's machine to enable multi-arch builds.

In addition, the Makefiles have been streamlined to reduce code duplication between rules (e.g. between `build` and `build-no-cache` rules) and to make a better use of Make's features (e.g. call `make -C docker/odklite` to build *odklite* instead of `cd` manually into the `docker/odklite` directory).

(It is still possible to use the top-level Makefile as before to build mono-arch images. `make build` will still build an image for the local architecture and `make publish` will publish that image to Docker Hub as before.)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,10 @@ CMD = ./odk/odk.py seed
 
 EMAIL_ARGS=
 
+CACHE=
+
+PLATFORMS=linux/amd64,linux/arm64
+
 .PHONY: .FORCE
 
 custom_tests: test_no_yaml_dependencies_none test_no_yaml_dependencies_ro_pato test_no_yaml_dependencies_ro_pato_cl test_go_mini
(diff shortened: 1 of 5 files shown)


matentzn:
WOW! This is amazing!!

I didnt have time yet to look in detail but what I see is great :) So cool! Thank you, thank you!

matentzn:
(accidental close)

gouttegd:
> WOW! This is amazing!!

Except that it may not have been the best of ideas to change the rule names to remove the `docker-` prefix, that's why the CI check failed, because I stupidly forgot that the CI check relied on the existence of the `docker-test` rule. Sorry about that, I'll fix that and update the PR.

matentzn:
No wait.

matentzn:
The build is totally broken - we basically have to move to github actions anyways, it does not matter what the commands are called in this case.

matentzn:
I could not figure out how to build an image with Github actions, and then run something in that same built image.. :/ so at the moment I rely on local tests.

gouttegd:
Ah, that's good to know, thanks. Well, for what it's worth, the local tests still run fine for me with the above changes applied. :)

matentzn:
closed in favour of https://github.com/INCATools/ontology-development-kit/pull/429

PR #423, opened 2021-05-13
GitHub actions for deploying multi-arch built on dockerhub

matentzn:
@gouttegd I only started with this, but I don't really know how to properly do it.. 

Not sure if you know anything about github actions.. The idea would be to achieve the exact same as if running make docker-test locally..

### DIFF
--- a/.github/workflows/qc.yml
+++ b/.github/workflows/qc.yml
@@ -1,0 +1,33 @@
+name: ODK Continuous Integration
+
+on:
+  push:
+    branches:
+      - 'master'
+
+jobs:
+  docker:
+    runs-on: ubuntu-latest
+    steps:
+      -
+        name: Checkout
+        uses: actions/checkout@v2
+      -
+        name: Set up QEMU
+        uses: docker/setup-qemu-action@v1
+      -
+        name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+      -
+        name: Login to DockerHub
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+      -
+        name: Build and push
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          push: true
+          tags: obolibrary/odkdev


gouttegd:
Unfortunately as of now I know *nothing* about GitHub Actions beyond the fact that they exists... :/ But I can try at some point, if I can be of any help.

gouttegd:
@matentzn If you want to have a look at gouttegd/ontology-development-kit#1 and especially at the [.github/workflows/build-image.yml](https://github.com/gouttegd/ontology-development-kit/pull/1/files#diff-8b50c1c8ac07bc956388add66e256e8336ae22fd2cac8d83f770e1dc63674c00), I have something that seems to work.

(Please do *not* pull, as my test branch is not intended to be merged – it contains commits that have nothing to do with running Github Actions).

matentzn:
Wow, this works? its so weird that ubuntu-latest has docker installed so it can build this! Thank soo much for trying that!

gouttegd:
Yeah, what they call "ubuntu-latest" is not a "vanilla" Ubuntu distribution, [it comes with a bunch of pre-installed software](https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md) for development, including Docker. It makes things quite easy, really. We can just basically translate the travis file into a GA workflow file and (almost) everything works the same.

The only thing that doesn't work is that we can't run an *interactive* container, so I had to edit the `seed-via-docker.sh` script to remove the `-ti` options (otherwise the tests would immediately fail for lack of a TTY). It should not be an issue since the tests are *not* interactive anyway.

I'll soon send a proper PR.

matentzn:
Awesome I didn't know any of that! Thanks for doing it :) looking forward to finally having QC up again!

gouttegd:
@matentzn So I [tested](https://github.com/gouttegd/ontology-development-kit/actions/runs/857481691) creating a GA workflow to automatically build a multi-arch image upon publishing a release, and pushing those images to Docker Hub (in the `obotools` namespace, since I cannot push anything to the real `obolibrary` namespace).

Building the arm64 image on GA is painstakingly *slow* (it took nearly 3 hours, whereas building the amd64 image is done in barely 10 minutes), but it works. You can check the resulting images on the [Docker Hub page](https://hub.docker.com/u/obotools) for the `obotools` namespace.

I have no idea why building for arm64 is so slow (maybe the virtual machines running the GA workflows are somehow not tailored for this task – which seems strange, but in any case that is in GitHub’s hands, I don’t think we can do anything about it), but I guess we don’t really care as the job is done on GitHub’s servers – no developers’ machine is immobilized during the process.

The only potential issue is that, with a free GitHub account, we are limited to 2000 minutes of GA runtime per month, so we have to take care of not exceeding that. But I think we have quite a margin:
* it takes ~10 minutes to run the test suite upon each pull request;
* it takes ~180 minutes to build the images upon each release.

So we have enough runtime for ~200 pull requests or ~10 releases per month… I don’t think we’ll ever get close to such numbers. :)

matentzn:
I would say this is reasonable; Lets say we do one release per month, there is still plenty of time around the edges. I am wondering now: Should we try and long term fully automate the entire process? Something like this:

1. Test on push/pull request (done)
2. Every 1st of the month (?), run a release

I am a bit worried to push broken images (broken in the sense of, some untested thing in there somehow gets tagged with `latest` and breaks everyone's pipeline), but maybe that is exaggerated. But you have almost made this possible now. These are the steps for release:

1. Build push dockerhub
2. Create tagged github release with release notes that describe what happened since last release

For the release notes, I was typically curating the [Changes.md](https://github.com/INCATools/ontology-development-kit/blob/master/Changes.md) manually, so maybe a smart use of `git diff` on that file could serve as input to release notes on the GitHub release.

I don't know, what are your thoughts?

matentzn:
I think this echos my feelings as well, lets keep this manual. 

As for the `docker-test`, great catch (the 60 min is also what I am used to). I am mostly concerned about the fact that every push (commit) triggers the tests - so its not really on a pull - its on every single commit to that pull as well..

gouttegd:
We can [fine-tune](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request) what triggers the workflow. For example, we could run the tests only when a review is explicitly requested on a pull request, instead of doing so every time the PR is updated with new commits.

matentzn:
closed in favour of https://github.com/INCATools/ontology-development-kit/pull/429

PR #425, opened 2021-05-18
Do not use the docker-dash branch of OBO-Dashboard.

gouttegd:
The docker-dash branch of OBOFoundry/OBO-Dashboard has been deleted
after being merged into master [1], thereby breaking the ODK build
because we explicitly check out that branch.

Fix the build by checking out OBO-Dashboard's master branch.

[1] https://github.com/OBOFoundry/OBO-Dashboard/pull/24


gouttegd:
Never mind, I didn't notice this is already fixed in the PR #424  for the upcoming ODK 1.2.29...

matentzn:
Thanks for noticing again :) yeah, I should be more diligent with feature branches but there is just too much stuff going on, it's basically like a dev branch model now..

PR #426, opened 2021-05-19
Create qc.yml

matentzn:


gouttegd:
For now, it fails because the ODK cannot be built since the deletion of the `docker-dash` branch of OBO-Dashboard. That will be automatically fixed when PR #424 is merged.

But as I mentioned moments ago in #423, after the images have been built, the tests themselves will *still* fail because we cannot run *interactive* containers in GA. You need to remove the `-ti` option in the `seed-via-docker.sh` script.

matentzn:
I was just curious :) will close this and wait for your PR instead!

PR #429, opened 2021-05-20
General Makefile update and support for building multi-arch images

gouttegd:
This PR replaces both PR #422 and #423.

It provides:
* streamlined Makefile rules [1];
* support for building multi-arch images (x86_64 and arm64) with `buildx`;
* automatic building and publishing of images upon release [2];
* the possibility to manually trigger the GitHub Actions workflows if necessary.

It also fixes running the tests in situations where the user calling `make tests` does not have a `~/.gitconfig` file.

[1] As part of this change, `make docker-build` becomes `make build` and `make docker-test` becomes `make tests`; the README has been updated accordingly to provide up-to-date instructions.

[2] Automatically publishing the images requires the repository to be configured with two _secrets_ for the username and password for the `obolibrary` Docker account (`DOCKER_USERNAME` and `DOCKER_PASSWORD`).


.travis.yml
```diff
@@ -7,7 +7,7 @@ services:
 # https://github.com/INCATools/ontology-starter-kit/issues/25
 # EMAIL_ARGS='-e obo-ci-reports-all@groups.io'
 script:
-  - make docker-test
+  - make tests
```
matentzn, line 10: travis.yml should be deleted altogether

docker/odklite/Makefile
```diff
@@ -1,20 +1,30 @@
 
 # Building docker image
-VERSION = "v1.2.27" 
+VERSION = "v1.2.27"
```
matentzn, line 3: the VERSION variable should probably be delted and taken from the main Makefile as an ARG?

docker/robot/Makefile
```diff
@@ -1,16 +1,32 @@
 
 # Building docker image
-VERSION = "v1.8.1" 
+VERSION = "v1.8.1"
```
matentzn, line 3: This variable should stay as is, because the Version will be in step with the ROBOT, not ODK, versions.

matentzn (commented):
Just some comments, but will try to test soon! Thanks _a ton_ for this! Great work!

.travis.yml
gouttegd: I was not sure about that since I have never used Travis at all. Is deleting that file enough to cut any link with Travis?

docker/odklite/Makefile
gouttegd: When the `docker/odklite/Makefile` is called from the top-level Makefile, then the top-level Makefile provides the version by overriding that variable (in other words, the `VERSION` variable in the top-level Makefile has precedence over the same variable in the subfolder Makefile).

The rationale for keeping the variable defined in the subfolder Makefile is that it allows to use that Makefile in a "standalone" way, independently of the top-level Makefile. If someone, for any reason, is only interested in building _odklite_, they can do so by going into the `docker/odklite` folder and calling `make` there.

Of course, the version number in the `odklite` Makefile should be kept "in sync" with the version number in the top-level Makefile (otherwise `odklite` would have a different version number depending on whether it has been built from its own Makefile or from the top-level Makefile), which is currently not the case. But we can update that at anytime.

docker/robot/Makefile
gouttegd: Agreed. Here I just removed the trailing whitespace – just to satisfy my OCD about this kind of things...

docker/odklite/Makefile
matentzn: Perfect thank you, this makes sense.

docker/robot/Makefile
matentzn: Yeah I didnt mean to say anything about the whitespace, just to mention that the versioning for the ROBOT image works somehow different from the versioning for the others! Thank you!

.travis.yml
matentzn: I will cut ties with travis entirely now, you can delete the file.

PR #440, opened 2021-06-10
Fix running the test suite

gouttegd:
The `seed-via-docker.sh` script, used to start the tests, hardcodes the name of the Docker image to use (`obolibrary/odkfull`). As a result, the Makefile rules intended to test a specific image (`test-full` to test `odkfull`, `test-dev` to test `odkdev`, `test-lite` to test `odklite`) do not work as advertised, since no matter which rule is invoked it's always the `odkfull` image that ends up being tested.

This PR fixes that by allowing to change the image used by the `seed-via-docker.sh` script.

(Of note, the `seed-via-docker.bat` script has _not_ been similarly updated, because it's been too long since I last used Windows to remember how variables work in a `.bat` script...)

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -21,4 +21,4 @@ jobs:
       - name: Build the images
         run: make build
       - name: Run the tests
-        run: make tests
+        run: make test-full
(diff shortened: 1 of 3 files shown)


matentzn (approved):

matentzn:
re the bat file, i was thinking about that, but it does not matter in my opinion. No windows user will use anything but odkfull.

My only just triple check question: `${IMAGE:-odkfull}` is legal sh, right, does not need `bash`?

PR #452, opened 2021-08-09
Update ecto-odk.yaml

cmungall:


matentzn:
I redid this PR in the right place.

https://github.com/EnvironmentOntology/environmental-exposure-ontology/pull/200

Note that NPO does not disappear this way - in only stops being updated.

Issue #466, opened 2021-09-10
Consider using multi-stage builds


gouttegd:
[Multi-stage builds](https://docs.docker.com/develop/develop-images/multistage-build/) are a method to install into a Docker image files that have been prepared inside another Docker image.

Among other things, it allows to build a program in a (temporary) Docker image and copy *only* the resulting program to another (final) image, leaving all the build-time dependencies (that are only required to build the program) out of the final image. It should therefore allow to considerably reduce the size of an image.

matentzn:
That would be so awesome..

gouttegd:
Done with #475.

Issue #459, opened 2021-09-07
Include soufflé?


balhoff:
I've been finding [soufflé](https://souffle-lang.github.io/index.html) incredibly useful lately, and would like to incorporate it into some workflows (for Ubergraph and for GO). Would it be too specialized to include in the ODK docker image?

They have Ubuntu install instructions here: https://souffle-lang.github.io/install  Will that work for Apple M1?

matentzn:
I dont see why not! Feel free to make a PR; this does not look too bad to me..

gouttegd:
Soufflé has been added to `odkfull` as part of the overhaul of the build process in PR #475.

balhoff:
Thanks so much @gouttegd!

PR #467, opened 2021-09-11
Install Soufflé using a multi-stage build

gouttegd:
This PR uses the problem of adding Soufflé to the ODK (#459) as an opportunity to test multi-stage builds (#466).

Soufflé is built on an intermediate image containing all its required build-time dependencies, and only the compiled program and its run-time dependencies are installed to the final ODK image. With all the build-time dependencies left out of the final image, the size increase due to the addition of Soufflé is minimal (~15 Mb).

@balhoff Can you build the ODK from this PR and test that Soufflé is working as you expect? Or provide some tests that I could use to check myself?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -186,7 +186,26 @@ RUN cd /tools/ && chmod +x /tools/obodash && git clone --depth 1 https://github.
 
 
 ###### Souffle ######
-RUN curl -s https://packagecloud.io/install/repositories/souffle-lang/souffle/script.deb.sh | bash && apt-get install -y souffle
+RUN apt-get install -y \
+    bison \
+    clang \
+    cmake \
+    doxygen \
+    flex \
+    g++ \
+    libffi-dev \
+    libncurses5-dev \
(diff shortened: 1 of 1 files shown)


balhoff:
Wow, thank you so much @gouttegd! I will give it a try.

balhoff:
@gouttegd here is the result:

```
jim (master)$ docker run -m 12g  -v $PWD:/work -w /work --rm -ti obolibrary/odkfull souffle -c phenoscape-query.dl
terminate called after throwing an instance of 'std::runtime_error'
  what():  failed to locate mcpp pre-processor
```

gouttegd:
Thanks @balhoff .

I'll push an update with the run-time dependencies suggested by Soufflé's Debian packager (I should have done that from the start…).

balhoff:
@gouttegd any thoughts on this comment? https://github.com/souffle-lang/souffle/discussions/2065#discussioncomment-1330314

gouttegd:
Closing as Soufflé is now installed as part of #475.

PR #471, opened 2021-09-18
Overhaul odklite

gouttegd:
This PR proposes to update the `odklite` image so that: (i) it contains the minimal set of tools absolutely required for a "standard" ODK-generated repository (that is, all the commands in the ODK-generated `src/ontology/Makefile` can be run), and (ii) the image is as small as possible.

To achieve (i), the main tools installed in the image include:
- the JRE,
- ROBOT, DOSDPTOOLS, and OWLTOOLS,
- Ammonite (needed for the `validate_idranges` target),
- Python3 with the minimal set of packages needed by the ODK scripts,
- other tools such as git, make, jq, rsync.

To minimize the size (ii) :
- we install the _headless_ variant of the JRE (so that we don't bring an entire X window system),
- we forbid `apt-get` from installing recommended packages (so that, for example, installing `pip` does not bring a C++ compiler).

The resulting image is ~670 Mb large, compared to the ~2.5 Gb of the `odkfull` image. It has been successfully tested against FlyBase's _Drosophila_ Anatomy Ontology (FBbt).

### DIFF
--- a/docker/odklite/Dockerfile
+++ b/docker/odklite/Dockerfile
@@ -1,61 +1,79 @@
-### From https://stackoverflow.com/questions/51121875/how-to-run-docker-with-python-and-java
-### 1. Get Linux
 FROM ubuntu:20.04
-LABEL maintainer="obo-tools@googlegroups.com" 
+LABEL maintainer="obo-tools@googlegroups.com"
 
-### 2. Get Java, Python and all required system libraries (version control etc)
-ENV JAVA_HOME="/usr"
+ENV ROBOT_VERSION 1.8.1
+ENV DOSDP_VERSION 0.17
+ENV OWLTOOLS_VERSION 2020-04-06
+ENV AMMONITE_VERSION 2.0.3
+
 WORKDIR /tools
-ENV PATH "/tools/:$PATH"
-COPY requirements.txt.lite /tools/requirements.txt
-COPY odk/make-release-assets.py /tools/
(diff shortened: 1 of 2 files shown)


matentzn (commented):
This is great.. I am wondering about the Ammonite dependency.. The command lives outside the general ODK workflows, (`validate_idranges` is not connected to any goals). I would be happy to drop the ammonite dependencies here to get some more space.

For pip install, shouldn't we go down the same route as with the main container? Having a requirements.txt file with fixed versions etc?

Beautiful!

gouttegd:
Do *not* merge!

This may be beautiful, but I realised when testing on my work M1 machine this morning that this does *not* work on M1! (For the same reason as usual: some Python packages are not available as pre-built binaries for arm64.) I will be overhauling the overhaul to get something that hopefully works just as fine on x86_64 and arm64… when I'll be done cursing.

For the Python packages, yes, we should also freeze versions in `odklite`.

Ultimately, the goal of this overhaul is to make the `odkfull` image a strict superset of `odklite`. That is, to build `odkfull` _on top of_ `odklite` (adding all the tools that we don't want in `odklite`), so that we don't have to worry about keeping `odklite` and `odkfull` in sync (they'll be in sync by construction).

matentzn:
No worries, I wouldnt in any case merge your PRs - just approve them so you can merge them yourself :)

Issue #472, opened 2021-09-20
Add tests for some of the external tools..


matentzn:
As per slack between @gouttegd and @Clare72, its possible that as the underlying architecture of the ODK are changing (system libraries etc), the behaviours of the tools change. We should consider building a manual test suite using make that invoke such tools (like oort, Konclude, scala scripts etc.).

gouttegd:
It would be nice if we could also make it so that the test suite does not depend on fetching ontologies on remote servers… Over the weekend I have not been able to successfully run the entire test suite even once because there's always a _connection time out_ when fetching something from purl.obolibrary.org…

Not sure if it is even feasible, though, but if not then we should at least allow the test suite to gracefully skip the current test and go on with the next one instead of aborting completely.

Clare72:
>Over the weekend I have not been able to successfully run the entire test suite even once because there's always a connection time out when fetching something from purl.obolibrary.org…

Is it chebi? I added retries/timeouts for this in dpo because I got fed up of it taking > 30 min going through the release just to fail when it had to download chebi:
```
mirror/chebi.owl: mirror/chebi.trigger
	curl -L http://purl.obolibrary.org/obo/chebi.owl.gz --create-dirs -o mirror/chebi.owl.gz --retry 4 --max-time 120 &&\
	$(ROBOT) convert -i mirror/chebi.owl.gz -o $@.tmp.owl &&\
	mv $@.tmp.owl $@
```

matentzn:
@Clare72 this is a great idea! We should implement that by default in ODK! Would you be able to make a PR reflecting that change?

gouttegd:
When running the tests for the ODK, from what I have seen it can be any ontology, really. Most often it seems to be purl.obolibrary.org/obo/ro/annotations.owl, but sometimes it's http://purl.obolibrary.org/obo/ro/bfo-axioms.owl, sometimes yet another one…

matentzn:
I think if we get that retry logic in there, the failures will go down to a negligible amount..

Clare72:
@matentzn I don't have write access!

This seems fairly straightforward for the last 4/5 variants of the ```mirror/{{ ont.id }}.owl``` goals, but for the ```{%- if ont.mirror_from %}``` we don't know whether it will be .owl or .owl.gz and robot needs to be given the correct extension. I am guessing that we would need to get the filename from the ONT-odk.yaml file and give this to curl for output and robot for input, but I am not sure how to do this.

[...]

PR #474, opened 2021-09-20
retries for mirrors

Clare72:
related to #472 
As mentioned on the ticket, I am not sure how to implement retries for the first variant.
Also we may want a longer timeout for some ontologies - I find 120s sufficient for chebi.owl.gz, but might be worth increasing this here.

### DIFF
--- a/template/src/ontology/Makefile.jinja2
+++ b/template/src/ontology/Makefile.jinja2
@@ -415,18 +415,18 @@ mirror/{{ ont.id }}.owl: mirror/{{ ont.id }}.trigger
 {%- elif ont.use_base %}
 {%- if ont.use_gzipped %}
 mirror/{{ ont.id }}.owl: mirror/{{ ont.id }}.trigger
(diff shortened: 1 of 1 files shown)


matentzn:
Can you remind me why a larger timeouty may be harmful? Say 1000 seconds? Is it just that it may get stuck and you will have to wait the full 4*1000 seconds before the fail?

Also, can you do a quick local test if ROBOT actually knows how to deal with local gzips? `robot merge -i x.owl.gz ..`.

Thank you so much!

Clare72:
Yes, that is why. I found that my machine would just keep waiting (seemingly indefinitely) without downloading anything and I would have to cancel manually using ```^-c```, then start again :-(
The code I pasted on the ticket from dpo.Makefile works for me for chebi.owl.gz, so yes, robot can handle this.

matentzn:
Thank you!

matentzn (approved):

Clare72:
You don't think we should increase the timeout?

gouttegd:
With your patch I have been able to run the full ODK test suite without the kind of errors I was systematically running into before, so it seems to do the trick for me. :+1: Thanks!

matentzn (approved):

matentzn:
I will deal with this before the next release - I want the parameter to be customisable, its easy enough to set this up. Thanks!

Issue #473, opened 2021-09-20
Konclude does not work in latest ODK


matentzn:

gouttegd:
Fixed by #475.

Issue #463, opened 2021-09-09
Consider freezing python versions in requirements.txt


matentzn:
This will allow us to a) circumvent the problems with the new resolved and b) create a more reproducible overall experience with the image.

See #461

gouttegd:
Done as part of #475.

matentzn:
Thanks @gouttegd

Issue #217, opened 2019-05-21
Shrink ODK image


matentzn:
- [ ] Remove GCC (needed for pandas/numpy)

matentzn:
Breakdown of large culprits (layers) in current dockerfile:

```
DOSDP: wget -nv https://github.com/INCATools/dosdp-tools.. 322MB
AMMONITE: (echo "#!/usr/bin/env sh" && curl -L https://github.com/lihaoyi/Ammonite/releases/download/2.0.3/2.13-2.0.3) >/tools/amm && chmod +x /tools/amm 31.2MB
ROBOT: wget $ROBOT_JAR -O /tools/robot.jar.. 114MB
KONCLUDE: wget https://github.com/konclude/Konclude.. 27.2MB
OWLTOOLS: wget http://build.berkeleybop.org/userContent/owltools/owltools.. 69.4MB
PYTHON LIBS: pip3 install -r requirements.txt && pip3 install... 174MB
SYSTEM LIBS: apt-get update && apt-get install -y python3-pip python3-dev subve.. 123MB
JAVA8: apt-get update && apt-get upgrade -y && apt-get install -y softwa.. 691MB
<missing> 7 days ago  #(nop) ADD file:08e718ed0796013f5957a1be7da3bef6225f3d82d8be0a86a7114e5caad50cbc in / 63.2MB
```

Any help appreciated @kltm @althonos @balhoff to identify major culprits.

balhoff:
Why is dosdp-tools taking 322MB? The release is about 70MB: https://github.com/INCATools/dosdp-tools/releases/tag/v0.14-beta.1

matentzn:
No idea! Is it because it is untarred?

```
###### DOSDPTOOLS ######
ENV PATH "/tools/dosdp-tools/bin:$PATH"
RUN wget -nv https://github.com/INCATools/dosdp-tools/releases/download/v$DOSDPVERSION/dosdp-tools-$DOSDPVERSION.tgz \
&& tar -zxvf dosdp-tools-$DOSDPVERSION.tgz \
&& mv dosdp-tools-$DOSDPVERSION /tools/dosdp-tools \
&& wget --no-check-certificate https://raw.githubusercontent.com/INCATools/dead_simple_owl_design_patterns/master/src/simple_pattern_tester.py -O /tools/simple_pattern_tester.py \
&& chmod +x /tools/*
```

jamesaoverton:
[...]

PR #475, opened 2021-09-21
Use a multi-stage build process.

gouttegd:
This PR completely overhauls the way both `odkfull` and `odklite` are built, in order to:

- reduce the size of each image (#217);
- on arm64, allow to install from source some programs that are not available as pre-compiled binaries (e.g. Soufflé, #459), without cluttering the final images with build-time dependencies.

Along the way, it also includes the proposed freeze of Python packages (#463), fixes the broken Konclude (#473), and adds some basic checks to ensure the various programs of the ODK are installed correctly (#472).

Basically, building `odkfull` is now a 3-steps process:

(1) First, an intermediate image called `odkbuild` is built. This image serves as a sort of "staging area", where we can install all the build-time dependencies we want (without worrying about image size) and build everything we need. This includes installing the Python modules, because some of them have to be compiled on arm64 for lack of pre-compiled arm64 wheels.

(2) Second, `odklite` is built, using some of the tricks proposed in #471 to keep the image size minimal (mostly forbidding `apt-get` from installing "recommended" packages and using a headless variant of the JRE). That image contains ROBOT, DOSDPTOOLS, OWLTOOLS, and Ammonite, along with the minimal set of Python modules required to run `odk.py`. It is intended that this image should be sufficient for use with a "standard" ODK-generated repository. Size of the image: ~650 Mb.

(3) Third, `odkfull` is built on top of `odklite` (which is now, by construction, a strict subset of `odkfull`), adding Konclude, Soufflé, Jena, SPARQL-prog, along with a richer set of Python modules and build tools. Size of the image: ~1.8 Gb.


matentzn:
This is so awesome, I cant contain myself. Running builds locally now!

constraints.txt
```diff
@@ -0,0 +1,239 @@
...
+mistune==0.8.4
+mkdocs==1.2.2
+mypy==0.910
+mypy-extensions==0.4.3
+nbclient==0.5.4
+nbconvert==6.1.0
+nbformat==5.1.3
+neo4jrestclient==2.1.1
+nest-asyncio==1.5.1
+networkx==2.6.3
```
matentzn, line 112: Its odd that this is here, and requirements.txt has a fixed `networkx` version

gouttegd: I thought I had removed all version numbers from the requirements.txt file… I must have done that in another branch and forgot to bring it to that branch… :facepalm:

matentzn: All the better then! I would like that, because the old networkx breaks several of my pipelines now.

gouttegd: I'll push an updated `requirements.txt` later today. Just re-building now to check that everything still works with the new `requirements.txt`. (I don't expect any breakage but…)

matentzn:
Ok I did a few checks locally.. I am wondering if more needs to be done. If you have no reasonable concerns, I am happy.

I:
1) rebuild multiple times
2) ran ODK on cell (ok)
3) ran ODK on dashboard (ok)

Happy to merge if yo uare.

gouttegd:
The one thing that still needs to be tested is building multi-arch images. It *should* work as expected (i.e., the `build-multiarch.yml` Github Action should still do the job automatically upon the next release), but there's still a possibility that the Github builders would somehow be unable to deal with multi-stage images. I didn't find any report suggesting that would be the case, but who knows…

If you agree, what I would like to do is to temporarily bump the version number to "1.2.30-alpha" and manually trigger the `build-multiarch` action on this PR, so that we can check that the multi-arch build process still works. (The rationale for adding an "-alpha" suffix to the version number is that the generated images will end up on Docker Hub – that's the whole point of the `build-multiarch` action after all –, so we need to make sure users know the new images are *not* an actual release.)

matentzn:
I am not worried about the -alpha; if the build succeeds and the image is pushed, then a `docker pull` will obtain that image, wether the tag says `alpha` or not, right?

gouttegd:
I was hoping there would be a way to mark an image on Docker Hub as "beta", "not-production-ready", or anything like that, but that does not seem to be the case. Scheiße.

The next best option is then to publish under the _obotools_ name  (where we initially published the M1-specific images when we were still figuring out how to build multi-arch images) instead of the _obolibrary_ name. I'll do that from my own repo, which is already set up for publishing in _obotools_.

matentzn:
Ok, if this works as well, great.. I would anyways like to aim for an ODK update soon, but want to deal with some of the smaller tickets first. Maybe first week of october.

gouttegd:
GitHub builders are so _slow_ … 5 hours in and it's still building the `odkbuild` image, which is the _first_ step of the build process…

gouttegd:
… And the build was forcibly cancelled because it was taking too long… There's a 360 minutes limit for each job, and building the `odkbuild` image takes longer than that on GitHub builders (mostly because of the need to build Konclude from source on arm64).

There are a few things we can try to circumvent this, such as creating _another_ intermediate image just to build Konclude (so that it is built in a different job, which would be subjected to its own time limit), but I am _very_ reluctant to make the build process even more complex than it already is just to accommodate GitHub's arbitrary limits… Maybe we should just go back to building the images-to-be-released on one of our own machines and forget the `build-multiarch` GitHub Action entirely.

matentzn:
It's a pity but I don't mind. I have servers etc that do this work for me! Whatever you think is best. I have no problem building locally and pushing!

gouttegd:
I'd say merge as it is then. At the next release, we can try again the automated build, and if it fails again (which is likely, unless in the meantime GitHub switches to faster machines for its builders), you can build and push the images yourself. Let me know if you need any help setting up your local builder for multi-arch builds, if you have not done so already.

gouttegd:
I've just pushed one last bit to remove intermediate object files from the `odkbuild` image. This is not strictly necessary but it reduces the size of `odkbuild` by up to 1 GB.

I am done tweaking things with that PR, I promise! :)

matentzn:
Thank you running tests now :) Is it ready to merge then? Are you happy? If my tests pass here, I will just merge it then.

gouttegd:
Still disappointed by how slow the GitHub builders are, and the subsequent inability to use the `build-multiarch` action, but yes, I am happy with that PR. :)

matentzn:
First fail in my third server:

```
Get:95 http://archive.ubuntu.com/ubuntu focal/universe amd64 dos2unix amd64 7.4.0-2 [374 kB]
E: Failed to fetch http://security.ubuntu.com/ubuntu/pool/main/l/linux/linux-libc-dev_5.4.0-84.94_amd64.deb  404  Not Found [IP: 91.189.88.152 80]
E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?
Fetched 69.8 MB in 7s (9826 kB/s)
The command '/bin/sh -c DEBIAN_FRONTEND="noninteractive" apt-get install -y --no-install-recommends     build-essential     openjdk-8-jdk-headless     maven     python3-dev     subversion     automake     aha     dos2unix     sqlite3     libjson-perl     pkg-config     xlsx2csv' returned a non-zero code: 100
Makefile:69: recipe for target 'build' failed
make: *** [build] Error 100
```

Lets see whether it was a fluke

matentzn:
Unfortunately it was not.. building the odklite does not work on my Ubunto machine fails..

```
NAME="Ubuntu"
VERSION="18.04.5 LTS (Bionic Beaver)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 18.04.5 LTS"
VERSION_ID="18.04"
HOME_URL="https://www.ubuntu.com/"
```

I am not too worried but so far, I was able to run the builds there.. What is your opinion?

gouttegd:
I can't reproduce the error here, whether on my MacOS machine or my Linux machine… What's the Docker version on your Ubuntu 18.04 machine?

matentzn:
I completely uninstalled and reinstalled docker:

```
Docker version 20.10.8, build 3967b7d
```

gouttegd:
So that's the same version than the one on my MacOS machine, and an even fresher version than the one on my Linux machine… Sorry, I have no idea what could cause the problem here. It looks as if `apt-get` was working from an obsolete list of packages, but the cache _is_ updated at the beginning of building the `odklite` image, so that shouldn't happen…

matentzn (approved):

matentzn:
Maybe its OS version related somehow in some weird way we don't understand. Lets merge it; works on two of my machines so its fine.

PR #476, opened 2021-09-21
Add mkdocs-materials to requirements.txt

matentzn:
@gouttegd this is not of critical importance right now, but I guess this begs the question on how to update constraints.txt when adding something to requirements.txt..


gouttegd:
What I did was basically:

```
$ virtualenv tmp-python-packages
$ source tmp-python-packages/bin/activate
(virtualenv) python3 -m pip install -r requirements.txt
[…]
(virtualenv) python3 -m pip freeze > constraints.txt
(virtualenv) deactivate
$ rm -rf tmp-python-packages
```

This doesn't need to be done _every time_ a package is added to `requirements.txt`, though. It can be done only once in a while, as long as it is done at least once before a release (unless no new packages have been added since the last release).

To be strict, this should be done from within a Docker container loaded with Ubuntu:20.04 (to be as close to the ODK as possible), but I think this is probably overkill…

PR #477, opened 2021-09-23
adding fix-missing param

matentzn:


gouttegd:
The `--add-missing` makes `apt-get` simply ignore the offending package, which according to your error message seems to be a security update. I am not too much concerned about that, since whatever security vulnerability the missing package is supposed to fix will be confined to the Docker container anyway.

I am more concerned by the fact that this is needed at all and I don't understand why. The list of packages is refreshed in the `odklite` image, so when we're building `odkfull` we should have the updated list (since `odkfull` is built on top of `odklite`) – and this is indeed what's happening on my machines, which makes me believe it could be a bug from an older version of Docker.

gouttegd (approved):
I don't like the fact that I don't understand why this is necessary, but this doesn't break anything on my side, so if it makes the build pass on your machine, go for it.

matentzn:
Maybe lets not merge it for now. I am afraid it would introduce an inconsistent behaviour when built on one platform vs. another.

matentzn:
@gouttegd can you check, I think now we cannot rebuild ODK without this fix - even on our Mac.. seems my ubuntu was just peeking into the future :P

gouttegd:
Remember that I can't push an image to _obolibrary_ on DockerHub! I don't have write access to that account.

PR #481, opened 2021-10-11
ensure_owl2dl_profile should be True by default

cmungall:
I can't think of any situation where we want to disable this; if it does exist it will be an advanced user who will be capable of making the change


matentzn:
I totally agree with this PR, but it will break many running pipelines.. Like so many, it will just cause mayhem and us being the ones that have to sort it out again. I guess we can do it with a big warning message.

matentzn:
I think merging this is very risky, but given its importance, ok.

PR #484, opened 2021-10-18
Build the obolibrary/robot image when publishing.

### LINKED ISSUE #483: Make sure that docker/robot/Dockerfile is build and pushed when running a multistage release
@gouttegd this is not urgent, but if you could pick it up that would be great. If not, also no problem, I can handle it.

gouttegd:
Make sure the `obolibrary/robot` image is built and pushed on DockerHub when building the other multi-arch images. This was the case before, but has been inadvertently lost during the overhaul of the build process in #475.

closes #483

### DIFF
--- a/docker/robot/Dockerfile
+++ b/docker/robot/Dockerfile
@@ -9,8 +9,8 @@ ENV ROBOT_JAR ${ROBOT_JAR}
 
 RUN apt-get update &&\
   apt-get upgrade -y &&\
-  apt-get install -y make \
-    openjdk-8-jdk \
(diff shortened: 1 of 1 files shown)


gouttegd:
Forbidding `apt` from installing “recommended” dependencies (`--no-install-recommends`) and switching from the full JDK to the headless variant of the JRE (`openjdk-8-jre-headless` instead of `openjdk-8-jdk`) could easily cut the size of `obolibrary/robot` to probably less than 300 Mb, and should still allow `robot` to work (`robot` does not need the full JDK and does not need any of the graphical component of the JRE).

PR #489, opened 2021-11-01
make ROBOT_JAR parametrisable

matentzn:
No time to deal with this now, but as remidner.

### DIFF
--- a/docker/odklite/Dockerfile
+++ b/docker/odklite/Dockerfile
@@ -15,6 +15,9 @@ ENV PATH "/tools:/tools/dosdp-tools/bin:$PATH"
 ARG ODK_VERSION 0.0.0
 ENV ODK_VERSION $ODK_VERSION
 
+ARG ROBOT_JAR https://github.com/ontodev/robot/releases/download/v$ROBOT_VERSION/robot.jar
+ENV ROBOT_JAR $ROBOT_JAR
+
 # Install base tools from Ubuntu.
 RUN apt-get update && \
     DEBIAN_FRONTEND="noninteractive" apt-get install -y --no-install-recommends \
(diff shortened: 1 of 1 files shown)


matentzn:
@gouttegd Should be ok right? this if for cases where we need to compile ODK with unreleased robot version.

gouttegd (approved):
Yes, looks good.

gouttegd:
Well, the CI check seems to disagree with me (thanks God for CI checks!).

Having a closer look now…

docker/odklite/Dockerfile
```diff
@@ -15,6 +15,9 @@ ENV PATH "/tools:/tools/dosdp-tools/bin:$PATH"
 ARG ODK_VERSION 0.0.0
 ENV ODK_VERSION $ODK_VERSION
 
+ARG ROBOT_JAR https://github.com/ontodev/robot/releases/download/v$ROBOT_VERSION/robot.jar
```
gouttegd, line 18: It should be `ARG ROBOT_JAR=https://…`

While the `ENV` command accepts a form without the equal sign, the `ARG` command [does not](https://docs.docker.com/engine/reference/builder/#arg).

(Actually even for the `ENV` command the equal-less form [is discouraged](https://docs.docker.com/engine/reference/builder/#env) and may disappear in the future, so it could be a good idea at some point to update all the Docker files to make sure we always use the `=` syntax.)

gouttegd (changes requested):

docker/odklite/Dockerfile
matentzn: Ok, for this dockerfile i changed that! Good to know!

gouttegd (approved):

PR #492, opened 2021-11-03
Move to Java 11.

balhoff:
I propose we move to Java 11. It's been out as the LTS version for 3 years. Some libraries are starting to drop compatibility with Java 8 (e.g., Jena 4.x). In my experience Java 11 has better performance as well. Making the change here doesn't force any particular tool (like ROBOT) to drop support for Java 8.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,7 +15,7 @@ ENV COURSIER_CACHE "/tools/.coursier-cache"
 # Install tools provided by Ubuntu.
 RUN DEBIAN_FRONTEND="noninteractive" apt-get install -y --no-install-recommends \
     build-essential \
-    openjdk-8-jdk-headless \
+    openjdk-11-jdk-headless \
     maven \
     python3-dev \
     subversion \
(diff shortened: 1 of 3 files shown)


cmungall (approved):
Seems like a good idea - I don't know if there are any additional tests we should run but I am happy to proceed with merging

gouttegd:
I’ll build and run the test suite on my M1 (arm64) machine tomorrow. I don’t expect any issue but just in case…

gouttegd:
All tests passed successfully on arm64. :)

matentzn:
I will run some additional tests over the weekend, but an important test case is the fbbt prepare release process due to the custom Java jars in there. @gouttegd can you run a full release with the Java 11 images of fbbt, just to be sure all components still work?

gouttegd:
@matentzn Done, no issues. :)

balhoff:
Today I will add a blazegraph-runner update to this PR. I made a newer release which addresses usage of some old removed Java API. Let's not merge until then.

PR #493, opened 2021-11-15
Update Makefile.jinja2

shawntanzk:
Added explain_unsat command

### DIFF
--- a/template/src/ontology/Makefile.jinja2
+++ b/template/src/ontology/Makefile.jinja2
@@ -785,7 +785,7 @@ $(ONT)-basic.owl: $(SRC) $(OTHER_SRC) $(SIMPLESEED) $(KEEPRELATIONS)
 # ----------------------------------------
 
 explain_unsat: $(SRC) 
-	$(ROBOT) explain -i $(SRC) -M unsatisfiability --unsatisfiable random:10 --explanation $(TMP)
+	$(ROBOT) explain -i $(SRC) -M unsatisfiability --unsatisfiable random:10 --explanation $(TMP)/$@.md
 
 {%- if project.public_release != 'none'  %}
 # ----------------------------------------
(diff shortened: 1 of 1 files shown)


matentzn (approved):

matentzn (changes requested):
Sorry I meant to "request changes" before/

template/src/ontology/Makefile.jinja2
```diff
@@ -780,6 +780,13 @@ $(ONT)-basic.owl: $(SRC) $(OTHER_SRC) $(SIMPLESEED) $(KEEPRELATIONS)
 {% endif -%}
 {% endfor -%}
 
+# ----------------------------------------
+# Debugging Tools
+# ----------------------------------------
+
+explain_unsat: $(SRC) 
+	$(ROBOT) explain -i $(SRC) -M unsatisfiability --unsatisfiable random:10 --explanation tmp/$@.md
```
matentzn, line 788: Sorry one last, use $(TMP) (make sure this is the one) instead of `tmp`

matentzn (changes requested):

template/src/ontology/Makefile.jinja2
shawntanzk: Done - I assume it will automatically make it a markdown file?

matentzn: Ah sorry, this would try to save the explanation as a  file named `$(TMP)` - that would not work. The explanation needs to be called `--explanation $(TMP)/$@.md`. If ROBOT sees the md, it understands that it should be markdown. Sorry about the confusion.

shawntanzk: yeah wasn't sure what $(TMP) pointed to, but got it now :) have made the changes. Thanks

matentzn (approved):
Thank you @shawntanzk

template/src/ontology/Makefile.jinja2
matentzn: Thank you!

PR #495, opened 2021-11-17
Update the run.sh wrapper to display runtime stats.

gouttegd:
Add the `time` package to odklite. Use the provided `time` command in the run.sh wrapper to print elapsed time and peak memory consumption of the invoked command in the container.

The feature is conditional on the variable `ODK_DEBUG` being set to `yes` when invoking the wrapper. Either set at the time of invocation:

```sh
$ ODK_DEBUG=yes ./run.sh make prepare_release
[... normal output of "make prepare_release" ...]
### DEBUG STATS ###
Elapsed time: 0:55.75
Peak memory: 4407880 kb
```

or to always get the stats, export `ODK_DEBUG` in the shell environment:

```sh
$ export ODK_DEBUG=yes
$ ./run.sh make prepare_release
[... normal output of "make prepare_release" ...]
### DEBUG STATS ###
Elapsed time: 0:55.75
Peak memory: 4407880 kb
```

### DIFF
--- a/template/src/ontology/run.sh.jinja2
+++ b/template/src/ontology/run.sh.jinja2
@@ -12,4 +12,14 @@
 
 IMAGE=${IMAGE:-odkfull}
 
-docker run -v $PWD/../../:/work -w /work/src/ontology {% if project.robot_java_args is defined %}-e ROBOT_JAVA_ARGS='{{ project.robot_java_args }}' -e JAVA_OPTS='{{ project.robot_java_args }}'{% endif %} {% if project.use_external_date is sameas True %}-e TODAY=`date +%Y-%m-%d` {% endif %}--rm -ti obolibrary/$IMAGE "$@"
+ODK_DEBUG=${ODK_DEBUG:-no}
+
+TIMECMD=
+if [ x$ODK_DEBUG = xyes ]; then
+    # If you wish to change the format string, take care of using
+    # non-breaking spaces (U+00A0) instead of normal spaces, to
+    # prevent the shell from tokenizing the format string.
+    TIMECMD="/usr/bin/time -f ### DEBUG STATS ###\nElapsed time: %E\nPeak memory: %M kb"
+fi
+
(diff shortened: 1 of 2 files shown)


template/src/ontology/run.sh.jinja2
```diff
@@ -12,4 +12,14 @@
 
 IMAGE=${IMAGE:-odkfull}
 
-docker run -v $PWD/../../:/work -w /work/src/ontology {% if project.robot_java_args is defined %}-e ROBOT_JAVA_ARGS='{{ project.robot_java_args }}' -e JAVA_OPTS='{{ project.robot_java_args }}'{% endif %} {% if project.use_external_date is sameas True %}-e TODAY=`date +%Y-%m-%d` {% endif %}--rm -ti obolibrary/$IMAGE "$@"
+ODK_DEBUG=${ODK_DEBUG:-no}
+
+TIMECMD=
+if [ x$ODK_DEBUG = xyes ]; then
```
matentzn, line 18: I have never seen that pattern with the `x` - I can see its not wrong, but just out of curiosity - whats the reason?

matentzn:
This is pretty amazing man, thank you so much... Very cool!

template/src/ontology/run.sh.jinja2
gouttegd: Suppose you have the following instead: `if [ $ODK_DEBUG = yes ]`. When `ODK_DEBUG` is not set (or is set to an empty string), that line will become `if [ = yes ]`, which is a syntax error (the shell would complain about a missing operand for the `=` operator or something like that). The `x` is here to avoid that.

An alternative is to quote the variable: `if [ "$ODK_DEBUG" = yes ]`, which would then result, if the variable is not defined, in `if [ "" = yes ]`. But some (mostly old) shells won’t accept that.

matentzn: Cheers, thanks for the info!

PR #496, opened 2021-11-18
BASE module extraction pipeline (and mo)

matentzn:
We are running tests with PATO here:

https://github.com/pato-ontology/pato/pull/453

This pull request changes the module extraction pipeline in such a way that we can use bases instead of merged files. This will get rid of all our current interoperability nightmares.

We still need to decide though whether the additional complexity we have no is worth it. There are two approaches to base module extraction:

1. Extract from merged. This means we simply use the normal seed signature to extract a module from the merged mirrors. Very easy to implement. Downside: there is only one module (external.owl).
2. Extract signature from merged, than extract module from individual bases. This is a bit more complicated as we use the normal seed to extract a module from merged (same as above), then extract the signature S from that module, and use S to extract a module from the base mirror. The upside is we keep individual import files for the edit file, which makes looking for "where does this axiom come from" easier during debugging. The downside is computational overhead (needless IMO, given that imports are merged at release time).


template/src/ontology/Makefile.jinja2
```diff
@@ -841,6 +869,17 @@ public_release:
 	$(GITHUB_RELEASE_PYTHON) --release $(TAGNAME) $(RELEASEFILES)
 {%- endif %}
 
+{% if 'obo' in project.edit_format %}
+.PHONY: normalize_obo_src
+normalize_obo_src: $(SRC)
+	$(OWLTOOLS) $< --merge-axiom-annotations -o -f obo $(TMPDIR)/NORM.obo && robot convert -i $(TMPDIR)/NORM.obo -o $(TMPDIR)/NORM.tmp.obo && mv $(TMPDIR)/NORM.tmp.obo $(SRC)
```
gouttegd, line 875: We should probably use `$(ROBOT)` here instead of `robot`. This will allow the use of the XML catalog, and should automatically fix obophenotype/uberon#2169.

gouttegd (changes requested):
Looks good to me. My only concern is the lack of a test fixture for the new mode.

template/src/ontology/Makefile.jinja2
matentzn: Ooops. THANKS!

matentzn:
Ok, I mixed some other things into this PR, but its the last thing for the ODK release.

gouttegd (approved):

PR #497, opened 2021-11-18
Make 'allow_equivalents' default to 'asserted-only'.

gouttegd:
We want to promote the avoidance of inferred equivalencies, so we switch the `allow_equivalents` setting from its default value of `all` (default value for the `--equivalent-classes-allowed` of ROBOT’s `explain` command) to `asserted-only`.

See Chris Mungall’s [blog post](https://douroucouli.wordpress.com/2018/09/04/debugging-ontologies-using-owl-reasoning-part-2-unintentional-entailed-equivalence/) for more information on the rationale for this switch.

This PR only changes the _default_ value and therefore won’t affect the ontologies that have set an explicit value for the `allow_equivalents` setting in their `src/ontology/$OID-odk.yaml` file. Ontologies that were relying on the previous default value of `all` may need to update their configuration to explicitly set `allow_equivalents` to `all`, if they happen to contain non-asserted inferred equivalencies that they need to keep (or cannot remove).

### DIFF
--- a/odk/odk.py
+++ b/odk/odk.py
@@ -491,8 +491,8 @@ class OntologyProject(JsonSchemaMixin):
     release_date : bool = False
     """if true, releases will be tagged with a release date (oboInOwl:date)"""
     
-    allow_equivalents : str = "assert-only"
-    """can be all, none or assert-only (see ROBOT documentation: http://robot.obolibrary.org/reason)"""
+    allow_equivalents : str = "asserted-only"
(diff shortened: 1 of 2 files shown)


cmungall (approved):

gouttegd:
@matentzn Do you want to wait for more people to weigh in on this?

matentzn:
Its fine, can be merged.

PR #498, opened 2021-11-22
Make $(SRCMERGED) depend on $(OTHER_SRC).

gouttegd:
The entire point of the rule generating the `$(SRCMERGED)` file is to merge any component listed in `$(OTHER_SRC)`, so that rule should depend on `$(OTHER_SRC)` in addition to `$(SRC)`, so that any changes in a component can trigger a rebuild of `$(SRCMERGED)`.

### DIFF
--- a/template/src/ontology/Makefile.jinja2
+++ b/template/src/ontology/Makefile.jinja2
@@ -227,7 +227,7 @@ IMPORTSEED=$(TMPDIR)/seed.txt
 SRCMERGED=$(TMPDIR)/merged-$(SRC)
 PRESEED=$(TMPDIR)/pre_seed.txt
 
-$(SRCMERGED): $(SRC)
+$(SRCMERGED): $(SRC) $(OTHER_SRC)
 	$(ROBOT) remove --input $< --select imports --trim false \
 		merge  $(patsubst %, -i %, $(OTHER_SRC)) -o $@
 


matentzn (approved):

matentzn:
This is such an odd omission.. Thank you!

matentzn:
Thank you QC.. Error suggests there is now a cycle in the Makefile..

gouttegd:
So… `$(OTHER_SRC)` contains both the _components_, plus some files derived from the use of DOSDP, notably `patterns/definitions.owl`.

`patterns/definitions.owl` depends, among other things, on `dosdp_patterns_default`, which depends on `all_imports`.

Each `import/XXX_import.owl` (part of `all_imports`) depends on `$(IMPORTSEED)`, which depends on `$(ALLSEED)`, which depends on `$(PRESEED)`, which depends on `$(SRCMERGED)`.

And with this PR, `$(SRCMERGED)` depends on `$(OTHER_SRC)`, so it depends on `patterns/definitions.owl`.

Argh.

I’ll think about how to fix that later. For now, I have a sudden urge to become a dairy farmer in the English countryside.

matentzn:
Yes, this is a hugely complicated problem actually, and right now the ODKs seeding strategy is far less then optimal.. Very unfortunate! No worries, it just means we need to think about fixing the dependencies more carefully during a hack session at some point.

matentzn:
I will merge this now into my dev branch. The goal is to get rid of all dependencies between components and imports. components, such as definitions.owl should _never_ depend on imports, only the other way around.

PR #513, opened 2022-01-13
Update TaxonRestriction.md

dosumis:
Updating + adding more practical suggestion for AP annotation.

### DIFF
--- a/docs/TaxonRestriction.md
+++ b/docs/TaxonRestriction.md
@@ -4,19 +4,21 @@
 
 See [Daily Workflow](http://ontology-development-kit.readthedocs.io/en/latest/index.html#daily-workflow) for creating branches and basic Protégé instructions. 
 
-1. ```Only in taxon``` relations are added as ```Subclasses```. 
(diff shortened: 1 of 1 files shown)


dosumis:
@shawntanzk - think I'm editing general docs in correct place.  Please let me know if not.  Also I seem to remember we had some general/theoretical doc on taxon restrictions.  Do you know where that might live?

shawntanzk:
@dosumis literally facing the same problem atm - I can't seem to remember where the central location for documents that get distributed to all ODK ontologies reside. I don't think it's the docs here as they are different from the ones in ontology docs.

matentzn:
Also note

https://github.com/OBOFoundry/obo-patterns.github.io

and in particular

http://obofoundry.org/obo-patterns.github.io/key-topics/taxon_constraints/

shawntanzk:
I've moved this to https://github.com/OBOAcademy/obook/pull/96/files as part of ODK docs migration

Issue #514, opened 2022-01-13
Documentation for managing ontologies with merged/base file imports 


shawntanzk:
We need to write up documentation on how to manage ontologies that have switched to base file imports such as CL and PATO. 
Suggest that we edit UpdateImports section in https://github.com/INCATools/ontology-development-kit/edit/master/template/_dynamic_files.jinja2 to include a subsection on base file imports. 

We further need documentation on how to switch ontologies to base file imports but that can come later

shawntanzk:
Related to https://github.com/obophenotype/cell-ontology/issues/1404

matentzn:
I would not recommend this strategy anymore, especially for people that are able to run odk: the main strategy should be to add the id, as we previously did to the `imports/cl_terms.txt` file, then refresh the import. However, documenting as you have done is good, as editors with no ability to run odk may need to do it using the Protege-based strategy.

shawntanzk:
oh I thought that the individual import files are not used anymore - so basically if we add it to terms then refresh merged it will work?

matentzn:
Yes, if everything goes as planned, this here:
https://github.com/INCATools/ontology-development-kit/blob/master/template/src/ontology/Makefile.jinja2#L273

Should merge all the term files into one big heap and then use that to extract the module.

shawntanzk:
Thanks @matentzn - I'll update the docs accordingly

PR #519, opened 2022-01-17
Updated import docs

shawntanzk:
See https://github.com/INCATools/ontology-development-kit/issues/514#issuecomment-1014272986

Sorry about that @bvarner-ebi & @dosumis - my bad for misunderstanding the import system. Fixed the docs accordingly here.

### DIFF
--- a/template/_dynamic_files.jinja2
+++ b/template/_dynamic_files.jinja2
@@ -948,7 +948,8 @@ There are three ways to declare terms that are to be imported from an external o
 
 #### Protege-based declaration
 
-This workflow is to be avoided, but may be appropriate if the editor _does not have access to the ODK docker container_.
+This workflow is to be avoided, but may be appropriate if the editor _does not have access to the ODK docker container_. 
+This approach also applies to ontologies that use base module import approach.
 
 1. Open your ontology (edit file) in Protege (5.5+).
 1. Select 'owl:Thing'
--- a/template/_dynamic_files.jinja2
+++ b/template/_dynamic_files.jinja2
@@ -934,7 +934,7 @@ This page discusses how to update the contents of your imports, like adding or r
 
 Note: some ontologies now use a merged-import system to manage dynamic imports, for these please follow instructions in the section title "Using the Base Module approach".
 
-Importing a new term is split into to sub-phases:
+Importing a new term is split into two sub-phases:
 
 1. Declaring the terms to be imported
 2. Refreshing imports dynamically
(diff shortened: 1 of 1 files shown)


template/_dynamic_files.jinja2
```diff
@@ -1083,15 +1083,11 @@ To check if your ontology uses this method, check src/ontology/{{ project.id }}-
...
 
-Start by declaring the term to be imported in Protege:
+To check if your ontology uses this method, check src/ontology/{{ project.id }}-odk.yaml to see if `use_base_merging: TRUE` is declared under `import_group`
 
-1. Open your ontology (edit file) in Protege (5.5+).
-1. Select 'owl:Thing'
-1. Add a new class as usual.
-1. Paste the _full iri_ in the 'Name:' field, for example, http://purl.obolibrary.org/obo/CHEBI_50906.
-1. Click 'OK'
+If your ontology uses Base Module approach, please use the following steps:
```
gouttegd, line 1088: Lines 1086–1088 are duplicate of lines 1082–1084.

gouttegd (changes requested):
Looks good to me except for the duplicated lines around 1082–1084.

template/_dynamic_files.jinja2
shawntanzk: whoops, my bad, i accidentally commited on a dirty branch, and was moving things around and missed that. Thanks for spotting :)

matentzn:
I am not sure why the instruction for the Protege based approach are removed entirely.. How should people work without access to docker/odk? Are these instructions going somewhere else?

shawntanzk:
> I am not sure why the instruction for the Protege based approach are removed entirely.. How should people work without access to docker/odk? Are these instructions going somewhere else?

ok looking at the diff for this, something looks odd -let me go relook at the whole thing again and redo this properly. (there was some removal cause there was duplicated things, but it seems that it ended up fully removing it. Sorry about that will fix

shawntanzk:
@matentzn ok looked at the whole docs, the protege-based declaration is still up top, however I removed it from the base file approach after recommendation that that is not great a way. With base file approach, ODK is kinda needed anyway no? (eg if someone is adding using the protege-based declaration, they have the instructions up top, but if someone is working with base-file, they would not need to do with protege-based declaration.) 
Happy to add back the protege-based declaration to the base-file section if you think its useful there too

matentzn:
Regardless of base or anything else, adding a term can be done using Protege if necessary, even if that is discouraged. In order to actual _run_ the import (import the term, not just declare it (step1)), you need ODK in both cases!

shawntanzk:
> Regardless of base or anything else, adding a term can be done using Protege if necessary

This step is already present in ln949 - should I copy and paste it somewhere around 1084?

matentzn:
No, if its clear that it applies to all approaches, then no need for duplication

shawntanzk:
I'll make sure it is :)

shawntanzk:
ready to go - will merge after a thumbs up

gouttegd (approved):

PR #527, opened 2022-01-31
Add sparql check for taxon and subset range

### LINKED ISSUE #520: Add SPARQL test for taxon annotation property range
never_in_taxon (RO_0002161) and present_in_taxon (RO_0002175) are annotation properties that require a taxon (NCBI taxon class) as a value.  Ideally this range would be specified in RO, but limitations on specifying range for annotationProperties means there is no obvious way to do this.  To compensate, we need a SPARQL check.  

I think this should have two components:
1. test that the value is an IRI
2. cast to string and test that it begins with 'http://purl.obolibrary.org/obo/NCBITaxon_'

anitacaron:
Fixes #520

### DIFF
--- a/odk/odk.py
+++ b/odk/odk.py
@@ -317,7 +317,7 @@ class ReportConfig(JsonSchemaMixin):
     release_reports : bool = False
     """ If true, release reports are added as assets to the release (top level directory, reports directory)"""
     
-    custom_sparql_checks : Optional[List[str]] = field(default_factory=lambda: ['equivalent-classes', 'owldef-self-reference'])
+    custom_sparql_checks : Optional[List[str]] = field(default_factory=lambda: ['equivalent-classes', 'owldef-self-reference', 'taxon-range', 'iri-range'])
     """Chose which additional sparql checks yoy want to run. The related sparql query must be named CHECKNAME-violation.sparql, and be placed in the src/sparql directory"""
 
     custom_sparql_exports : Optional[List[str]] = field(default_factory=lambda: ['basic-report', 'class-count-by-prefix', 'edges', 'xrefs', 'obsoletes', 'synonyms'])
--- a/odk/odk.py
+++ b/odk/odk.py
@@ -317,7 +317,7 @@ class ReportConfig(JsonSchemaMixin):
     release_reports : bool = False
     """ If true, release reports are added as assets to the release (top level directory, reports directory)"""
     
-    custom_sparql_checks : Optional[List[str]] = field(default_factory=lambda: ['equivalent-classes', 'owldef-self-reference', 'taxon-range', 'iri-range'])
+    custom_sparql_checks : Optional[List[str]] = field(default_factory=lambda: ['equivalent-classes', 'owldef-self-reference', 'iri-range'])
     """Chose which additional sparql checks yoy want to run. The related sparql query must be named CHECKNAME-violation.sparql, and be placed in the src/sparql directory"""
 
     custom_sparql_exports : Optional[List[str]] = field(default_factory=lambda: ['basic-report', 'class-count-by-prefix', 'edges', 'xrefs', 'obsoletes', 'synonyms'])
--- a/template/_dynamic_files.jinja2
+++ b/template/_dynamic_files.jinja2
@@ -486,6 +486,30 @@ SELECT ?x ?xl ?y ?yl ?z ?zl WHERE {
 }
 {% endif %}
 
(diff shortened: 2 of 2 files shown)


odk/odk.py
```diff
@@ -317,7 +317,7 @@ class ReportConfig(JsonSchemaMixin):
     release_reports : bool = False
     """ If true, release reports are added as assets to the release (top level directory, reports directory)"""
     
-    custom_sparql_checks : Optional[List[str]] = field(default_factory=lambda: ['equivalent-classes', 'owldef-self-reference'])
+    custom_sparql_checks : Optional[List[str]] = field(default_factory=lambda: ['equivalent-classes', 'owldef-self-reference', 'taxon-range', 'iri-range'])
```
matentzn, line 320: dont add 'taxon-range' to default better

matentzn (changes requested):

odk/odk.py
anitacaron: Ok. Do I include both `present_in_taxon` and `never_in_taxon` in the iri-range?

matentzn: Yeah. And rdfs:seeAlso, dcterms:contributor and whatever else you can think of

anitacaron: It doesn't work `isIRI` check with `anyURI` type that are the cases you included. I'll only include the three properties `in_subset`, `present_in_taxon` and `never_in_taxon`.

matentzn: Yeah I forgot. you can check "isIRI(?x)" or (DATATYPE(?x)=xsd:anyUri). What do you think?

anitacaron: Sure, I was also thinking about it. Found many violations using `xsd:string`.

matentzn:
@anitacaron I started a discussion in the `#sparql` channel on obo-slack, lets get some feedback first on this anyUri vs RDF resource issue.

template/_dynamic_files.jinja2
```diff
@@ -486,6 +486,43 @@ SELECT ?x ?xl ?y ?yl ?z ?zl WHERE {
...
+WHERE {
+  VALUES ?property { 
+    never_in_taxon: 
+    present_in_taxon: 
+    oboInOwl:inSubset
+    rdfs:seeAlso
+    dct:contributor
+    foaf:depicted_by  }
+  ?term ?property ?value .
+  FILTER (!isIRI(?value) && (datatype(?value) != xsd:anyURI))
```
matentzn, line 522: Lets have one qc query for isIRI (never_in_taxon: 
    present_in_taxon: 
    oboInOwl:inSubset
    dct:contributor)

and a separate one for xsd:anyUri

foaf:depicted_by
rdfs:seeAlso

template/_dynamic_files.jinja2
```diff
@@ -486,6 +486,55 @@ SELECT ?x ?xl ?y ?yl ?z ?zl WHERE {
...
+PREFIX foaf: <http://xmlns.com/foaf/0.1/>
+PREFIX xsd: <http://www.w3.org/2001/XMLSchema#>
+
+SELECT ?term ?property ?value
+WHERE {
+  VALUES ?property { 
+    rdfs:seeAlso
+    foaf:depicted_by  }
+  ?term ?property ?value .
+  FILTER (datatype(?value) != xsd:anyURI)
```
matentzn, line 534: @balhoff would like your opinion on this QC check. (and the two above)

Thank you :)

balhoff: Look good to me!

matentzn: grammatically or do you also agree with this check (i.e. restricting the datatype of the range of rdfs:seeAlso and foaf:depicted_by to xsd:anyUri?

balhoff: Mostly grammatically, I think foaf:depicted_by sounds good. Not sure about rdfs:seeAlso; I guess it makes sense but I see some direct IRI usage (e.g., in RO).

matentzn: Yeah I am wondering the same. I am still not exactly certain why not everything is an IRI.. What is the reason for xsd:anyUri to exist exactly?

balhoff: I don't really know. I would be in favor of using IRIs. I think originally @cmungall suggested using anyURI for our GitHub links. @cmungall do you have criteria for a choice?

matentzn:
I will merge it but we will check back based on feedback later.

PR #530, opened 2022-02-05
Last push 1.2.33

matentzn:
- Move Mirror pipeline to the .PHONY / cmp model, i.e. attempt the download, and only overwrite the goal if the file has changed.
- Redesign the entire DOSDP dependency pipeline to get rid of phony goals
- Add a new `make help` goal that prints usage information to the command line

### DIFF
--- a/template/src/ontology/Makefile.jinja2
+++ b/template/src/ontology/Makefile.jinja2
@@ -58,7 +58,6 @@ ONTOLOGYTERMS =             $(TMPDIR)/ontologyterms.txt
 
 {%- if project.use_dosdps %}
 PATTERNDIR=                 ../patterns
-DOSDP_SCHEMA=                http:// # change to PURL when ready.
 PATTERN_TESTER=              dosdp validate -i
 DOSDPT=                      dosdp-tools
 PATTERN_RELEASE_FILES=       $(PATTERNDIR)/definitions.owl $(PATTERNDIR)/pattern.owl
@@ -250,10 +249,10 @@ $(IMPORT_MODULE_SIGNATURE): $(IMPORT_MODULE)
 	cat $@.tmp | sort | uniq >  $@
 {% endif %}
 
-ALLSEED = $(PRESEED) {% if project.use_dosdps %} $(PATTERNDIR)/all_pattern_terms.txt {% endif %}\
+ALLSEED = $(PRESEED) {% if project.use_dosdps %} $(TMPDIR)/all_pattern_terms.txt {% endif %}\
 {% if project.use_custom_import_module %} $(IMPORT_MODULE_SIGNATURE) {% endif %}
 
-$(IMPORTSEED): {% if project.use_dosdps %} prepare_patterns {% endif %} $(ALLSEED) 
+$(IMPORTSEED): $(ALLSEED)
 	if [ $(IMP) = true ]; then cat $(ALLSEED) | sort | uniq > $@; fi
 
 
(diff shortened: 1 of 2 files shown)


gouttegd:
@matentzn I see nothing obviously wrong and the FBbt release pipeline seems to work just fine. Still, if you’re not in a hurry to merge and release I’d like to have yet another look by next week.

matentzn:
I am not in a rush - I tried a few different repos so far, but I want to get this right :P Thanks!

gouttegd:
Correction: there’s one bit that the FBbt release pipeline doesn’t like: the fact that `$(PATTERN_RELEASE_FILES)` is now included in `$(ASSETS)`.

This breaks our custom `prepare_release` target, apparently because `rsync` doesn’t like the fact the `$(PATTERN_RELEASE_FILES)` refers to a parent directory (judging by `rsync` error message: `found ".." dir in relative path: ../patterns/definitions.owl`).

But it’s our (FBbt) fault for messing with the default `prepare_release` target, I guess. We can fix that on our side after ODK 1.3.0 is released.

gouttegd:
Actually, given that the standard Makefile has been overhauled, it would probably a good idea for all the ontologies that are using a customised Makefile to review it and check whether the overridden rules are still relevant…

template/src/ontology/Makefile.jinja2
```diff
@@ -393,233 +425,231 @@ no-mirror-refresh-%:
...
 {% if component.source is not none %}
-$(COMPONENTSDIR)/{{ component.filename }}: .FORCE
+.PHONY: component-download-{{ component.filename }}
+component-download-{{ component.filename }}: | $(TMPDIR)
 	if [ $(IMP) = true ]; then $(ROBOT) merge -I {{ component.source }} \
-	annotate --ontology-iri $(ONTBASE)/$@ $(ANNOTATE_ONTOLOGY_VERSION) -o $@; fi
+	annotate --ontology-iri $(ONTBASE)/$@ $(ANNOTATE_ONTOLOGY_VERSION) -o $(TMPDIR)/$@.owl; fi
+
+$(COMPONENTSDIR)/{{ component.filename }}:  component-download-{{ component.filename }}
+	cmp -s $(TMPDIR)/component-download-{{ component.filename }}.owl $@; RETVAL=$$?; if [ $$RETVAL -eq 0 ]; then echo "Component identical, ignoring."; else echo "Component is different, updating." && cp $(TMPDIR)/component-download-{{ component.filename }}.owl $@; fi
```
matentzn, line 440: From @Clare72 (thanks!)

Missing conditional, see https://github.com/FlyBase/drosophila-anatomy-developmental-ontology/pull/1388#discussion_r806829649

gouttegd: Beyond the missing test on the existence of the file, I am wondering about the use of that intermediate variable `RETVAL`. What does it bring compared to a simpler form:

```sh
if cmp -s $(TMPDIR)/component-download-{{ component.filename }}.owl $@ ; then
    echo "Component identical, ignoring."
else
    echo "Component different, updating..."
fi
```

?

template/src/ontology/Makefile.jinja2
```diff
@@ -393,233 +425,231 @@ no-mirror-refresh-%:
...
+.PHONY: download_patterns
+download_patterns:
+	if [ $(PAT) = true ]; then rm -f $(TMPDIR)/dosdp/*.yaml.1 || true; fi
+	if [ $(PAT) = true ] && [ -s $(PATTERNDIR)/dosdp-patterns/external.txt ]; then wget -i $(PATTERNDIR)/dosdp-patterns/external.txt --backups=1 -P $(TMPDIR)/dosdp; fi
+	if [ $(PAT) = true ]; then rm -f $(TMPDIR)/dosdp/*.yaml.1 || true; fi
 
-$(PATTERNDIR)/pattern.owl: pattern_schema_checks update_patterns
-	if [ $(PAT) = true ]; then $(DOSDPT) prototype --obo-prefixes true --template=$(PATTERNDIR)/dosdp-patterns --outfile=$@; fi
+$(PATTERNDIR)/dospd-patterns/%.yml: download_patterns
+	cmp -s $(TMPDIR)/dosdp-$*.yml $@; RETVAL=$$?; if [ $$RETVAL -eq 0 ]; then echo "Mirror identical, ignoring."; else echo "Mirrors different, updating." && mv $(TMPDIR)/dosdp/$*.yml $@; fi
```
gouttegd, line 589: Same as above: missing test on the existence of the file before attempting the comparison, and unnecessary use of `RETVAL`.

gouttegd (changes requested):
I don’t see other problems than those already discussed, and with the missing tests added FBbt builds and checks fine (if we update our custom `prepare_release` rule).

Two general comments (which we may or may not want to address for this release):

* Every time `make` is invoked in a rule, it’s a good practice to invoke it as `$(MAKE)` – this ensure we always call the _same_ `make` as the one initially invoked by the user. In the context of the ODK container this does not really matter (there will always be only one `make` in the ODK container), but I sometimes run some rules “directly” (without going through Docker), and I do have at least two `make`s on my machine.
* Cosmetic only: I’d like to generalise the use of line breaks (with `\`) to avoid overly long lines. We already do that for _some_ rules, but not for all.

gouttegd (approved):

matentzn:
THANK YOU so much @gouttegd 

I am running some last tests, then will merge it today.

PR #538, opened 2022-02-08
Preserve original permissions when copying template files.

### LINKED ISSUE #196: ./run.sh make syntax missing privileges
Since the move to the new Makefile, the `./run.sh make` command syntax does not work anymore, I think because run.sh does not have +x executable privileges.

gouttegd:
When instanciating an ODK repository, copy files from the template using `shutil.copy` instead of `shutil.copyfile`. The former has the benefit of preserving the permissions of the source file.

For files that are generated from a Jinja2 template, explicitly transfer the permission bits from the source template to the generated file.

close #196

### DIFF
--- a/odk/odk.py
+++ b/odk/odk.py
@@ -20,7 +20,7 @@ import yaml
 import os
 import subprocess
 import shutil
-from shutil import copyfile
+from shutil import copy, copymode
 import logging
 
 logging.basicConfig(level=logging.INFO)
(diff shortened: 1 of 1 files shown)


matentzn (approved):

matentzn:
This is awesome! Will try it now!

matentzn:
Seems to work!

PR #542, opened 2022-02-10
Add check to look for IRIs in label

### LINKED ISSUE #536: Adding check to look for IRIs in textual annotations
When running DOSDP or similar pipeline, it can happen that DOSDP tools does not have access to a term to autogenerate its metadata, resulting in something like this:

```
AnnotationAssertion(<http://purl.obolibrary.org/obo/IAO_0000115> <http://purl.obolibrary.org/obo/CATO_0000001> "A coat of hair with a http://purl.obolibrary.org/obo/PATO_0000333 colour pattern."^^xsd:string)
```

We need a check that:

- [x] Is added to custom_sparql_queries
- [x] Looks for `http[s]?[:]` in rdfs:label and IAO:0000115
- [x] Is restricted to the looking at entities from the namespace of the ontology (see `^^^ src/sparql/{{ project.id }}_terms.sparql` in dynamic jinja2

anitacaron:
Fixes #536

### DIFF
--- a/odk/odk.py
+++ b/odk/odk.py
@@ -317,7 +317,7 @@ class ReportConfig(JsonSchemaMixin):
     release_reports : bool = False
     """ If true, release reports are added as assets to the release (top level directory, reports directory)"""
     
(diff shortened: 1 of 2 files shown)


matentzn:
Some problem with the query syntax, see QC.

matentzn (approved):

PR #543, opened 2022-02-10
Update relation-graph

balhoff:
New version has quieter error reporting and version output.


matentzn:
Thank you for keeping it updated!

PR #550, opened 2022-02-16
Make docs deployment daily

shawntanzk:
As requested by @dosumis to make docs deployment daily


matentzn:
Not sure I understand - this action is run every time something is merged into master, why would you want it to run once more every day?

shawntanzk:
oh whoops, didn't check here properly, saw in uberon that it was manual and assumed that it was that way here >.< I'll copy over the on: push to uberon instead.

matentzn:
You may have to copy the whole action, if the uberon one is old!

PR #557, opened 2022-02-20
Fixing a critical bug in `make clean` (unpublished, phew) (and some other last second changes).

matentzn:
A Bug in the Makefile almost cost me dearly - if it was not for its safeguards (and docker).. please make sure that I am getting it right. This command:

```
rm -rf $(MIRROR)/
```

translated to

```
rm -rf /
```

because `$(MIRROR)` does not exist.

I tried to make rm -f safer by adding `./` everywhere (maybe not necessary?), and removed all -r except for one, which I now call with hard coded paths rather than variable names, to avoid some crazy environment variable madness. If you have a better idea let me know

EDIT Other changes I wrapped into this pull:

- if basefile approach is active, module refresh didn't refresh merged (fixed)
- if basefile approach is active, `all_imports` should only refresh `merged`
- Using http://purl.org/dc/terms (dcterms) across the board instead of elements
- Speeding up some of the QC queries by adding the correct new filter pattern
- Adding a test for the matches pipeline


template/src/ontology/Makefile.jinja2
```diff
@@ -232,14 +232,14 @@ CLEANFILES=$(MAIN_FILES) $(SRCMERGED)
 prepare_release: all
 	rsync -R $(RELEASE_ASSETS) $(RELEASEDIR) &&\{% if project.use_dosdps %}
 	mkdir -p $(RELEASEDIR)/patterns && cp -rf $(PATTERN_RELEASE_FILES) $(RELEASEDIR)/patterns &&\{% endif %}
-	rm -f $(CLEANFILES) &&\
+	rm -f $(patsubst %, ./%, $(CLEANFILES)) &&\
```
matentzn, line 235: cc @gouttegd

template/src/ontology/Makefile.jinja2
```diff
@@ -552,14 +552,16 @@ PAT=true # Global parameter to bypass pattern generation
...
+# the environment does not malicously set `PATTERN_CLEAN_FILES` to `\`.
 .PHONY: pattern_clean
 pattern_clean:
-	rm -f $(DOSDP_OWL_FILES_DEFAULT)
-	rm -f $(DOSDP_TERM_FILES_DEFAULT){% if project.pattern_pipelines_group is defined %}
-	{%- for pipeline in project.pattern_pipelines_group.products %}
-	rm -f $(DOSDP_OWL_FILES_{{ pipeline.id.upper() }})
-	rm -f $(DOSDP_TERM_FILES_{{ pipeline.id.upper() }})
-	{%- endfor %}{% endif %}
+	rm -f $(patsubst %, ./%, $(PATTERN_CLEAN_FILES))
```
matentzn, line 564: cc @gouttegd  I only need you to review these lines (555-564), the rest is fine.

template/src/ontology/Makefile.jinja2
```diff
@@ -946,12 +948,13 @@ update_docs:
...
+# Note to future generations: prepending ./ is a safety measure to ensure that 
+# the environment does not malicously set `CLEANFILES` to `\`.
 .PHONY: clean
 clean:{% if project.use_dosdps %}
 	make pattern_clean{%- endif %}
-	rm -rf $(TMPDIR)/
-	rm -rf $(MIRROR)/
-	rm -f $(CLEANFILES)
+	rm -rf ./tmp/* ./mirror/*
+	rm -f $(patsubst %, ./%, $(CLEANFILES))
```
matentzn, line 957: cc @gouttegd

gouttegd: OK, but this may break the user’s expectation that they are allowed to redefine `MIRRORDIR` or `TMPDIR` in the custom Makefile, and have the entire Makefile take this redefinition into account.

I’d prefer to keep referring to the mirror and temporary directories using `MIRRORDIR` and `TMPDIR`, with possibly some checks before `rm`-ing them if you are concerned they could have been re-defined in some harmful way.

E.g., check that the variable is not empty and does point to a directory existing in the current folder:

```
[ -n "$(MIRRORDIR)" ] && [ -d ./$(MIRRORDIR) ] && rm -rf ./$(MIRRORDIR)/*
```

Depending on your paranoia level, you may also want to protect against the case where `MIRRORDIR` is maliciously set to `.` or `..`:

```
[ -n "$(MIRRORDIR)" ] && [ $(MIRRORDIR) != "." ] && [ $(MIRRORDIR) != ".." ] && [ -d ./$(MIRRORDIR) ] && rm -rf ./$(MIRRORDIR)/*
```

Note that this does not protect against the case where `MIRRORDIR` is even more malicious, e.g. `MIRRORDIR=normal/path/../../../../oh/noes`). But I am not convinced  we should go into a mode where we systematically assume someone _will_ try to play tricks with us (we could go into a rabbit hole if we start thinking like that).

gouttegd (approved):

template/src/ontology/Makefile.jinja2
matentzn: No, lets do it right, changed it back to `rm -f $(CLEANFILES)`.

template/src/ontology/Makefile.jinja2
matentzn: Thanks a ton man, I did what you suggested. I will ask you during the call later to review one last time.

gouttegd (approved):

matentzn:
Thank you @gouttegd, again!

PR #577, opened 2022-03-16
added components doc

shawntanzk:


template/_dynamic_files.jinja2
```diff
@@ -1356,6 +1356,55 @@ This requires quite a bit of memory on your local machine, so if you encounter a
...
+```
+Import(<http://purl.obolibrary.org/obo/{{ project.id }}/components/your-component-name.owl>)
+```
+5. Refresh your repo by running `sh run.sh make update_repo` - this should create a new file in src/ontology/components.
+6. In your custom makefile (src/ontology/{{ project.id }}.Makefile) add a goal for your custom make file as follows (for ROBOT template goals, please see section below): 
+
+```
+YOUR_COMPONENT_NAME = {insert how your component is made, for example a url it is downloaded from}
+$(COMPONENTSDIR)/your-component-name.owl: $(SRC)
+	$(ROBOT) annotate -I $(YOUR_COMPONENT_NAME) --ontology-iri $(ONTBASE)/$@ -o $@
```
matentzn, line 1392: This example is confusing, just add a simple working example instead.

matentzn (changes requested):

shawntanzk:
> This example is confusing, just add a simple working example instead.

Changed it to just use ROBOT template since that worked clearly with Ray

matentzn (approved):

PR #590, opened 2022-05-02
issue 589

### LINKED ISSUE #589: Add rdftab
https://github.com/ontodev/rdftab.rs

Allows fast loading of sqlite databases from RDF/XML files

cmungall:
- pre-datadict
- Adding rdftab, fixes #589 bupoing oaklib


Dockerfile
```diff
@@ -108,3 +108,8 @@ RUN wget -nv https://github.com/balhoff/relation-graph/releases/download/v$RG/re
 && tar -zxvf relation-graph-$RG.tgz \
 && mv relation-graph-$RG /tools/relation-graph \
 && chmod +x /tools/relation-graph
+
+# rdftab
+ENV RDFTAB=0.1.1
+RUN wget -nv https://github.com/ontodev/rdftab.rs/releases/download/v$(RSDTAB)/rdftab-x86_64-unknown-linux-musl -O /tools/rdftab \
```
gouttegd, line 114: This won’t work on M1-powered Macs.

There doesn’t seem to be [pre-built binaries](https://github.com/ontodev/rdftab.rs/releases/tag/v0.1.1) for rdftab targeting the arm64 architecture, so for M1 we will (again) need to build from source. Luckily we already have Rust development tools on the `odkbuilder` image (we need them to build Fastobo).

matentzn:
https://github.com/jamesaoverton/cmi-pb-terminology/blob/master/Makefile#L32

Dockerfile
jamesaoverton: We would be open to building and releasing an arm64 M1. Would it be as simple as adding to our GitHub Action here? https://github.com/ontodev/rdftab.rs/blob/master/.github/workflows/rust.yml#L31

gouttegd: Cargo seems to have built-in support for cross-compiling, so it could indeed be as simple as calling `cargo build --target aarch64-unknown-linux-musl`.

matentzn: @gouttegd Can you slowly slowly sheperd this PR through to completion? If not, let me know and I will try to find some time. Thank you!

gouttegd: @matentzn On it. :) I am currently testing building `rdftab` from source from within the `odkbuild` image (same approach as the one we are already using for `fastobo`). I’ll push an amendment to this PR as soon as the test is conclusive.

matentzn:
@jamesaoverton Can you check this one more time? @gouttegd updated the image build process using `cargo`. Hard for me to judge this, but of course, I will try it later.

@gouttegd are build tools in the build image discarded from the final image, or do they still exist down the line?

gouttegd:
@matentzn Build tools (in this case the Rust compiler and the Cargo package manager) are still present in the `odkbuild` image (that’s what this image is for), but not in the final `odkfull` image. Only the resulting binary is copied over from `odkbuild` to `odkfull`.

matentzn:
@gouttegd whats the state of this PR here?

gouttegd:
@matentzn From my side, it’s working and ready to merge. `rdftab` is built from source and available in the `odkfull` image. But I only tested that the resulting binary can successfully be invoked (it can). I did no functional testing, as I personally have no test case for `rdftab`.

PR #593, opened 2022-05-03
Build rdftab from source.

gouttegd:
This PR amends #590 to install `rdftab` by compiling it from source, as no pre-built `rdftab` binary for arm64 is available (yet?).

### DIFF
--- a/docker/builder/Dockerfile
+++ b/docker/builder/Dockerfile
@@ -118,6 +118,16 @@ RUN wget -nv https://github.com/fastobo/fastobo-validator/archive/refs/tags/v0.4
     cd /build && \
     rm -rf fastobo-validator-0.4.0 fastobo-validator-0.4.0.tar.gz /root/.cargo
 
+# Compile rdftab.
+RUN wget -nv https://github.com/ontodev/rdftab.rs/archive/refs/tags/v0.1.1.tar.gz \
+        -O /build/rdftab-0.1.1.tar.gz && \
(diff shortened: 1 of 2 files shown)


docker/builder/Dockerfile
```diff
@@ -118,6 +118,16 @@ RUN wget -nv https://github.com/fastobo/fastobo-validator/archive/refs/tags/v0.4
     cd /build && \
     rm -rf fastobo-validator-0.4.0 fastobo-validator-0.4.0.tar.gz /root/.cargo
 
+# Compile rdftab.
+RUN wget -nv https://github.com/ontodev/rdftab.rs/archive/refs/tags/v0.1.1.tar.gz \
```
matentzn, line 122: Can we extract the version here as we usually do? Just to make updating to later versions easier.

matentzn: $RDFTAB_VERSION

gouttegd: From my experience it’s not that useful when building from source, as updating to a later version most often requires some work anyway beyond merely changing the version number. But sure.

matentzn: Usually I agree, but because of the git diff not showing lines around, it is very hard to review a pull request "bumped version number of X" - you always then have to "view file" if you want to be sure they bumped it in all places..

matentzn (approved):
This is hard for me to judge, but definitely ready to be merged into the other PR!

matentzn:
Thank you @gouttegd looks great!

PR #594, opened 2022-05-06
Improving run.sh.jinja2

matentzn:
- Warning to update ODK
- Making memory parameter configurabe

### DIFF
--- a/Changes.md
+++ b/Changes.md
@@ -10,6 +10,7 @@
 - Technical
    - Run internal `make` and $(MAKE) to ensure that parameters are propagated correctly ([issue](https://github.com/INCATools/ontology-development-kit/issues/554)).
    - Mirror goal is only run when mirrors have been downloaded ([pull](https://github.com/INCATools/ontology-development-kit/pull/570))
+   - src/ontology/run.sh can now be configured with a MEMORY_JAVA parameter. It also warns the user in certain cases to update their local ODK image.
 - Documentation
    - Some improved documentation on components ([pull](https://github.com/INCATools/ontology-development-kit/pull/577))
    - All ODK-related general docs now moved to https://oboacademy.github.io/obook.
--- a/Changes.md
+++ b/Changes.md
@@ -10,7 +10,8 @@
 - Technical
    - Run internal `make` and $(MAKE) to ensure that parameters are propagated correctly ([issue](https://github.com/INCATools/ontology-development-kit/issues/554)).
    - Mirror goal is only run when mirrors have been downloaded ([pull](https://github.com/INCATools/ontology-development-kit/pull/570))
-   - src/ontology/run.sh can now be configured with a MEMORY_JAVA parameter. It also warns the user in certain cases to update their local ODK image.
+   - src/ontology/run.sh can now be configured with a ODK_JAVA_OPTS parameter at runtime, rather than just with a `robot_java_args` option in the projects ODK config file. This allows to more easily run releases across environments with different levels of memory. Example: `ODK_JAVA_OPTS=-Xmx4G ODK_DEBUG=yes ./run.sh make odkversion`
+   - src/ontology/run.sh warns the user in certain cases to update their local ODK image.
 - Documentation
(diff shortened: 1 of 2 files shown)


matentzn:
The if [[]] statement in the end seems to be syntactically off, my atom gives me warnings about it:

image


But it works. If you have ideas how to do better I am all ears. :)

template/src/ontology/run.sh.jinja2
```diff
@@ -11,15 +11,22 @@
 # See README-editors.md for more details.
 
 IMAGE=${IMAGE:-odkfull}
-
+MEMORY_GB=${MEMORY_GB:-8}
```
gouttegd, line 14: Since this `MEMORY_GB` parameter can be taken from the environment (contrary to `MEMORY_JAVA` which is taken from the ODK configuration file), it should have a name that more explicitly ties it the ODK, e.g. `ODK_MEMORY_GB` (similarly to `ODK_DEBUG`).

Changes.md
```diff
@@ -10,6 +10,7 @@
 - Technical
    - Run internal `make` and $(MAKE) to ensure that parameters are propagated correctly ([issue](https://github.com/INCATools/ontology-development-kit/issues/554)).
    - Mirror goal is only run when mirrors have been downloaded ([pull](https://github.com/INCATools/ontology-development-kit/pull/570))
+   - src/ontology/run.sh can now be configured with a MEMORY_JAVA parameter. It also warns the user in certain cases to update their local ODK image.
```
gouttegd, line 13: This does not accurately reflect the way the configuration works in the `run.sh` script.

The max memory can be configured with a `robot_java_args` parameter in the ODK configuration file, _or_ with a `MEMORY_GB` (which should be renamed to `ODK_MEMORY_GB`, see other comment) environment variable.

template/src/ontology/run.sh.jinja2
```diff
@@ -11,15 +11,22 @@
...
+MEMORY_JAVA={% if project.robot_java_args is defined %}{{ project.robot_java_args }}{% else %}-Xmx${MEMORY_GB}G{% endif %}
```
gouttegd, line 15: Here, the value defined in the configuration file takes precedence from what is specified in the environment. I think it should be the other way round. It should be possible to occasionally override the value from the configuration file with a value from the environmemt.

gouttegd (changes requested):

template/src/ontology/run.sh.jinja2
```diff
@@ -11,15 +11,22 @@
...
     # prevent the shell from tokenizing the format string.
+    echo "Running ${IMAGE} with ${MEMORY_JAVA} of memory for ROBOT and Java-based pipeline steps."
     TIMECMD="/usr/bin/time -f ### DEBUG STATS ###\nElapsed time: %E\nPeak memory: %M kb"
 fi
 
-docker run -v $PWD/../../:/work -w /work/src/ontology {% if project.robot_java_args is defined %}-e ROBOT_JAVA_ARGS='{{ project.robot_java_args }}' -e JAVA_OPTS='{{ project.robot_java_args }}'{% endif %} {% if project.use_external_date is sameas True %}-e TODAY=`date +%Y-%m-%d` {% endif %}--rm -ti obolibrary/$IMAGE $TIMECMD "$@"
+docker run -v $PWD/../../:/work -w /work/src/ontology -e ROBOT_JAVA_ARGS="$MEMORY_JAVA" -e JAVA_OPTS="$MEMORY_JAVA" {% if project.use_external_date is sameas True %}-e TODAY=`date +%Y-%m-%d` {% endif %}--rm -ti obolibrary/$IMAGE $TIMECMD "$@"
+
+
+if [[ "$@" == *"update_repo"* || "$@" == *"release"*  ]]; then
```
gouttegd, line 30: I’d rather use

```case "$@" in
*update_repo*|*release*)
    ...
    ;;
esac
```

which AFAIK is POSIX-compliant.

Changes.md
matentzn: Updated text

gouttegd (approved):

PR #603, opened 2022-05-20
Fix argument passing in the Windows run.bat script.

### LINKED ISSUE #583: ODK failing on Windows
I have  a user that tried to use the ODK on Windows when they try to do a `run.bat make all`, they get `Could not find or load main class '-Xmx8G'`. See screenshot below:

image

Did she use the correct commands? Or is this indeed an ODK issue?

gouttegd:
Windows has a peculiar way of interpreting quotes in command line arguments: if the opening quote is not at the beginning of the argument, then it is deemed to be part of the argument.

That is, with

```
ARGUMENT='something'
```

the `ARGUMENT` variable is assigned a value of `'something'`, *including* the quotes.

To make sure the quotes are only used, well, for quoting, and do not end up in the assigned value, the opening quote must be the first character of the argument, as in:

```
'ARGUMENT=something'
```

This results in `ARGUMENT` being correctly assigned the value `something`, *without* quotes.

closes #583


matentzn:
WOW that was it? You tested it on a windows machine?

gouttegd:
I got temporary access to a Windows machine and I could at least test that the shell does behave as described, which makes me pretty confident this is the cause of #583.

I did *not* test the ODK itself on that machine, though – I though it would have been rude to install the ODK on a machine that a friend let me use after I asked him “can I quickly check something on your computer”… :D

matentzn (approved):

PR #604, opened 2022-05-20
Upgrade base system

### LINKED ISSUE #591: bump oaklib version
I tried to bump up to 0.1.12 in #590 this seemed to cause issues..

### LINKED ISSUE #601: Upgrading the base system?
The ODK is currently built on top on Ubuntu 20.04, which is more than 2 years old.

As a result, we’re going to see the first problems caused by the age of the underlying system, such as #591 (`oaklib` being stuck to version 0.1.2 because more recent versions require Python 3.9 while we only have Python 3.8).

We should consider moving to a more recent Ubuntu base (22.04?).

gouttegd:
Upgrade the base system for all images to Ubuntu 22.04.

Among other things, this brings us Python 3.10 (and would therefore automatically unblock and fix #591). This also removes the need for compiling SWI-Prolog from source, allowing us to simplify the build process a little bit.

closes #601


matentzn:
Absolutely awesome. How far did you go in terms of testing?

gouttegd:
Tested against the FBbt release pipeline, which ran smoothly. Didn’t run the full test suite on the ODK itself, since we have the GitHub CI for that. :p

matentzn (approved):

PR #611, opened 2022-05-27
Removing History page and add nice default config for mkdocs

### LINKED ISSUE #605: Add some of the cool mkdocs setting by default
see SSSOM: 

https://github.com/mapping-commons/sssom/blob/master/mkdocs.yml

- making pages editable
- links to github
- awesome search

matentzn:
fixes #605

- No one is ever writing anything in the history.md so I will remove it from the default. Anyone that wants to can just drop it back in the usual way.
- Adding some cool features to the default mkdocs, to allow for things like better searching, editing straight from the docs etc. See https://mapping-commons.github.io/sssom/ for an example deployment which has all these setting activated.


matentzn:
@shawntanzk I will merge now but feel free to come back and complain! :)

PR #612, opened 2022-06-09
Add break lines to makefile

### LINKED ISSUE #553: Format the Makefile a bit nicer using line breaks \
From @gouttegd: 

> Cosmetic only: I’d like to generalise the use of line breaks (with \) to avoid overly long lines. We already do that for some rules, but not for all.

anitacaron:
Fixes #553


matentzn:
Thank you! What line break rule did you apply? What is the maximum line break length now that is permitted?

matentzn:
Also did you try building an example to see if you didn't remove too much whitespace?

anitacaron:
> What line break rule did you apply? What is the maximum line break length now that is permitted? 

I put line breaks when the command line was more extensive than the general commands and when there was more than one command in the same line. I was looking into a real Makefile (UBERON).

> Also did you try building an example to see if you didn't remove too much whitespace?

Looking better in the result, I'll fix some whitespaces.

matentzn:
Try this on PATO and CL - there are a lot of empty lines generated in the resulting `Makefile`!

matentzn (approved):
I think its fine, but I wanted a brief look of @gouttegd when he is back just in case.

matentzn:
I will risk it.

PR #617, opened 2022-06-16
Make Base IRI configurable

matentzn:


odk/odk.py
```diff
@@ -552,6 +552,10 @@ class OntologyProject(JsonSchemaMixin):
     uribase : str = "http://purl.obolibrary.org/obo"
     """Base URI for PURLs. DO NOT MODIFY AT THIS TIME, code is still hardwired for OBO """
```
gouttegd, line 553: I guess this should be updated? `uribase` *can* be modified now, that’s even the entire point of this PR? Or do we want to allow modifying it while still strongly discouraging the practice?

gouttegd (approved):

template/_dynamic_files.jinja2
```diff
@@ -1125,6 +1124,59 @@ extract -T imports/ro_terms_combined.txt --force true --individuals exclude --me
...
+    - filename: mycomp.owl
+```
+
+When running `sh run.sh make update_repo`, a new file `src/ontology/components/mycomp.owl` will 
+be created which you can edit as you see fit. Typical ways to edit:
+
+1. Using a ROBOT template to generate the component (see below)
+1. Manually curating the component separately with Protege or any other editor
+1. Providing a `components/mycomp.owl:` make target in `src/ontology/{{ project.id }}.Makefile`
+and provide a custom command to generate the component.
```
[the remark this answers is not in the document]
matentzn: Yep, agreed. Made a warning comment.

odk/odk.py
matentzn: Updated documentation. No I dont want to discourage it. I want to widen the scope of ODK so that non OBO ontologies are supported the same way as OBO ones!

PR #635, opened 2022-07-13
Add docker requirement to top-level README

kltm:
See https://github.com/berkeleybop/bbops/issues/26


matentzn (approved):
Thank you @kltm !

PR #640, opened 2022-08-02
Enable use of singularity as an option

pcm32:
This PR gives the resulting run.sh script the ability to run with singularity as an option to docker, facilitating execution in HPC environments.

### DIFF
--- a/template/src/ontology/run.sh.jinja2
+++ b/template/src/ontology/run.sh.jinja2
@@ -9,14 +9,18 @@
 # we therefore map the whole repo (../..) to a docker volume.
 #
 # To use singularity instead of docker, please issue
-#
 # export USE_SINGULARITY=<any-value>
-#
 # before running this script.
 #
 # See README-editors.md for more details.
 
 IMAGE=${IMAGE:-odkfull}
+TAG_IN_IMAGE=$(echo $IMAGE | awk -F':' '{ print $2 }')
+if [ "$TAG_IN_IMAGE" ]; then
+  # Override TAG env var if IMAGE already includes a tag
+  TAG=$TAG_IN_IMAGE
+  IMAGE=$(echo $IMAGE | awk -F':' '{ print $1 }')
+fi
 TAG=${TAG:-latest}
 ODK_JAVA_OPTS={% if project.robot_java_args is defined and project.robot_java_args|length %}${ODK_JAVA_OPTS:-{{ project.robot_java_args }}}{% else %}${ODK_JAVA_OPTS:--Xmx8G}{% endif %}
 ODK_DEBUG=${ODK_DEBUG:-no}
@@ -37,7 +41,11 @@ if [ -n "$USE_SINGULARITY" ]; then
     export ROBOT_JAVA_ARGS="$ODK_JAVA_OPTS"
     export JAVA_OPTS="$ODK_JAVA_OPTS"
     {% if project.use_external_date is sameas True %}export TODAY=$(date +%Y-%m-%d){% endif %}
-    singularity exec --bind $VOLUME_BIND -W $WORK_DIR docker://obolibrary/$IMAGE:$TAG $TIMECMD "$@"
+    singularity exec --cleanenv \
+        --env "ROBOT_JAVA_ARGS=$ODK_JAVA_OPTS,JAVA_OPTS=$ODK_JAVA_OPTS" \
+        --bind $VOLUME_BIND \
+        -W $WORK_DIR \
+        docker://obolibrary/$IMAGE:$TAG $TIMECMD "$@"
 else
     docker run -v $VOLUME_BIND -w $WORK_DIR -e ROBOT_JAVA_ARGS="$ODK_JAVA_OPTS" -e JAVA_OPTS="$ODK_JAVA_OPTS" {% if project.use_external_date is sameas True %}-e TODAY=`date +%Y-%m-%d` {% endif %}--rm -ti obolibrary/$IMAGE $TIMECMD "$@"
 fi
--- a/template/src/ontology/run.sh.jinja2
+++ b/template/src/ontology/run.sh.jinja2
@@ -38,8 +38,6 @@ VOLUME_BIND=$PWD/../../:/work
 WORK_DIR=/work/src/ontology
 
 if [ -n "$USE_SINGULARITY" ]; then
-    export ROBOT_JAVA_ARGS="$ODK_JAVA_OPTS"
-    export JAVA_OPTS="$ODK_JAVA_OPTS"
     {% if project.use_external_date is sameas True %}export TODAY=$(date +%Y-%m-%d){% endif %}
     singularity exec --cleanenv \
         --env "ROBOT_JAVA_ARGS=$ODK_JAVA_OPTS,JAVA_OPTS=$ODK_JAVA_OPTS" \
--- a/template/src/ontology/run.sh.jinja2
+++ b/template/src/ontology/run.sh.jinja2
@@ -8,9 +8,16 @@
 # The assumption is that you are working in the src/ontology folder;
 # we therefore map the whole repo (../..) to a docker volume.
 #
+# To use singularity instead of docker, please issue
+#
+# export USE_SINGULARITY=<any-value>
+#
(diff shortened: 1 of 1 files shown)


template/src/ontology/run.sh.jinja2
```diff
@@ -23,10 +30,20 @@ if [ x$ODK_DEBUG = xyes ]; then
     TIMECMD="/usr/bin/time -f ### DEBUG STATS ###\nElapsed time: %E\nPeak memory: %M kb"
 fi
 
-docker run -v $PWD/../../:/work -w /work/src/ontology -e ROBOT_JAVA_ARGS="$ODK_JAVA_OPTS" -e JAVA_OPTS="$ODK_JAVA_OPTS" {% if project.use_external_date is sameas True %}-e TODAY=`date +%Y-%m-%d` {% endif %}--rm -ti obolibrary/$IMAGE $TIMECMD "$@"
+VOLUME_BIND=$PWD/../../:/work
+WORK_DIR=/work/src/ontology
+
+if [ -n "$USE_SINGULARITY" ]; then
+    export ROBOT_JAVA_ARGS="$ODK_JAVA_OPTS"
```
matentzn, line 37: I am not very familiar with singularity.. So contrary to `docker`, you set the environment variables in the main environment, and somehow the singularity container has access to these globally set variables? In ODK its necessary to set these parameters inside the running container..

template/src/ontology/run.sh.jinja2
```diff
@@ -23,10 +30,20 @@ if [ x$ODK_DEBUG = xyes ]; then
...
+    export JAVA_OPTS="$ODK_JAVA_OPTS"
+    {% if project.use_external_date is sameas True %}export TODAY=$(date +%Y-%m-%d){% endif %}
+    singularity exec --bind $VOLUME_BIND -W $WORK_DIR docker://obolibrary/$IMAGE:$TAG $TIMECMD "$@"
```
matentzn, line 40: Have you tested this?

template/src/ontology/run.sh.jinja2
```diff
@@ -8,9 +8,16 @@
 # The assumption is that you are working in the src/ontology folder;
 # we therefore map the whole repo (../..) to a docker volume.
 #
+# To use singularity instead of docker, please issue
+#
+# export USE_SINGULARITY=<any-value>
+#
+# before running this script.
+#
```
matentzn, line 16: Collapse the above to one row of comment.

template/src/ontology/run.sh.jinja2
```diff
@@ -23,10 +30,20 @@ if [ x$ODK_DEBUG = xyes ]; then
...
+VOLUME_BIND=$PWD/../../:/work
+WORK_DIR=/work/src/ontology
+
+if [ -n "$USE_SINGULARITY" ]; then
+    export ROBOT_JAVA_ARGS="$ODK_JAVA_OPTS"
+    export JAVA_OPTS="$ODK_JAVA_OPTS"
+    {% if project.use_external_date is sameas True %}export TODAY=$(date +%Y-%m-%d){% endif %}
+    singularity exec --bind $VOLUME_BIND -W $WORK_DIR docker://obolibrary/$IMAGE:$TAG $TIMECMD "$@"
+else
+    docker run -v $VOLUME_BIND -w $WORK_DIR -e ROBOT_JAVA_ARGS="$ODK_JAVA_OPTS" -e JAVA_OPTS="$ODK_JAVA_OPTS" {% if project.use_external_date is sameas True %}-e TODAY=`date +%Y-%m-%d` {% endif %}--rm -ti obolibrary/$IMAGE $TIMECMD "$@"
```
matentzn, line 42: Its odd that we do not use the $TAG variable for the docker run command - There is a tiny risk that a client is using:

```
IMAGE=odklite:1.2.3 ./run.sh make ...
```

Which will now cause the script to fail. 

Is adding the `$TAG` variable strictly necessary for the singularity case above?

matentzn (commented):
Hey @pcm32 This is great, thank you. I will ask for one or two more reviews here to be sure that we do not break anything.

template/src/ontology/run.sh.jinja2
pcm32: yes, the singularity call is working like this on the EBI cluster.

template/src/ontology/run.sh.jinja2
pcm32: Well, on all the singularity setups I have used in 4 different HPCs, they normally inherit all variables from the parent environment (which is sometimes problematic as well).

matentzn: Thanks for the clarification!

template/src/ontology/run.sh.jinja2
pcm32: Yes, singularity won't assume latest implicitly. We could add some checks to see if image already includes a TAG and we could then set TAG from there.

template/src/ontology/run.sh.jinja2
matentzn: (Just out of curisioty, _can_ you set th variables inside the container?)

template/src/ontology/run.sh.jinja2
pcm32: I haven't tested the ansible part though.

template/src/ontology/run.sh.jinja2
matentzn: Basically like: `if ':' not in IMAGE`, IMAGE=IMAGE+TAG`?

template/src/ontology/run.sh.jinja2
gouttegd: According to the documentation, yes. It should be possible to prevent Singularity from inheriting variables from the underlying system with `--cleanenv` and instead to explicitly pass _only_ the variables we need with `--env VAR1=value1,VAR2=value2`.

(I can’t test that today as I am on the move and my laptop is not properly configured to run Singularity.)

gouttegd: Actually, yes, this works:

```sh
singularity exec --cleanenv --env "ROBOT_JAVA_ARGS=$ODK_JAVA_OPTS,JAVA_OPTS=$ODK_JAVA_OPTS" --bind $VOLUME_BIND -W $WORK_DIR docker://obolibrary/$IMAGE:$TAG $TIMECMD "$@"
```

gouttegd (changes requested):
Very nice, I didn’t know you could use a Docker image directly with Singularity.

Could you change the way you pass environment variables to the Singularity container as discussed above (`--cleanenv --env "VAR1=value,VAR2=value`)? I don’t think it is a good idea to allow the container to get all environment variables straight from the host.

matentzn:
@pcm32 Just to get this merged in, can you address the review comments please? :) Thanks!

pcm32:
Apologies, I have been on leave for a few days, will get back to this as soon as I return.

matentzn:
no rush at all, take your time!

template/src/ontology/run.sh.jinja2
pcm32: Addressed in [d5d5513](https://github.com/INCATools/ontology-development-kit/pull/640/commits/d5d5513920b74e9729e7c3f173b9064f480d1621).

template/src/ontology/run.sh.jinja2
pcm32: Addressed in [d5d5513](https://github.com/INCATools/ontology-development-kit/pull/640/commits/d5d5513920b74e9729e7c3f173b9064f480d1621).

pcm32:
@matentzn @gouttegd this should be ready for merging after your comments, thanks for the reviews.

gouttegd (approved):
Looks all good to me.

template/src/ontology/run.sh.jinja2
matentzn: Why do we still need the `export` commands then?

pcm32: Yes, not needed, have removed them now.

matentzn (approved):

PR #653, opened 2022-08-11
Allow to only push `dev`-tagged images to Docker Hub.

### LINKED ISSUE #648: Adjust Makefile to allow pushing only :dev tag to dockerhub
If I only push `obolibrary/odkfull:dev` tag to dockerhub, will people using

```
docker pull obolibrary/odkfull
docker pull obolibrary/odkfull:latest
```

be unaffected @gouttegd ?

gouttegd:
This PR adds a new main target to the top-level Makefile: `publish-multiarch-dev`. This target builds multi-architecture images of `odkfull` and `odklite`,¹ tag them with a `dev` tag, and publish them to the Docker Hub.

Inside the `dev` images, the ODK has version number `$(VERSION)-dev`, to allow users who may use alternatively a `dev` image and a released image to know which one they are using at any time.

With `dev` tagged images (`obolibrary/odkfull:dev`), the `obolibrary/odkdev` image becomes redundant and this PR removes it. The `build-dev` target, which uses to build the `odkdev` image, now builds `odkfull:dev`.

closes #648 

¹ Building and pushing `odklite:dev` is necessary even if we only need `odkfull:dev`, since `odkfull` is built on top of `odklite`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -60,15 +60,15 @@ schema/project-schema.json:
 VERSION = "v1.3.2"
 IM=obolibrary/odkfull
 IMLITE=obolibrary/odklite
-DEV=obolibrary/odkdev
 ROB=obolibrary/robot
 #ROBOT_JAR="https://github.com/monarch-ebi-dev/odk_utils/raw/master/robot_maven_test.jar"
 ROBOT_JAR_ARGS=#--build-arg ROBOT_JAR=$(ROBOT_JAR)
+TAGS_OPTION=-t $(IM):$(VERSION) -t $(IM):latest
 
 build: build-odklite
 	docker build $(CACHE) --platform $(ARCH) \
 	    --build-arg ODK_VERSION=$(VERSION) $(ROBOT_JAR_ARGS) \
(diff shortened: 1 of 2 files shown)


matentzn:
Awesome! Truly amazing. Thank you so much @gouttegd!

if I run `make publish-multiarch-dev` today, are you confident I wont accidentally overwrite the current latest? I would like to try it right away.

gouttegd:
I couldn’t test that (the problem with the `-multiarch` targets is that by definition they suppose pushing someone to the Docker Hub — not something you can test on your own), but see for yourself the core of the `publish-multiarch-dev` recipe:

```sh
docker buildx build $(CACHE) --push --platform $(PLATFORMS) \
                --build-arg ODK_VERSION=$(VERSION)-dev \
                -t $(IM):dev \
                .
```

There’s no chance this could cause the publication of an image tagged with anything else than `dev`. If it does, it’s a bug in `docker buildx`, and I’ll decline all responsibility. :D

matentzn (approved):

PR #655, opened 2022-08-11
Rename IMAGE to ODK_IMAGE in the run.sh script.

gouttegd:
In the run.sh wrapper script, rename the `IMAGE` variable to `ODK_IMAGE`. This is for consistency with the other variables in that script such as `ODK_DEBUG` and `ODK_JAVA_OPTS`. Likewise for `TAG`, renamed to `ODK_TAG`.

This will allow ODK users/developers who always want to use a specific image to safely export `ODK_IMAGE=...` and/or `ODK_TAG=...` in in their profile scripts, without fearing that it could interfere with any other program that might look for a `IMAGE` or `TAG` environment variable.

Likewise for the `seed-via-docker.sh` script, again for consistency.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # v1.3.2 (unpublished, target date 31 July 2022)
 
+- Changes in the `src/ontology/run.sh` wrapper script:
+  - [It is now possible](https://github.com/INCATools/ontology-development-kit/pull/640) to execute the Docker image through [Singularity](https://apptainer.org).
+  - The `IMAGE` variable, which can used to specify an alternative ODK image, [has been renamed](https://github.com/INCATools/ontology-development-kit/pull/655) to `ODK_IMAGE`.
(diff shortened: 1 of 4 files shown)


matentzn (commented):
Isnt this change very risky? I feel like we could be breaking quite a bit of client code if we did that.. For example, I have an alias for `IMAGE=odklite ODK_DEBUG=yes ./run.sh make update_repo` which would have to be updated. Otherwise, the risk is that a different image is used than the intended one if someone just updates their repo?

gouttegd:
What’s the risk? The default image remains `odkfull`.

So the worst thing that could happen, if you have been using the `IMAGE` variable to force the use of `odklite` (as in your example), is that this will now be ignored, and you will run your `update_repo` with the default `odkfull` image instead of `odklite`. `odkfull` being strictly a superset of `odklite`, this cannot break anything.

And if people have started using the `IMAGE` parameter to do this kind of things (which is good: this is why this parameter exists), I’d rather have them change to `ODK_IMAGE` sooner rather than later.

matentzn (approved):
Ok I trust your judgement :) make sure you make a note in the CHANGELOG to announce the change

gouttegd:
> I trust your judgement

Maybe you shouldn’t! There was a mistake in the new `run.sh` script, one instance of `IMAGE` had not been renamed as it should have been. 🤦 Fixed now.

Issue #370, opened 2020-11-03
Provide better documentation of config options


dosumis:
It is currently hard to work out what config options do except by looking at examples.  I was hoping that [schema](https://github.com/INCATools/ontology-development-kit/blob/master/schema/project-schema.json) would be usable for this, but descriptions are largely empty and it seems from the [README](https://github.com/INCATools/ontology-development-kit/tree/master/schema) this is just autogenerated from [odk.py](https://github.com/INCATools/ontology-development-kit/blob/master/odk/odk.py).

Could we have a complete schema with descriptions?  There are some easy ways to convert JSON schema to browsable html.

Motivating use-case:  setting up a repo with Robot templates & related build infrastructure.  Can I use `robotemplate_group` for that?

matentzn:
Yeah, sounds reasonable! I added this ticket to our meeting agenda as well. Providing better documentation has been an open ticket here for 3 years :D I think we should be able to generate a schema, but you can also look at the file called "project.yaml" when you generate any project (in the target dir, for example: `target/pato/project.yaml`). This seems to have most if not all options:

```
allow_equivalents: all
catalog_file: catalog-v001.xml
components:
  directory: components
  filename: null
  products:
  - filename: pato_ext.owl
    source: null
  source: null
contact: null
contributors: null
creators: null
description: None
dosdp_tools_options: --obo-prefixes=true
edit_format: obo
exclude_tautologies: structural
export_formats:
- owl
- obo
- json
git_user: ''
github_org: pato-ontology
gzip_main: false
id: pato
import_group:
  directory: imports/
  disabled: false
  ids: null
  products:
  - description: null
    id: ro
    mirror_from: null
    rebuild_if_source_changes: true
    robot_settings: null
  - description: null
    id: chebi
    mirror_from: http://purl.obolibrary.org/obo/chebi.owl.gz
    rebuild_if_source_changes: true
    robot_settings: null
  - description: null
    id: go
    mirror_from: null
    rebuild_if_source_changes: true
    robot_settings: null
  - description: null
    id: pco
    mirror_from: null
    rebuild_if_source_changes: true
    robot_settings: null
  - description: null
    id: uberon
    mirror_from: null
    rebuild_if_source_changes: true
    robot_settings: null
  rebuild_if_source_changes: true
import_pattern_ontology: false
license: https://creativecommons.org/licenses/unspecified
namespaces: null
obo_format_options: ''
pattern_group: null
pattern_pipelines_group: null
[...]

PR #656, opened 2022-08-11
Schema config documentation

hkir-dev:
Related with issue #370

New documentation: https://github.com/INCATools/ontology-development-kit/blob/issue-370/docs/project-schema.md

Generates ODK schema config documentation from the 'project-schema.json'

Schema json doesn't contain field descriptions so extracts those descriptions from 'odk.py' using ast.

Generates the initial (plain) documentation using jsonschema2md library. Then makes documentation more readable through
in-lining referred objects and 'allOf' definitions.


cmungall:
looks good!

Did we already consider linkml for this?

matentzn:
No we are still using your old odk.py for generating schema after the fact. The source of truth are the python data classes. While you could probably generate a linkml model from the json generated from the data classes, it would take a long time to replace the whole pipeline with a purely linkml-based approach. It would be cool of course, especially since I would like to enable other ontology tools like Protege or DROID to refer to this as a standard representation of onto repo config, but.. one urgent plaster on a bloody patch at a time.

requirements.txt
```diff
@@ -35,3 +35,4 @@ tsvalid
 j2cli[yaml]
```
matentzn, line 35: Rename file back to what it was, as the original name is needed to build ODK!

matentzn (changes requested):
Excellent starting point I think! After fixing the one issue, I will mere this is. What do you think @gouttegd - Given there was _nothing_ at all like this in ODK previously, I think this is a huge step forward!

@hkir-dev can you also just publish the markdown straight in the docs/ directory please?

requirements.txt
gouttegd: FYI, the reason why this file is not named `requirements.txt` is given in commit 3fdd6bf1ada6c4e3ee9b2c7190e2660a69bd7732.

gouttegd:
@matentzn _Some_ documentation is always better than _no_ documentation. :) Looks good for me, modulo the `requirements.txt.full` issue.

matentzn:
Thanks @gouttegd :)

requirements.txt
hkir-dev: Thank you, seems I forgot to restore the file name when development finished.And updated default output location to docs/ directory now.

gouttegd (approved):

matentzn (approved):

matentzn:
Thank you @hkir-dev super cool.

PR #657, opened 2022-08-13
Version 1.3.2: JUST TRACKING CHANGES

matentzn:


odk/odk.py
```diff
@@ -507,7 +507,13 @@ class OntologyProject(JsonSchemaMixin):
     
     use_mappings : bool = False
     """if true use SSSOM mapping files."""
+
+    set_docker_user : bool = False
+    """if true continue to use your own user when using docker, instead of the default docker user."""
```
gouttegd, line 512: Making that configurable at the ODK level is a bad idea IMO. This bakes the setting into the generated `run.sh` file, and so assumes all editors of the ontology need the same setting (either they all need to run Docker with their own privileges or none of them does). This will not always be the case, if for example one editor is working under macOS and another is working on GNU/Linux.

It can even happen that the _same_ editor is sometimes using macOS and sometimes using GNU/Linux: I do most of my ontology editing work on macOS, but for FBcv, I have to do the release on GNU/Linux, because my macOS machine does not have enough memory.

If we make that configurable at the ODK level, it would be better to invert the default: the default should be to run with the user’s credentials, and only if an ontology has a workflow that requires the ability to install supplementary packages should that setting be changed.

template/src/ontology/run.sh.jinja2
```diff
@@ -65,18 +65,18 @@ ROBOT_JAVA_ARGS=$ODK_JAVA_OPTS
...
+        --env-file run.sh.env \{% endif %}
         --bind $VOLUME_BIND \
         -W $WORK_DIR \
         docker://obolibrary/$ODK_IMAGE:$ODK_TAG $TIMECMD "$@"
 else
     BIND_OPTIONS="-v $(echo $VOLUME_BIND | sed 's/,/ -v /')"
-    docker run -u $(id -u):$(id -g) $ODK_DOCKER_OPTIONS $BIND_OPTIONS -w $WORK_DIR \
-        --env-file run.sh.env \
+    docker run {% if project.set_docker_user is sameas True %}-u $(id -u):$(id -g) {% endif %}$ODK_DOCKER_OPTIONS $BIND_OPTIONS -w $WORK_DIR \{% if project.use_env_file_docker is sameas True %}
+        --env-file run.sh.env \{% endif %}
```
gouttegd, line 79: So if `use_env_file_docker` is False (which is the default), _no_ environment variables are passed to the container, which means that `ROBOT_JAVA_ARGS` (which is used to set the amount of memory ROBOT can use) is not passed either. Is this really what we want?

matentzn: s**t didn't think of this.. This ODK release will never happen.

I just cant have so many of my pipelines breaking now with introducing a non-configurable option.. Literally _all_ pipelines I run have a small requirements.txt install step associated with it now. I don't know what else to do - I don't want other peoples stuff to break now the same way, especially since it is really hard to debug this.

odk/odk.py
matentzn: I would agree with this but it introduces a breaking change. It literally took me hours, or days, to figure out why my `oak` pipelines stopped working (oak is trying to cache data in the user directory `~/.data`, which does not exist when you go in with `-u`). Literally half (or more) of my pipelines have some kind of requirements.txt step in them - and they just fail with obscure errors. We should avoid introducing breaking changes as much as possible..

template/src/ontology/run.sh.jinja2
gouttegd: > Literally all pipelines I run have a small requirements.txt install step associated with it now. 

How is that related to avoiding the use of the `run.sh.env` file?

matentzn: Sorry, I read your comment wrong. This one I had to make optional because for some reason, sometimes my docker crashes / the container does not start, and removing the run.sh.env logic causes this problem not to happen -- I don't have a clue how they relate though. It makes no sense. I just guessed that maybe it has something to do with the `trap` but maybe it is entirely unrelated..

odk/odk.py
gouttegd: Solution on its way in #768.

matentzn: We dropped this from 1.4 and I moved your PR to 1.5

matentzn (commented):
Ok @anitacaron and I are done reviewing this, I will still run a few tests, but all the big ones look good to me.

PR #661, opened 2022-08-17
Build `odkfull:dev` from `odklite:dev`.

### LINKED ISSUE #659: The `odkfull:dev` image is built from `odklite:latest`
The `build-dev` rule (which builds a local `odkfull:dev` image) and the `publish-multiarch-dev` rule (which builds and pushes the multi-architecture `odkfull:dev` image) doesn’t do what it says on the tin.

Even when those rules are used, the `odkfull` image is always built on top of the `odklite:latest` image (_not_ `odklite:dev`), because that’s hardcoded within the main Dockerfile.

So, what happens when one runs `make build-dev` is that:
* `odkbuild:latest` will be built (normal);
* `odklite:dev` will be built, using the just built `odkbuild:latest` (normal);
* `odkfull:dev` will be built, using a previously built `odklite:latest` (WRONG) — or, if no `odklite:latest` image is available in the local cache, a `odklite:latest` freshly downloaded from the Docker Hub (also WRONG).

(Same for the multiarch version.)

As the result, the `odkfull:dev` image, supposed to represent the “bleeding-edge” state of the current branch, will be built on top of a `odklite:latest` image that may be up to several months old!

The fix should be to stop hardcoding the `FROM` clause in the main Dockerfile:

```Dockerfile
FROM obolibrary/odklite:latest
```

The tag from which we build `odkfull` should instead be passed from the Makefile, so that we can choose to build from `odklite:dev` when we are supposed to build the `-dev` version.

gouttegd:
Update the main Dockerfile to allow to specify a different `odklite` version to build `odkfull` from, instead of always building from `odklite:latest`.

Update the main Makefile so that the `build-dev` and `publish-multiarch-dev` rules build the `odkfull:dev` image from `odklite:dev`, so that the resulting image does contain all the latest changes as intended.

closes #659

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,7 @@
 # Final ODK image
 # (built upon the odklite image)
-FROM obolibrary/odklite:latest
+ARG ODKLITE_TAG=latest
+FROM obolibrary/odklite:${ODKLITE_TAG}
 LABEL maintainer="obo-tools@googlegroups.com"
 
 ENV PATH "/tools/apache-jena/bin:/tools/sparqlprog/bin:$PATH"
--- a/Makefile
+++ b/Makefile
@@ -88,7 +88,12 @@ build-odklite-dev: build-builder
 	$(MAKE) TAGS_OPTION="-t $(IMLITE):dev" VERSION=$(VERSION)-dev build-odklite
 
 build-dev: build-odklite-dev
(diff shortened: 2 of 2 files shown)


matentzn (approved):

matentzn:
Awesome, thank you @gouttegd!

gouttegd:
It would have been more awesome if I had caught that before merging PR #653, but better late than never I guess. :D

PR #662, opened 2022-08-18
Add constraints.txt make goal

matentzn:


gouttegd (changes requested):
It works, but it’s up to whoever is running the command to make sure the Python version they’re using is the same major Python version as the one provided within the ODK. If you’re running it with an older Python, you could end up being stuck with old versions of packages (as has happened here with `oaklib`); if you’re running it with a newer version, you could end up with packages that won’t work within the ODK…

I didn’t want to do that, but I think the best way to make sure we are always running the correct version of Python is to run this command _from within the ODK container_. I’ll work on that.

PR #663, opened 2022-08-18
Update constraints from within the `odkbuild` image

gouttegd:
Updated version of PR #662 

This runs the updating command from within the `odkbuild` image, which ensures we are always using the same Python version as the one available within the ODK itself.

This is slightly inefficient as, to update the constraints, we then need to build the `odkbuild` image itself if it is not already available, update the `constraints.txt` file, then build again `odkbuild` to use the new constraints. But as we only update the constraints infrequently, it shouldn’t be much of an issue.


matentzn (approved):
My appreciation know no bounds. THANK YOU!

PR #672, opened 2022-08-27
fixed parentheses in inject queries

ehartley:
Moved ")" so that the last regex function is not inside the previous regex function.


matentzn (approved):

matentzn:
Excellent, great spot. I would have missed this and released this into production for sure!

PR #676, opened 2022-08-31
`make`: `component-download-mondo`

joeflack4:
Updates
```
- Update: template/src/ontology/Makefile.jinja2: Added phony target `component-download-mondo`.
```

### DIFF
--- a/template/src/ontology/Makefile.jinja2
+++ b/template/src/ontology/Makefile.jinja2
@@ -469,6 +469,11 @@ $(COMPONENTSDIR)/%: | $(COMPONENTSDIR)
 	touch $@
 .PRECIOUS: $(COMPONENTSDIR)/%
 
+.PHONY: component-download-mondo.owl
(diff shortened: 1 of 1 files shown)


template/src/ontology/Makefile.jinja2
```diff
@@ -469,6 +469,11 @@ $(COMPONENTSDIR)/%: | $(COMPONENTSDIR)
 	touch $@
 .PRECIOUS: $(COMPONENTSDIR)/%
 
+.PHONY: component-download-mondo.owl
```
joeflack4, line 472: FYI: The change.

matentzn: This does not make too much sense for ODK - this has to go into the custom makefile of your project!

matentzn (changes requested):
See comments - at this state, this PR is in the wrong repo!

PR #678, opened 2022-08-31
Allow extra volume bindings and custom wrapper config file

### LINKED ISSUE #675: Make it easier to pass additional volume bindings into run.sh
This way, we do not need to hack run.sh when we need to, for example, bind config files to tools which are not located in the working directory.

gouttegd:
This PR:

a) Adds a new `ODK_BINDS` variable to the `run.sh` wrapper script, which can be used to specify extra volume bindings within the Docker or Singularity container. Use as follows:

```sh
ODK_BINDS=/path/to/host/dir1:/path/to/container/dir1[:options][,...] sh run.sh ...
```

b) Adds a new `ODK_DOCKER_OPTIONS` variable, to allow user to pass any custom options to the invocation of `docker run`. Another variable, `ODK_SINGULARITY_OPTIONS`, does the same for the invocation of `singularity exec`.

c) Allows to set variables for the `run.sh` wrapper in a file named `run.sh.conf` (in the same directory as the wrapper script itself). That file should contain standard Bourne-shell variable declarations, and will be sourced (if it exists) by the wrapper script. It allows to use variables such as the new `ODK_BINDS`, `ODK_DOCKER_OPTIONS`, but also the pre-existing `ODK_IMAGE` or `ODK_TAG`, once and for all in the `run.sh.conf` file, instead of having to export them to the user’s environment or set them every time the wrapper script is invoked.

closes #675

### DIFF
--- a/template/src/ontology/run.sh.jinja2
+++ b/template/src/ontology/run.sh.jinja2
@@ -43,14 +43,17 @@ fi
 
 if [ -n "$USE_SINGULARITY" ]; then
     {% if project.use_external_date is sameas True %}export TODAY=$(date +%Y-%m-%d){% endif %}
-    singularity exec --cleanenv \
+    singularity exec --cleanenv $ODK_SINGULARITY_OPTIONS \
         --env "ROBOT_JAVA_ARGS=$ODK_JAVA_OPTS,JAVA_OPTS=$ODK_JAVA_OPTS" \
         --bind $VOLUME_BIND \
         -W $WORK_DIR \
         docker://obolibrary/$ODK_IMAGE:$ODK_TAG $TIMECMD "$@"
 else
     BIND_OPTIONS="-v $(echo $VOLUME_BIND | sed 's/,/ -v /')"
-    docker run $BIND_OPTIONS -w $WORK_DIR -e ROBOT_JAVA_ARGS="$ODK_JAVA_OPTS" -e JAVA_OPTS="$ODK_JAVA_OPTS" {% if project.use_external_date is sameas True %}-e TODAY=`date +%Y-%m-%d` {% endif %}--rm -ti obolibrary/$ODK_IMAGE:$ODK_TAG $TIMECMD "$@"
+    docker run $ODK_DOCKER_OPTIONS $BIND_OPTIONS -w $WORK_DIR \
+        -e ROBOT_JAVA_ARGS="$ODK_JAVA_OPTS" -e JAVA_OPTS="$ODK_JAVA_OPTS" \
+        {% if project.use_external_date is sameas True %}-e TODAY=`date +%Y-%m-%d` {% endif %} \
+        --rm -ti obolibrary/$ODK_IMAGE:$ODK_TAG $TIMECMD "$@"
 fi
 
 case "$@" in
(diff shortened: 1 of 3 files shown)


template/src/ontology/run.sh.jinja2
```diff
@@ -37,15 +41,23 @@ fi
...
     {% if project.use_external_date is sameas True %}export TODAY=$(date +%Y-%m-%d){% endif %}
-    singularity exec --cleanenv \
+    singularity exec --cleanenv $ODK_SINGULARITY_OPTIONS \
         --env "ROBOT_JAVA_ARGS=$ODK_JAVA_OPTS,JAVA_OPTS=$ODK_JAVA_OPTS" \
         --bind $VOLUME_BIND \
         -W $WORK_DIR \
         docker://obolibrary/$ODK_IMAGE:$ODK_TAG $TIMECMD "$@"
 else
-    docker run -v $VOLUME_BIND -w $WORK_DIR -e ROBOT_JAVA_ARGS="$ODK_JAVA_OPTS" -e JAVA_OPTS="$ODK_JAVA_OPTS" {% if project.use_external_date is sameas True %}-e TODAY=`date +%Y-%m-%d` {% endif %}--rm -ti obolibrary/$ODK_IMAGE:$ODK_TAG $TIMECMD "$@"
+    BIND_OPTIONS="-v $(echo $VOLUME_BIND | sed 's/,/ -v /')"
```
matentzn, line 56: is this platform independent?

template/src/ontology/run.sh.jinja2
```diff
@@ -37,15 +41,23 @@ fi
...
+if [ -n "$ODK_BINDS" ]; then
+    VOLUME_BIND="$VOLUME_BIND,$ODK_BINDS"
+fi
+
 if [ -n "$USE_SINGULARITY" ]; then
     {% if project.use_external_date is sameas True %}export TODAY=$(date +%Y-%m-%d){% endif %}
-    singularity exec --cleanenv \
+    singularity exec --cleanenv $ODK_SINGULARITY_OPTIONS \
         --env "ROBOT_JAVA_ARGS=$ODK_JAVA_OPTS,JAVA_OPTS=$ODK_JAVA_OPTS" \
         --bind $VOLUME_BIND \
```
matentzn, line 52: Should this receive the same love as the standard docker thing below? (in case ODK_BINDS is set)?

template/src/ontology/run.sh.jinja2
gouttegd: Yes, there’s nothing specific to GNU or bash here. :)

template/src/ontology/run.sh.jinja2
gouttegd: No need. Contrary to Docker’s `-v` option, Singularity’s `--bind` option accepts multiple binding specifications in a single argument (provided they’re comma-separated), so passing `--bind $VOLUME_BIND` is enough. There is no need to split `$VOLUME_BIND` into multiple `--bind` options as is the case for Docker.

matentzn: Thanks

matentzn (approved):

PR #683, opened 2022-09-06
Install GitHub-CLI from the Ubuntu repository.

gouttegd:
We can no longer install GitHub-CLI (`gh`) from the GitHub-maintained package repository because the signing key for this repository has been left to expire.

Instead, we install the version from Ubuntu’s own repository. As of now, Ubuntu provides version 2.4.0 dated from March 23, 2022, which should be recent enough for our needs.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,10 +13,6 @@ ENV ODK_VERSION $ODK_VERSION
 # docker run -v $HOME/.coursier/cache/v1:/tools/.coursier-cache ...
 ENV COURSIER_CACHE "/tools/.coursier-cache"
 
-# Install GH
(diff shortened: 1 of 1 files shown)


gouttegd:
While the CI checks were running, `gh`’s developers have generated and uploaded a new signing key, so we can build `odkfull` again without needing this PR.

I would still vote for merging and use Ubuntu’s `gh` from now on, unless we happen to critically require the very latest version of `gh`. It’s always better to avoid depending on third-party package repositories when possible.

matentzn:
Are you sure ubuntu has a `gh` client installed? If so, shouldnt we update it with `apt get`?

gouttegd:
Yes, `gh` is in Ubuntu’s package repository. That’s why this PR works: even after we remove GitHub’s custom package repository, `apt-get` can still be find `gh` in Ubuntu’s own repository.

matentzn (approved):
Thanks for this explanation!

PR #684, opened 2022-09-06
Run commands inside the container as the current user.

### LINKED ISSUE #682: Files created by the ODK should belong to the user who invoked the ODK
On GNU/Linux, files created by the ODK seeding script belong to the _root_ user.

This is because in most GNU/Linux installations, the Docker daemon itself runs under the identity of the super-user. On MacOS, the problem does not occur because the Docker daemon runs under the same identity as the user who started the Docker Desktop application.

It is possible to configure a GNU/Linux installation to have the Docker daemon run as a non-privileged user, but it’s non-standard and the behaviour of the ODK should not be dependent on the local Docker configuration — it should be up to the ODK to ensure that all the files it creates belong to the user who invoked the seeding script.

A solution to this problem had already been proposed in #225, but it involves the user building its own Docker image. We need a solution that works “out of the box” with the standard image.

gouttegd:
Make sure the command we run inside the ODK container is run under the identity of the user invoking the ODK, instead of `root`.

Running as `root` will cause any file created by the process to be owned by `root`, if the Docker daemon itself runs as a privileged user – which happens to be the case in most GNU/Linux setups.

Running as a normal user means that we can no longer bind the calling user’s own `~/.gitconfig` file to a predictable location inside the container (`/root/.gitconfig`), so instead we extract the informations we need from that file in the wrapper script and pass said informations to `odk.py` through command line arguments. The `odk.py` script will in turn pass the informations to Git using environment variables when initialising the repository.

closes #682

### DIFF
--- a/odk/odk.py
+++ b/odk/odk.py
@@ -799,8 +799,10 @@ def dump_schema():
               """)
 @click.option('-v', '--verbose',      count=True)
 @click.option('-g', '--skipgit',      default=False, is_flag=True)
+@click.option('-n', '--gitname',      default=None)
+@click.option('-e', '--gitemail',     default=None)
 @click.argument('repo', nargs=-1)
-def seed(config, clean, outdir, templatedir, dependencies, title, user, source, verbose, repo, skipgit):
+def seed(config, clean, outdir, templatedir, dependencies, title, user, source, verbose, repo, skipgit, gitname, gitemail):
     """
     Seeds an ontology project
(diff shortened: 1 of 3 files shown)


matentzn (approved):
One limitation of the run.sh change is that we still cant use git commands seamlessly (i.e. when using `gh` to run releases). If you have any idea on how to deal with that, that would be amazing!

PR #687, opened 2022-09-14
Install latest version of NodeJS and update `oaklib`

### LINKED ISSUE #686: Upgrade NPM to latest version
The version the ubuntu store is very outdated. Follow:

https://www.freecodecamp.org/news/how-to-install-node-js-on-ubuntu-and-update-npm-to-the-latest-version/

Here is roughly where the fix should be done: https://github.com/INCATools/ontology-development-kit/blob/master/Dockerfile#L32

hrshdhgd:
- [x] Install `nvm` for easy management of `node` versions
 - [x] Updated dependency to latest version of `oaklib`

  Fixes #686


hrshdhgd:
This works perfectly but `oak lib` throws an error at the end for `make viz`. I have opened a [ticket in `oaklib` ](https://github.com/INCATools/ontology-access-kit/issues/277) about this.

matentzn:
Thank you! I will leave it to @gouttegd to judge this. It looks a bit different than what he proposed in the issue, but I don't really know at all what is right or wrong with these kinds of things.

Dockerfile
```diff
@@ -13,6 +13,22 @@ ENV ODK_VERSION $ODK_VERSION
 # docker run -v $HOME/.coursier/cache/v1:/tools/.coursier-cache ...
 ENV COURSIER_CACHE "/tools/.coursier-cache"
 
+# nvm environment variables
+ENV NVM_DIR /usr/local/nvm
+ENV NODE_VERSION 16.17.0
+
+# install nvm, node and npm
+RUN mkdir $NVM_DIR \
+    && curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.1/install.sh | bash \
```
gouttegd, line 22: Running an arbitrary piece of code just downloaded from the Internet at every build is a no-go. Feel free to do that on your own machine if you don’t care, but not when building an image that will end up on the machine of every ODK user.

Just because everything runs within a container does not mean we are automatically immune to malware. Keep in mind that normal usage of the ODK implies binding a part of the host’s filesystem into the container, so the user’s files are _not_ protected from any malware that could have crept into the ODK.

Supply-chain attacks are trendy, and while we cannot eliminate them altogether (we depend on so many programs and packages that there will always be a risk), let’s at least avoid expanding the attack surface.

Dockerfile
```diff
@@ -73,7 +87,7 @@ RUN swipl -g "pack_install(sparqlprog, [interactive(false)])" -g halt && \
     ln -sf /root/.local/share/swi-prolog/pack/sparqlprog /tools/
 
 # Install obographviz
-RUN npm install obographviz && \
+RUN npm install -g obographviz && \
     ln -s /tools/node_modules/obographviz/bin/og2dot.js /tools/og2dot.js
```
gouttegd, line 91: If you go for a “global” install, then you can remove the `ln` command. It’s no longer needed since `og2dot` is now installed in `/usr/bin`, so already in the PATH.

gouttegd (changes requested):
I am strongly against the method used here to install Node. See inline comment.

Nodesource provides a repository of DEB packages for recent versions of Node, please use it instead. (I can do it if needed.)

Dockerfile
hrshdhgd: Done

Dockerfile
hrshdhgd: Thank you so much of the explanation. That makes total sense.

gouttegd: No! You’re still executing a script straight from the Internet.

What you should do here is to download this script (_without executing it!_), examine it to understand what it is supposed to do (here, it adds a new APT source), and then write your own code in the Dockerfile to do the same thing, without ever needing that script again.

gouttegd: See commit 1fc9a8ac0e2e880029b779af6912b5b285a8381a (that I just added) for the proper way of doing it (there was an additional difficulty due to the need of handling Nodesource’s signing key).

hrshdhgd: Thanks! I'm clearly new to this so I sincerely appreciate this!

gouttegd: No problem! Sorry for being quite a pain in the neck about this issue. :) Your method was working, it was just not following good practices, security-wise.

hrshdhgd: No problem at all! As I mentioned earlier, I truly appreciate it. So do you think this PR is ready to merge?

gouttegd: I manually triggered a run of the CI test suite, just to be sure. (https://github.com/INCATools/ontology-development-kit/actions/runs/3055880201) (it should have been triggered automatically by GitHub after I added the last commit, but for some reasons it has not). Once the run completes (and assuming it passes, obviously), yes, it will be ready to merge.

gouttegd (approved):

Dockerfile
gouttegd: All clear. You can go ahead and merge. :tada:

matentzn: THANK YOU BOTH!

PR #690, opened 2022-09-22
Make it possible to make basis when using components

### LINKED ISSUE #689: Make it possible to turn components into bases
When building application ontologies, see https://github.com/monarch-initiative/phenio/pull/9, we need to be able to turn a source, which is defined as a component in an application ontology, into a base-plus module.

matentzn:
This is only needed by very fringe use cases, but it is important:

Fixes #689.

cc @caufieldjh

### DIFF
--- a/tests/test-components-make-base.yaml
+++ b/tests/test-components-make-base.yaml
@@ -1,0 +1,24 @@
+id: tco
+title: "Test Components Make Base Ontology"
+github_org: INCATools
+repo: ontology-development-kit
+release_artefacts: 
+  - full
+  - simple
+primary_release: full
+export_formats:
+  - owl
+  - json
+import_group:
+  products: 
+    - id: ro
+    - id: bfo
+components:
+  products:
+    - filename: pato.owl
+      base_iris:
+        - http://purl.obolibrary.org/obo/PATO_
+        - http://purl.obolibrary.org/obo/pato#
+      make_base: True
+robot_java_args: '-Xmx8G'
+allow_equivalents: asserted-only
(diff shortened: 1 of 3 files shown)


tests/test-components-make-base.yaml
```diff
@@ -0,0 +1,24 @@
...
+  products: 
+    - id: ro
+    - id: bfo
+components:
+  products:
+    - filename: pato.owl
+      base_iris:
+        - http://purl.obolibrary.org/obo/PATO_
+        - http://purl.obolibrary.org/obo/pato#
+      make_base: True
```
gouttegd, line 22: Maybe I’m missing something, but shouldn’t there be a `source` parameter for the PATO component in this test case? The `make_base` parameter only has an effect in the `component-download-...` rule, which itself is only generated if there is an explicit `source`.

gouttegd (changes requested):

tests/test-components-make-base.yaml
matentzn: Oooops, yeap this was too hastily done. Thanks for catching!

caufieldjh:
Fantastic - thanks @matentzn.
I was also under the impression that components needed a `source` and wouldn't be able to find local owls.

gouttegd (approved):

gouttegd:
As a side note: I am not sure I understand the purpose of a component that is generated from an external `source`. How does such a component differ from an “import”?

gouttegd:
I see, thanks for the clarification. :)

PR #697, opened 2022-09-25
Add checks for RDF/XML products

### LINKED ISSUE #691: Add a validation check that the RDF/XML parses using Rust RDF/XML parser
Not all RDF/XML parsers behave the same way

I have seen cases where Jena is stricter than OWLAPI, and where the Rust parser is stricter still.

E.g pipe symbols in URIs: https://github.com/monarch-initiative/vertebrate-breed-ontology/issues/51

We should ensure that RDF/XML produced is consumable by the union of all production-level parsers.

I believe fastobo wraps the Rust RDF/XML parser but fastobo itself may impose an additional level of strictness

I think rdftab is distributed with ODK so that could also be used to check.

gouttegd:
This PR adds to the standard Makefile a new `check_rdfxml_assets` target, which test each of the RDF/XML-formatted release products for correctness.

The correctness check is performed by the LightRDF Python module (which this PR adds to the `odklite` image). This module wraps the Rust `rio_xml` RDF/XML parser, which is known to be quite strict.

A new ODK configuration is also added: `extra_rdfxml_checks`. If enabled, then each RDF/XML release product is also checked against the RDFLib and Jena parsers (only if the ODK is running from the `odkfull` image, where those parsers are present).

closes #691

### DIFF
--- a/template/src/ontology/Makefile.jinja2
+++ b/template/src/ontology/Makefile.jinja2
@@ -219,12 +219,18 @@ RELEASE_ASSETS = \
   $(REPORT_FILES){% endif %}
 
 .PHONY: all_assets
-all_assets: $(ASSETS)
+all_assets: $(ASSETS) check_rdfxml_assets
 
 .PHONY: show_assets
 show_assets:
 	echo $(ASSETS)
 	du -sh $(ASSETS)
+
+.PHONY: check_rdfxml_assets
+check_rdfxml_assets:
+	@for product in $(MAIN_PRODUCTS) ; do  \
+		check-rdfxml $${product}.owl ; \
+	done
 
 # ----------------------------------------
 # Release Management
--- a/template/src/ontology/Makefile.jinja2
+++ b/template/src/ontology/Makefile.jinja2
@@ -229,7 +229,7 @@ show_assets:
 .PHONY: check_rdfxml_assets
 check_rdfxml_assets:
 	@for product in $(MAIN_PRODUCTS) ; do  \
-		check-rdfxml $${product}.owl ; \
+		check-rdfxml {% if project.extra_rdfxml_checks %}--extra{% endif %} $${product}.owl ; \
 	done
 
 # ----------------------------------------
--- a/docker/builder/Dockerfile
+++ b/docker/builder/Dockerfile
@@ -28,6 +28,24 @@ RUN apt-get update && \
         python3-pip \
         python3-virtualenv
 
+# Install LightRDF.
+# It cannot be installed with pip as all the other Python modules below
+# because:
+#   - the developers do not provide a Linux arm64 wheel;
+#   - they don't even provide a sdist on PyPI, so pip cannot build from
+#     source.
+# So we have to build from source here -- which involves installing
+# yet another Python build system, otherwise it wouldn't be funny.
+RUN python3 -m pip install maturin && \
+    wget -O lightrdf-0.3.1.tar.gz https://github.com/ozekik/lightrdf/archive/refs/tags/v0.3.1.tar.gz && \
+    tar xf lightrdf-0.3.1.tar.gz && \
+    cd lightrdf-0.3.1 && \
+    maturin build -r && \
+    python3 -m pip install --root /staging/lite target/wheels/lightrdf-*.whl && \
+    python3 -m pip install --root /staging/full target/wheels/lightrdf-*.whl && \
+    cd .. && \
+    rm -rf lightrdf-0.3.1 lightrdf-0.3.1.tar.gz
+
 # Build the Python packages.
 # On x86_64, most if not all of these packages should be available as
 # pre-compiled wheels, so we should not need to build them on a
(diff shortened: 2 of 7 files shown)


template/src/ontology/Makefile.jinja2
```diff
@@ -219,13 +219,19 @@ RELEASE_ASSETS = \
   $(REPORT_FILES){% endif %}
 
 .PHONY: all_assets
-all_assets: $(ASSETS)
+all_assets: $(ASSETS) check_rdfxml_assets
```
matentzn, line 222: This is missing the jinja conditionals?

template/src/ontology/Makefile.jinja2
```diff
@@ -219,13 +219,19 @@ RELEASE_ASSETS = \
...
 
 .PHONY: show_assets
 show_assets:
 	echo $(ASSETS)
 	du -sh $(ASSETS)
 
+.PHONY: check_rdfxml_assets
+check_rdfxml_assets:
+	@for product in $(MAIN_PRODUCTS) ; do  \
```
matentzn, line 231: I wonder if this goal should dependent on $(MAIN_PRODUCTS). Remember that `prepare_release` moves those elsewhere.. alternatively, maybe in line with profile checking etc, we could use `check_rdfxml_%: %`?

matentzn (commented):
Awesome move looks great. Will look more later after ICBO presentations!

template/src/ontology/Makefile.jinja2
gouttegd: The check rule itself is intended to be unconditional – executed as part of the standard workflow with no opt-out option.

But by default, only _one_ check using the LightRDF parser is enabled. What the `extra_rdfxml_checks` option does is instructing the `check-rdfxml` script to also perform two more checks using the RDFLib and Jena parsers.

The rationale for this behaviour is that:

1) LightRDF is both quite fast and strict, so it doesn’t cost a lot to always run a check with it and it should ensure that any produced RDF/XML file is actually correct. LightRDF is basically high gain, low cost.

2) By contrast, the other parsers are slower (RDFLib in particular is _very_ slow) and don’t appear to be stricter than LightRDF (from some tests I did over the weekend, RDFLib seems to be just as strict as LightRDF, and Jena is laxer), so they don’t offer much. The option to use them is intended for people who want extra assurance that their RDF/XML files can really be consumed by most parsers out there, in the off-chance that they could catch an corner-case error that is somehow missed by LightRDF.

template/src/ontology/Makefile.jinja2
gouttegd: For now, the rule is a pre-requisite to `all_assets`, which means it is automatically called _after_ the assets are generated, and _before_ they are moved to the top-level directory by `prepare_release`. This seemed to me like a good place, but I am open to consider other options.

gouttegd:
We can dispense with the last commit of this PR (5750d94): we no longer need to install `lightrdf` from source, it is now available as a binary wheel also for Linux/arm64, thanks to its developer. :tada:

template/src/ontology/Makefile.jinja2
matentzn: Unfortunately I think we need a conditional - this could introduce breaking changes and by design, we should never introduced breaking changes in an ODK update unless we provide a way to switch it off. . I would prefer the whole rdfxml checking pipeline to only be activated if `ensure_valid_rdfxml: TRUE`. This is consistent with `ensure_valid_owl2dl`.

template/src/ontology/Makefile.jinja2
matentzn: All other PHONY goals can be run independently of any other goal - I would kind of like this to be generally the case. If my RDFXML is broken, and I am working on a fix in my edit file, I want to quickly run `sh run.sh  make check_rdfxml_assets` to ensure its fixed..

template/src/ontology/Makefile.jinja2
gouttegd: OK, makes sense. I would favour the lightrdf check being enabled by default, though. I am concerned almost nobody would turn it on otherwise, which would defeat the whole point of having this check in the first place.

If people find out they have invalid RDF/XML output files after upgrading their ODK instance, and if they decide they don’t care, all they will have to do to restore a “working” pipeline is to change one line in their config file.

matentzn: 100% switch on by default!

template/src/ontology/Makefile.jinja2
gouttegd: So what about having:

* `check_rdfxml_%: %` (as you initially suggested), allowing to do a quick check on a arbitrary file (possibly not even a release product, e.g. if your -edit file is already a RDF/XML file and you want to check it without having to run the pipeline to produce the release files, you could call `sh run.sh make check_rdfxml_myont-edit.owl`), and
* `check_rdfxml_assets` dependent on `check_rdfxml_X.owl`, with _X_ being each RDF/XML product file

?

matentzn: Ok, sounds good!

gouttegd:
Superseded by #702, which removes the no-longer-needed commit for source installation of LightRDF and implements the changes discussed above.

PR #702, opened 2022-09-29
Add rdfxml validation

### LINKED ISSUE #691: Add a validation check that the RDF/XML parses using Rust RDF/XML parser

gouttegd:
This PR adds to the standard Makefile a new `check_rdfxml_assets` target, which test each of the RDF/XML-formatted release products for correctness.

The correctness check is performed by the LightRDF Python module (which this PR adds to the odklite image). This module wraps the Rust rio_xml RDF/XML parser, which is known to be quite strict.

Two new ODK configuration options are added:

* `ensure_valid_rdfxml`: If enabled (this is the default), the `check_rdfxml_assets` target is invoked as part of the standard release workflow.
* `extra_rdfxml_checks`: If enabled (_not_ by default), then each RDF/XML release product is also checked against the RDFLib and Jena parsers (only if the ODK is running from the `odkfull` image, where those parsers are present; if running from `odklite`, that option has no effects).

closes #691
supersedes #697

### DIFF
--- a/template/src/ontology/Makefile.jinja2
+++ b/template/src/ontology/Makefile.jinja2
@@ -219,12 +219,18 @@ RELEASE_ASSETS = \
   $(REPORT_FILES){% endif %}
 
 .PHONY: all_assets
-all_assets: $(ASSETS)
+all_assets: $(ASSETS) {% if project.ensure_valid_rdfxml %}check_rdfxml_assets{% endif %}
 
 .PHONY: show_assets
 show_assets:
 	echo $(ASSETS)
 	du -sh $(ASSETS)
+
+check_rdfxml_%: %
+	@check-rdfxml {% if project.extra_rdfxml_checks %}--extra{% endif %} $<
+
+.PHONY: check_rdfxml_assets
+check_rdfxml_assets: $(foreach product,$(MAIN_PRODUCTS),check_rdfxml_$(product).owl)
 
 # ----------------------------------------
 # Release Management
--- a/docker/odklite/Dockerfile
+++ b/docker/odklite/Dockerfile
@@ -70,9 +70,12 @@ RUN wget -nv https://github.com/lihaoyi/Ammonite/releases/download/$AMMONITE_VER
     chmod 755 /tools/amm && \
     /tools/amm /dev/null
 
+# Install RDF/XML validation script
+COPY scripts/check-rdfxml.sh /tools/check-rdfxml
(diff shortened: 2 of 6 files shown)


template/src/ontology/Makefile.jinja2
```diff
@@ -219,13 +219,19 @@ RELEASE_ASSETS = \
...
 .PHONY: all_assets
-all_assets: $(ASSETS)
+all_assets: $(ASSETS) {% if project.ensure_valid_rdfxml %}check_rdfxml_assets{% endif %}
 
 .PHONY: show_assets
 show_assets:
 	echo $(ASSETS)
 	du -sh $(ASSETS)
 
+check_rdfxml_%: %
```
matentzn, line 229: should probably also be declared as .PHONY?

matentzn (approved):
I didnt review `scripts/check-rdfxml.sh`

All looks great.

template/src/ontology/Makefile.jinja2
```diff
@@ -219,13 +219,19 @@ RELEASE_ASSETS = \
...
+	@check-rdfxml {% if project.extra_rdfxml_checks %}--extra{% endif %} $<
```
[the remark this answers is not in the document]
gouttegd: Disable echoing the command before running it. There’s no need here to have Make print a `check-rdfxml file.owl` on the terminal, since the `check-rdfxml` script itself will print a message indicating what it does.

template/src/ontology/Makefile.jinja2
gouttegd: I have seen nothing in Make’s documentation suggesting this would work for pattern rules (and indeed it does not seem to work).

The only way to make it work would be to declare as `.PHONY` all the possible expansions of the rule, based on the known `MAIN_PRODUCTS`. Too cumbersome for the little benefits this would bring, in my opinion.

PR #708, opened 2022-10-10
Add QC to `term_tracker_item` type

### LINKED ISSUE #610: Consider replacing some of the SPARQL QC checks with OAK validate
OAK has a new but very experimental validation interface:

https://incatools.github.io/ontology-access-kit/interfaces/validator.html

We should start exploring that for QC rather than adding sparql. Good first issue:

#587 

@anitacaron would this be an issue you'd be interested in? For starters, the only thing we need is:

1. An example of a working OAK command (CLI) which takes in an ontology and validates it.
2. An implementation of #587 extending the ontology metadata schema provided by OAK (we will maintain our own here for now).
3. A make goal in Makefile.jinja2 that runs the validation on `$(ONT).owl`.

anitacaron:


anitacaron:
Ops, fixing the wrong issue.

PR #710, opened 2022-10-10
Add replaced by to QC as default

### LINKED ISSUE #703: Add replaced by functional test to QC (add to default)
```
PREFIX IAO: <http://purl.obolibrary.org/obo/IAO_>
PREFIX oboInOwl: <http://www.geneontology.org/formats/oboInOwl#>
PREFIX owl: <http://www.w3.org/2002/07/owl#>
PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#>

SELECT DISTINCT ?entity ?property ?value WHERE {
  VALUES ?property {
    IAO:0100001
  }
  ?entity ?property ?value1 .
  ?entity ?property ?value2 .
  FILTER(?value1!=?value2)
  BIND(CONCAT(str(?value1), CONCAT("|", str(?value2))) as ?value)
}
```

https://api.triplydb.com/s/yhUVUbzp_

anitacaron:
Fixes #703

### DIFF
--- a/odk/odk.py
+++ b/odk/odk.py
@@ -327,7 +327,7 @@ class ReportConfig(JsonSchemaMixin):
     release_reports : bool = False
     """ If true, release reports are added as assets to the release (top level directory, reports directory)"""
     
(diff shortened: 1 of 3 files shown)


matentzn (approved):
Excellent, thank you very much!

PR #711, opened 2022-10-10
Cleaning up the top-level directory

### LINKED ISSUE #707: Cleaning up the top-level directory
The top-level directory of the ODK repository contains quite a few files that are either obsolete or of dubious usefulness.

* `CONTRIBUTING.md`: Not updated since 2017. Only contains a single instruction about how to run the test suite, which is not even accurate anymore (the test suite will do more than create just one new ontology repo). Suggestion: replace with `README-developers.md`, which is both more complete and more up-to-date. Keep the name `CONTRIBUTING.md`, though, as it is a pretty standard name for a file containing instructions for contributors.

* `INSTALL.sh`: Dates from the time where the ODK was still a collection of tools to be installed, rather than a Docker image. Now completely obsolete (the URLs it tries to download tools from are no longer working), and very misleading (people cloning the ODK repository will be inclined to think they need to run that script in order to use the ODK). Suggestion: nothing to save here, just delete.

* `README-developers.md`: Rename to `CONTRIBUTING.md` (see above).

* `env.sh`: Made obsolete by `update-constraints.sh`. Suggestion: delete.

* `odk2.sh`: A variant of `odk`, with the sole difference that it uses the `obotools/odkfull` image. AFAIR, that image was only used to test the M1 build in spring 2021. Suggestion: delete. If it is still deemed useful to be able to use another image than `obolibrary/odkfull`, then update the `odk.sh` script to allow to override the default image, similarly to what we do in the `run.sh` script.

* `seed-via-docker-131.sh`: Was a temporary workaround following changes made on `seed-via-docker.sh` in the master branch. Suggestion: delete. Better workaround (as discussed on Slack) is to download the script from a versioned URL if needed.

* `upgrade-via-docker.sh`: Is anyone using this? It seems to me that it is redundant with the `update_repo.sh` script that is installed in any ODK instance.

* `upgrade_repos.sh` and `upgrade_repos_test.sh`? Is anyone beyond @matentzn using those scripts? They are highly specific to his system. I Don’t think the ODK repo is the right place for scripts like that… Suggestions: either
  * kick them out of the repo completely (sorry Nico!);
  * keep them, but elsewhere than in the top-level directory (e.g. move them to a new directory `helper-scripts` or similar);
  * if “bulk upgrading of multiple ODKs at once” is deemed useful for other people than just Nico, then refactor into a proper script that can really be used other people and does not hardcode paths.

gouttegd:
This PR removes a handful of obsolete files from the top-level directory of the repository.

It also moves the `README-developers.md` file to `CONTRIBUTING.md`, replacing the (obsolete) contents of the latter.

closes #707


gouttegd:
Note 1: CI check intentionally cancelled as this PR only touches auxiliary files.

Note 2: Waiting for @matentzn’s input (not before next week) regarding the usefulness (or lack thereof) of the following files:
- `seed-via-docker-toolbox-dev.bat`
- `upgrade-via-docker.sh`
- `upgrade_repos.sh`
- `upgrade_repos_test.sh`

matentzn (approved):
@gouttegd great I like the direction. The CONTRIBUTING.md file should be present in all repos (this one could do with some more work, but better the way you do it now than it was), so we can adhere to the "social standards" of GitHub a bit better: https://github.com/INCATools/ontology-development-kit/community

matentzn:
> seed-via-docker-toolbox-dev.bat

Can be deleted, never got this to work.

> upgrade-via-docker.sh

Can be removed, replaced by "update_repo"

> upgrade_repos.sh

Can be removed, replaced by "update_repo" - a service for updating ODKs would still be interesting, but ok for now.

> upgrade_repos_test.sh

Can be removed

PR #714, opened 2022-10-10
Add path constraints to the QC workflow.

### LINKED ISSUE #712: Add path constraints to ODK QC action
See https://github.com/obophenotype/bio-attribute-ontology/blob/master/.github/workflows/dosdp.yml#L6 as an example;

QC should only run when QC relevant files are edited!

gouttegd:
Inhibit the `build-and-test` GitHub Action workflow when a PR does not modify any testable file (documentation files or helper scripts that are not part of an ODK workflow).

We list the paths that should inhibit the CI check rather than the paths that should trigger a check, so that the default action is still to trigger a check (better to trigger a check on a file we forgot to exclude than to inhibit a check on a file we forgot to include).

closes #712

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -6,6 +6,21 @@ on:
       - master
     types:
       - opened
+    paths-ignore:
+      - '**.md'
+      - 'CITATION.cff'
+      - 'LICENSE'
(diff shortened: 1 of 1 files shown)


matentzn (approved):
Very minimal, zero concerns.

PR #720, opened 2022-10-18
Updating actions/checkout to v3

shawntanzk:
Not 100% sure if this is needed, but github has mentioned to update node12 to node16 (https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/) and checkout v3 (https://github.com/actions/checkout) in whats new says run node16 by default, so figured its a good idea to do this update.

Not sure if we need testing etc. and not sure how to do that

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -28,7 +28,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Check out the repository
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
       - name: Create dummy Git user name
         run: git config --global user.name Alice
       - name: Create dummy Git user email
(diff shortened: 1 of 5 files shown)


gouttegd:
Wait, the checkout action is implemented in Javascript? :scream: Oh my…

> Not sure if we need testing etc. and not sure how to do that

Nothing special to do. When you change a GitHub Action workflow that is supposed to run on a PR, it’s the updated version of the workflow that is applied on that PR (as if the PR had already been merged), not the one in master. So the CI check that is currently under way _is_ using `checkout-v3` — and the checking out step seems to have been done normally. :)

shawntanzk:
testing change on jinja template files on CL seem to work nicely too - https://github.com/obophenotype/cell-ontology/pull/1716

gouttegd (approved):

PR #721, opened 2022-10-19
fix wrong target-rule for recreating components in comment/help section

StroemPhi:
This PR fixes the wrong target-rule for recreating components in the comment/help section at the end of the Makefile, which previously stated `refresh-components` whereas the actual target-rule is called `recreate-components` (see [line 458](https://github.com/INCATools/ontology-development-kit/blob/master/template/src/ontology/Makefile.jinja2#L458))


anitacaron (approved):

anitacaron:
Thank you for this PR, @StroemPhi!

PR #724, opened 2022-10-26
List all custom sparql checks available

anitacaron:


matentzn (approved):
Thank you! We just need to remember a bit keeping that list up to date when adding new checks..

Issue #587, opened 2022-04-25
Please incorporate checks for the dateTime property 


bvarner-ebi:
Context: https://github.com/obophenotype/uberon/pull/2379#discussion_r845999223

To ensure users enter **date property** values using 'dcterms:date', will you please incorporate a quality control step that confirms these are entered as 'dcterms:date' and not a URL?

For example:

WRONG: property_value: **http://purl.org/dc/elements/1.1/date** 2022-04-04T11:06:09Z xsd:dateTime
CORRECT: property_value: **dcterms:date ** 2022-04-04T11:06:09Z xsd:date

shawntanzk:
I think there was an agreement in uberon call that we should in general use dcterms instead of dce - wonder if its worth having a blanket dcterms instead of dce check? I remember @anitacaron wrote a SPARQL query check for this, we can probably repurpose?

matentzn:
We will do that, but we are proposing a motion to use `dcterms:` with `xsd:date` (`2022-04-18`) rather than `dateTime`.

bvarner-ebi:
To confirm, should it be the following?

property_value: dcterm: 2022-04-04T11:06:09Z xsd:date

matentzn:
@bvarner-ebi you can do that for now, but we do not know the answer yet. There is another proposal in now that wants to use:

```
property_value: dcterms:modified 2022-04-04T11:06:09Z xsd:date
```

and

```
property_value: dcterms:created 2022-04-04T11:06:09Z xsd:date
```

instead.

bvarner-ebi:
OK, thank you. Is there a reference or could you provide an explanation on the impact these changes will make?

shawntanzk:
> an explanation on the impact these changes will make?

Having both shows when the term was created, and when it was last modified. I think the logic is many times you want to find out when the term was last modified so you can trace changes back to a PR for example, which can be useful for fixing problems.

matentzn:
do you mean positive or negative impact?

positive is
- using meaningful standard vocabulary
- using less error prone date strings rather than  complex dateTime

negative is
- people need to change their habits (costly)
- validation tools need to be introduced

I don't think date is a really risky one.. its probably the most straight forward of our current data standardisation issues..

@shawntanzk we really want to link issues to terms as well moving forward - this is super useful to just do it.. for everything. even multiple ones!

bvarner-ebi:
> do you mean positive or negative impact?

[...]

PR #727, opened 2022-11-01
Add optional date check

shawntanzk:
This PR adds a sparql query that can be added as an optional QC that only allows dates in a certain format (xsd:date, XXXX-XX-XX) 
Fixes https://github.com/INCATools/ontology-development-kit/issues/587
Proof of Concept: https://github.com/obophenotype/uberon/pull/2706


anitacaron:
Could you also add the name of the violation at the end of [this line here](https://github.com/INCATools/ontology-development-kit/blob/master/odk/odk.py#L332), please?

(Could you also add the name of a QC I did before and didn't add to the list? The name is `term_tracker_item type` - reference PR https://github.com/INCATools/ontology-development-kit/pull/709)

anitacaron:
Yes, you're right. 😃

PR #734, opened 2022-11-13
Add make command sequence for a local cronjob for dev image

matentzn:
This is basically the sequence of commands I need to run every time I want to publish a dev image. Now, I want to simply set up a cronjob on my server that does this for me.

The idea is to simply run:

```
make dev-cronjob
```

every Sunday.. I am not sure whether this is a great idea, or whether this way of doing it is right.. What do you think @gouttegd?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -174,3 +174,9 @@ constraints.txt: requirements.txt.full
 
 clean-tests:
 	rm -rf target/*
+
+dev-cronjob:
+	git pull
+	docker buildx rm multiarch
+	docker buildx create --name multiarch --driver docker-container --use
+	$(MAKE) tests publish-multiarch-dev
--- a/Makefile
+++ b/Makefile
@@ -175,7 +175,7 @@ constraints.txt: requirements.txt.full
 clean-tests:
 	rm -rf target/*
 
-dev-cronjob:
+dev-test-publish:
 	git pull
 	docker buildx rm multiarch
 	docker buildx create --name multiarch --driver docker-container --use


gouttegd:
I don’t understand why you are destroying your multiarch builder before immediately creating it again.

The multiarch builder only needs to be created once, then it can be used for all subsequent builds. So the only commands that are really needed in a cron job to build and publish the latest `dev` image are `git pull` and `make tests publish-multiarch-dev` – something short enough to be written directly in a crontab, without a need for a specific rule in the Makefile.

matentzn:
Good that you question it. If I don't destroy the multiarch, some cache is never being updated even when using no-cache, which can lead to certain Ubuntu packages not being installed correctly when the images are build. It basically just fails with "cant find suitable version for package xyz". You helped me last time to device this last resort issue of destroying the multiarch - that always works.

gouttegd:
> You helped me last time to device this last resort issue of destroying the multiarch

Did I? Oh my, I have no recollection of it. ^^'

I wonder if `docker buildx prune`, which is supposed to [“remove the build cache”](https://github.com/docker/buildx/blob/master/docs/reference/buildx_prune.md), would be a better long-term solution. Something worth to try, but in the meantime I have no objection to the approach used here.

Minor nitpick (feel free to ignore): I would change the name of the target to something like, maybe, `dev-auto` – basically not mentioning a cron job. The fact that the target is intended to be run as part of a cron job is irrelevant.

gouttegd (approved):

matentzn:
This was the whole problem, that `docker buildx prune` did not work!

Makefile
```diff
@@ -174,3 +174,9 @@ constraints.txt: requirements.txt.full
 
 clean-tests:
 	rm -rf target/*
+
+dev-cronjob:
```
matentzn, line 178:
```suggestion
dev-test-publish:
```

gouttegd:
> This was the whole problem, that docker buildx prune did not work!

OK, so I really forgot the whole discussion! I must be getting old… :D

PR #736, opened 2022-11-14
Pass a GitHub token to the container

### LINKED ISSUE #599: gh should get an login step with token
```
gh auth login --with-token < mytoken.txt
```

See 

https://cli.github.com/manual/gh_auth_login

Else we cant use the shiny new deploy release step

gouttegd:
This PR updates the `run.sh` file to try to obtain a GitHub token (either from the environment, or from a few selected locations) and pass it as a `GH_TOKEN` environment variable inside the container.

This allows to use GitHub clients such as `gh` from within the container.

closes #599

### DIFF
--- a/docker/odklite/Dockerfile
+++ b/docker/odklite/Dockerfile
@@ -73,6 +73,9 @@ RUN wget -nv https://github.com/lihaoyi/Ammonite/releases/download/$AMMONITE_VER
 # Install RDF/XML validation script
 COPY scripts/check-rdfxml.sh /tools/check-rdfxml
 
+# Force Git to accept working on a repository owned by someone else
+RUN echo "[safe]\n    directory = /work" >> /etc/gitconfig
+
 # Install the ODK itself.
 COPY odk/make-release-assets.py /tools
 COPY odk/odk.py /tools
--- a/template/.gitignore.jinja2
+++ b/template/.gitignore.jinja2
@@ -7,7 +7,7 @@ bin/
 *.tmp.owl
 *.tmp.json
 
-.github/token
+.github/token.txt
 
 src/ontology/mirror
 src/ontology/mirror/*
--- a/template/.gitignore.jinja2
+++ b/template/.gitignore.jinja2
@@ -7,6 +7,8 @@ bin/
 *.tmp.owl
 *.tmp.json
 
+.github/token
+
 src/ontology/mirror
 src/ontology/mirror/*
 src/ontology/reports/*
--- a/template/.gitignore.jinja2
+++ b/template/.gitignore.jinja2
@@ -32,6 +32,7 @@ src/ontology/tmp/*
 !src/ontology/tmp/README.md
 
 src/ontology/run.sh.conf
+src/ontology/run.sh.env
 
 src/ontology/imports/*_terms_combined.txt
 
--- a/template/src/ontology/run.sh.jinja2
+++ b/template/src/ontology/run.sh.jinja2
@@ -21,8 +21,8 @@ fi
 # Look for a GitHub token
 if [ -n "$GH_TOKEN" ]; then
     :
-elif [ -f ../../.github/token ]; then
-    GH_TOKEN=$(cat ../../.github/token)
+elif [ -f ../../.github/token.txt ]; then
+    GH_TOKEN=$(cat ../../.github/token.txt)
 elif [ -f $XDG_CONFIG_HOME/ontology-development-kit/github/token ]; then
     GH_TOKEN=$(cat $XDG_CONFIG_HOME/ontology-development-kit/github/token)
 elif [ -f "$HOME/Library/Application Support/ontology-development-kit/github/token" ]; then
--- a/template/src/ontology/run.sh.jinja2
+++ b/template/src/ontology/run.sh.jinja2
@@ -18,6 +18,17 @@ if [ -f run.sh.conf ]; then
     source run.sh.conf
 fi
 
+# Look for a GitHub token
+if [ -n "$GH_TOKEN" ]; then
+    :
+elif [ -f ../../.github/token ]; then
+    GH_TOKEN=$(cat ../../.github/token)
+elif [ -f $XDG_CONFIG_HOME/ontology-development-kit/github/token ]; then
+    GH_TOKEN=$(cat $XDG_CONFIG_HOME/ontology-development-kit/github/token)
+elif [ -f "$HOME/Library/Application Support/ontology-development-kit/github/token" ]; then
+    GH_TOKEN=$(cat "$HOME/Library/Application Support/ontology-development-kit/github/token")
+fi
+
 ODK_IMAGE=${ODK_IMAGE:-odkfull}
 TAG_IN_IMAGE=$(echo $ODK_IMAGE | awk -F':' '{ print $2 }')
 if [ -n "$TAG_IN_IMAGE" ]; then
@@ -49,6 +60,7 @@ fi
 cat <<EOF > run.sh.env
 ROBOT_JAVA_ARGS=$ODK_JAVA_OPTS
 JAVA_OPTS=$ODK_JAVA_OPTS
+GH_TOKEN=$GH_TOKEN
 EOF
 
 if [ -n "$USE_SINGULARITY" ]; then
--- a/template/src/ontology/run.sh.jinja2
+++ b/template/src/ontology/run.sh.jinja2
@@ -45,20 +45,28 @@ if [ -n "$ODK_BINDS" ]; then
     VOLUME_BIND="$VOLUME_BIND,$ODK_BINDS"
 fi
 
+# Gather environment variables into a single file
+cat <<EOF > run.sh.env
+ROBOT_JAVA_ARGS=$ODK_JAVA_OPTS
+JAVA_OPTS=$ODK_JAVA_OPTS
+EOF
+
 if [ -n "$USE_SINGULARITY" ]; then
     {% if project.use_external_date is sameas True %}export TODAY=$(date +%Y-%m-%d){% endif %}
     singularity exec --cleanenv $ODK_SINGULARITY_OPTIONS \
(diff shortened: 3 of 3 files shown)


docker/odklite/Dockerfile
```diff
@@ -73,6 +73,9 @@ RUN wget -nv https://github.com/lihaoyi/Ammonite/releases/download/$AMMONITE_VER
 # Install RDF/XML validation script
 COPY scripts/check-rdfxml.sh /tools/check-rdfxml
 
+# Force Git to accept working on a repository owned by someone else
+RUN echo "[safe]\n    directory = /work" >> /etc/gitconfig
```
matentzn, line 77: What exactly does this do? Never seen this before!

template/.gitignore.jinja2
```diff
@@ -7,6 +7,8 @@ bin/
 *.tmp.owl
 *.tmp.json
 
+.github/token
```
matentzn, line 10: you settled on this location as default for the token? I am wondering, since `.github` is actually checked in git, whether the name of the token file should start with period, i.e. `.token`, to indicate that we kinda don't want it checked in? Maybe its stupid because it will make using the infinitely harder for Windows users, who cant create files starting with a period without jumping through major hoops

matentzn: (what I am saying: you are probably right in calling it `token`

template/.gitignore.jinja2
```diff
@@ -32,6 +34,7 @@ src/ontology/tmp/*
 !src/ontology/tmp/README.md
 
 src/ontology/run.sh.conf
+src/ontology/run.sh.env
```
matentzn, line 37: Is this to circumvent the mess I made by starting to check in `conf`?

template/src/ontology/run.sh.jinja2
```diff
@@ -45,21 +56,30 @@ if [ -n "$ODK_BINDS" ]; then
...
-        --env "ROBOT_JAVA_ARGS=$ODK_JAVA_OPTS,JAVA_OPTS=$ODK_JAVA_OPTS" \
+        --env-file run.sh.env \
         --bind $VOLUME_BIND \
         -W $WORK_DIR \
         docker://obolibrary/$ODK_IMAGE:$ODK_TAG $TIMECMD "$@"
 else
     BIND_OPTIONS="-v $(echo $VOLUME_BIND | sed 's/,/ -v /')"
     docker run -u $(id -u):$(id -g) $ODK_DOCKER_OPTIONS $BIND_OPTIONS -w $WORK_DIR \
-        -e ROBOT_JAVA_ARGS="$ODK_JAVA_OPTS" -e JAVA_OPTS="$ODK_JAVA_OPTS" \
+        --env-file run.sh.env \
```
matentzn, line 76: You do this because it is a security to risk to pass the token with -e right?

template/src/ontology/run.sh.jinja2
```diff
@@ -45,21 +56,30 @@ if [ -n "$ODK_BINDS" ]; then
...
         {% if project.use_external_date is sameas True %}-e TODAY=`date +%Y-%m-%d` {% endif %} \
         --rm -ti obolibrary/$ODK_IMAGE:$ODK_TAG $TIMECMD "$@"
 fi
 
+rm run.sh.env
```
matentzn, line 81: What happens if you have two shells open:

1. one is running a release (long running, needs the token)
2. The other, different shell, is just checking something real quick `sh run.sh robot -i...`.

Will 2 not delete the file before 1 can use it and thus make it fail?

template/src/ontology/run.sh.jinja2
```diff
@@ -18,6 +18,17 @@ if [ -f run.sh.conf ]; then
     source run.sh.conf
 fi
 
+# Look for a GitHub token
+if [ -n "$GH_TOKEN" ]; then
+    :
+elif [ -f ../../.github/token ]; then
+    GH_TOKEN=$(cat ../../.github/token)
```
matentzn, line 25: What if our naive user does not _want_ to share their token with ODK (a potential security risk, from their perspective)?

template/.gitignore.jinja2
gouttegd: Yeah, since we already have a `.github` directory for, well, GitHub-related stuff, I didn’t see any point in imagining yet another hidden file or directory.

As for calling it `.github/.token`: I wouldn’t see the point either, the `.github` directory is already hidden. And the file is listed in `.gitignore` anyway, so users can only commit it if they force it to be committed – and if they do that, well, they’re free to do stupid things if they absolutely want to.

template/.gitignore.jinja2
gouttegd: No, it’s to use as an argument to Docker’s `--env-file` option. The `run.sh.conf` file may contain variables that are only needed in the `run.sh` script itself and don’t need to be passed to the container (e.g., the `ODK_IMAGE` variable).

But yeah, we also need to take into account that _some people_ commit their `run.sh.conf` files… ^^'

template/src/ontology/run.sh.jinja2
gouttegd: At least it’s a _theoretical_ security risk.

To be honest, I think it’s borderline security theatre. If an attacker has enough access to your machine to be able to read a process’ command line, you already have a serious security issue and your GitHub token (as well as any other credential you may have stored on your machine) is already at risk – making sure it does not appear in a command line will not save you.

I did it that way because I feel some people might be reluctant to use the feature if they see their GitHub token being used on a command line. But you wouldn’t have to push me too hard to convince me that it is not worth it.

template/src/ontology/run.sh.jinja2
gouttegd: As long as the second process happens _after_ the first process has reached the `docker run` command, that wouldn’t be a problem. At this point the container of the first process is already running with the variable in its environment.

There _is_ a very slight possibility for a race condition if you start the second process _immediately_ after the first, and the second process runs so fast that it terminates (and thus deletes the `run.sh.env` file) before the first process even has the time to reach the `docker run` command.

I consider that highly unlikely, but if we want to avoid even that possibility, we can give the `.env` a unique, process-dependent filename (e.g. `run.sh.env.$$`, where `$$` is the PID of the current process).

Or, we forget about the `.env` file altogether and go back to passing environment variables on the command line, if we decide that preventing the GitHub token from appearing in a command line is not worth (see previous comment).

docker/odklite/Dockerfile
matentzn: Any security risks you can think of?

template/.gitignore.jinja2
matentzn: Ok, makes sense! To make this easier for windows, would we consider `token.txt`?

template/src/ontology/run.sh.jinja2
gouttegd: You mean in the case they do have a `GH_TOKEN` variable in their environment, but don’t want to have it passed to the container?

Easy: they can just add

```
GH_TOKEN=invalid
```

to their `run.sh.conf` file. That file is read _before_ we lookup for `GH_TOKEN`, so the bogus definition in `run.sh.conf` will override what they have in their environment. (They can use whatever they want instead of “invalid”; all that matters is that it should not be an empty string.)

template/.gitignore.jinja2
matentzn: Haha thanks!

template/src/ontology/run.sh.jinja2
matentzn: No I think I agree with it and it sets us up for future similar situations.

template/src/ontology/run.sh.jinja2
matentzn: Ah ok, no no, I forgot that docker would evaluate the `run.sh.env` when the container is started, not like "binding a file into the container". Sounds good. Maybe add `rm -f` to avoid the minor risk of a race condition?

docker/odklite/Dockerfile
gouttegd: None. The repository _does_ belong to the calling user anyway, only the top-level directory to which it is bound does not.

matentzn: Thank you!

template/.gitignore.jinja2
gouttegd: Sure. Note however that this won’t be of any use for now, since I have not updated the `run.bat` script for Windows (and I have no plans of doing it, since I cannot test it), so the feature is for Unix-users only for now.

I am sorry for Windows users, but some of them (at least one!) will have to step up and contribute if they want the same level of features than Unix-like users…

template/src/ontology/run.sh.jinja2
matentzn: But it would require the user to know that ODK going to read this variable and do insecure stuff with it. I was more concerned about the 3rd and 4th `elif`, checking default directories for tokens. This seems like something a user would want more control over..

gouttegd: Well, they do have control over it: the only way their GitHub token is going to end up in a file called `$XDG_CONFIG_HOME/ontology-development-kit/github/token` is if they deliberately create such a file and write their token into it – it’s not going to happen behind their back. Presumably if they do put their token in such a file, it’s because they do want to have it used by the ODK, because that would be the only purpose of that file.

I would agree with your concerns if we tried to read the token from a non-ODK-specific location (e.g. `$XDG_CONFIG_HOME/github/token`) – users would not legitimately expect that. But they can’t possibly be surprised that the ODK is using the information contained in a file specifically intended for the ODK.

matentzn: Oh yes,  you are of course right. Ok I think I am convinced this is ok then!

matentzn (approved):
This is excellent, thank you.

PR #739, opened 2022-11-16
Add conditional to copy only if `dosdp-patterns/external.txt` exists

### LINKED ISSUE #574: `update_patterns` fails hard when called "cold"
@shawntanzk recently observed that `make update_patterns` fails hard when there `src/patterns/dosdp-patterns/external.txt` is empty:

```
cp: cannot stat 'tmp/dosdp/*.yaml': No such file or directory
```

We should probably add a -f or a conditional so pipelines do not fail hard in this case.

anitacaron:
Fixes #574

### DIFF
--- a/template/src/ontology/Makefile.jinja2
+++ b/template/src/ontology/Makefile.jinja2
@@ -645,7 +645,7 @@ pattern_schema_checks dosdp_validation: $(TMPDIR)/pattern_schema_checks
 
 .PHONY: update_patterns
 update_patterns: download_patterns
-	if [ -d $(TMPDIR)/dosdp ]; then if [ "$$(ls $(TMPDIR)/dosdp)" ]; then cp -r $(TMPDIR)/dosdp/*.yaml $(PATTERNDIR)/dosdp-patterns; fi; fi
+	if [ -n "$$(find $(TMPDIR) -type f -path '$(TMPDIR)/dosdp/*.yaml')" ]; then cp -r $(TMPDIR)/dosdp/*.yaml $(PATTERNDIR)/dosdp-patterns; fi
 
 # This command is a workaround for the absence of -N and -i in wget of alpine (the one ODK depend on now).
 # It downloads all patterns specified in external.txt
--- a/template/src/ontology/Makefile.jinja2
+++ b/template/src/ontology/Makefile.jinja2
@@ -634,7 +634,7 @@ pattern_schema_checks dosdp_validation: $(TMPDIR)/pattern_schema_checks
 
 .PHONY: update_patterns
 update_patterns: download_patterns
-	if [ -d $(TMPDIR)/dosdp ]; then if [ "$(ls $(TMPDIR)/dosdp)" ]; then cp -r $(TMPDIR)/dosdp/*.yaml $(PATTERNDIR)/dosdp-patterns; fi; fi
+	if [ -d $(TMPDIR)/dosdp ]; then if [ "$$(ls $(TMPDIR)/dosdp)" ]; then cp -r $(TMPDIR)/dosdp/*.yaml $(PATTERNDIR)/dosdp-patterns; fi; fi
 
 # This command is a workaround for the absence of -N and -i in wget of alpine (the one ODK depend on now).
 # It downloads all patterns specified in external.txt
--- a/template/src/ontology/Makefile.jinja2
+++ b/template/src/ontology/Makefile.jinja2
@@ -634,7 +634,7 @@ pattern_schema_checks dosdp_validation: $(TMPDIR)/pattern_schema_checks
 
 .PHONY: update_patterns
 update_patterns: download_patterns
-	cp -r $(TMPDIR)/dosdp/*.yaml $(PATTERNDIR)/dosdp-patterns
+	if [ -f $(TMPDIR)/dosdp/*.yaml ]; then cp -r $(TMPDIR)/dosdp/*.yaml $(PATTERNDIR)/dosdp-patterns; fi
 
 # This command is a workaround for the absence of -N and -i in wget of alpine (the one ODK depend on now).
 # It downloads all patterns specified in external.txt
--- a/template/src/ontology/Makefile.jinja2
+++ b/template/src/ontology/Makefile.jinja2
@@ -634,7 +634,7 @@ pattern_schema_checks dosdp_validation: $(TMPDIR)/pattern_schema_checks
 
 .PHONY: update_patterns
 update_patterns: download_patterns
-	if [ -f $(TMPDIR)/dosdp/*.yaml ]; then cp -r $(TMPDIR)/dosdp/*.yaml $(PATTERNDIR)/dosdp-patterns; fi
+	if [ -d $(TMPDIR)/dosdp ]; then cp -r $(TMPDIR)/dosdp/*.yaml $(PATTERNDIR)/dosdp-patterns; fi
 
 # This command is a workaround for the absence of -N and -i in wget of alpine (the one ODK depend on now).
 # It downloads all patterns specified in external.txt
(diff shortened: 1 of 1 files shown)


matentzn (approved):
Awesome, thank you!

matentzn:
As the pipeline failure seems relevant, can you track down what is causing it?

anitacaron:
It's related to `check_rdfxml_assets`. Not sure what the error is.

matentzn:
When running this test in ODK: tests/test-dosdp-matches.yaml, the RDFXML syntax checking is failing.. since this seems a very relevant test for this PR, its worth looking into it. You can just run the test in isolation on your machine, it takes about 2 seconds.

anitacaron:
I run it and didn't get this error, also when using in CL.

anitacaron:
Maybe something is wrong in more general because it also failed in the same place in [this PR](https://github.com/INCATools/ontology-development-kit/actions/runs/3482211416) where I only edited a comment. So I also think that this PR is not triggering the failure.

matentzn:
Oh oh.. :/

matentzn:
Sorry about that but can you look into it? :/ or ask Damian for help?

anitacaron:
I think @gouttegd would solve it better. I didn't change anything. 😬

gouttegd:
On it. No clue yet. Stay tuned.

gouttegd:
Can’t reproduce here under any conditions I can think of, and there are been no changes since the last successful test that could explain the failure… This one is gonna be fun.

As an aside though, on the PR itself: could there be cases where we would expect the `$(TMPDIR)/dosdp` directory to contain more than _one_ `.yaml` file? Because if so, there is a problem here:

```sh
if [ -f $(TMPDIR)/dosdp/*.yaml ]; then cp -r $(TMPDIR)/dosdp/*.yaml $(PATTERNDIR)/dosdp-patterns; fi
```

If there are several `.yaml` files, the shell will expand what’s inside the brackets to something like `[ -f $(TMPDIR)/dosdp/file1.yaml $(TMPDIR)/dosdp/file2.yaml $(TMPDIR)/dosdp/file3.yaml ]`, which will cause a `too many arguments` error because the `-f` operator only expects a single argument.

gouttegd:
So I still can’t reproduce, but after applying this PR, trying to run the `test-dosdp-matches` causes my Docker to violently crash… What the hell…?

anitacaron:
@gouttegd I think it's expected only one YAML file is in this folder.
However, is this should be the same as in [lines 643-645](https://github.com/INCATools/ontology-development-kit/pull/739/files#diff-ab539569446338e97f29dddc47eccfb63d931e4efa0157d240796a7e9add1d68L643-L645), @matentzn?

matentzn:
No there can be hundreds of files in that folder: 

https://github.com/obophenotype/upheno/tree/master/src/patterns/dosdp-patterns

Well spotted @gouttegd!

@anitacaron What should be the same?

gouttegd:
This bug is unbelievable… Opening another ticket because I am quite sure this is completely unrelated to this PR.

template/src/ontology/Makefile.jinja2
```diff
@@ -634,7 +634,7 @@ pattern_schema_checks dosdp_validation: $(TMPDIR)/pattern_schema_checks
 
 .PHONY: update_patterns
 update_patterns: download_patterns
-	if [ -f $(TMPDIR)/dosdp/*.yaml ]; then cp -r $(TMPDIR)/dosdp/*.yaml $(PATTERNDIR)/dosdp-patterns; fi
+	if [ -d $(TMPDIR)/dosdp ]; then cp -r $(TMPDIR)/dosdp/*.yaml $(PATTERNDIR)/dosdp-patterns; fi
```
matentzn, line 637: This wont work:

```
$ ls tmp/dosdp/

$ cp -r tmp/dosdp/*.yaml .
cp: tmp/dosdp/*.yaml: No such file or directory
```

anitacaron: Sorry, I don't understand. The issue is not to fail when `$(TMPDIR)/dosdp` doesn't exist. 
I tested it here in CL, and it works for this.

```bash
➜  ontology git:(master) ✗ sh run.sh make update_patterns
if [ true  = true ]; then rm -f tmp/dosdp/*.yaml.1 || true; fi
if [ true  = true ] && [ -s ../patterns/dosdp-patterns/external.txt ]; then wget -i ../patterns/dosdp-patterns/external.txt --backups=1 -P tmp/dosdp; fi
if [ true  = true ]; then rm -f tmp/dosdp/*.yaml.1 || true; fi
if [ -d tmp/dosdp ]; then cp -r tmp/dosdp/*.yaml ../patterns/dosdp-patterns; fi
➜  ontology git:(master) ✗ 
```

matentzn: Yes, but if you have tmp/dosdp/ directory, but no patterns in it the pipeline will still fail! (and it should not)..

matentzn (changes requested):
dosdp dir check insufficient

template/src/ontology/Makefile.jinja2
```diff
@@ -634,7 +634,7 @@ pattern_schema_checks dosdp_validation: $(TMPDIR)/pattern_schema_checks
 
 .PHONY: update_patterns
 update_patterns: download_patterns
-	cp -r $(TMPDIR)/dosdp/*.yaml $(PATTERNDIR)/dosdp-patterns
+	if [ -d $(TMPDIR)/dosdp ]; then if [ "$(ls $(TMPDIR)/dosdp)" ]; then cp -r $(TMPDIR)/dosdp/*.yaml $(PATTERNDIR)/dosdp-patterns; fi; fi
```
anitacaron, line 637: This works on the terminal but not calling `make update_patterns`. I don't know why.

gouttegd: Because the `$(ls ...)` is interpreted by Make – the shell never sees it. You need to escape the `$` character so that it is passed to the shell: `$$(ls $(TMPDIR)/dosdp)`.

anitacaron: Ah, thanks, Damien!

matentzn:
@gouttegd i trust you more with this stuff..  an you take a look at that one liner?

gouttegd:
```sh
if [ -d $(TMPDIR)/dosdp ]; then if [ "$$(ls $(TMPDIR)/dosdp)" ]; then cp -r $(TMPDIR)/dosdp/*.yaml $(PATTERNDIR)/dosdp-patterns; fi; fi
```

This is fine, _if_ we never expect the `$(TMPDIR)/dosdp` to contain something else than `*.yaml` files (that is, either it contains some `*.yaml` files – possibly with some other files, doesn’t matter –, or it is empty). If it contains any other type of file but no `*.yaml` files, then the two `if` conditions will be verified (the directory exists and it contains _something_), but the `cp` command will still fail.

Something like `"$$(ls $(TMPDIR)/dosdp/*.yaml)"` won’t do, because if there are no `*.yaml` files the shell will complain that it couldn’t match the pattern.

I think the cleanest way here is rely on `find` instead:

```sh
if [ -n "$$(find $(TMPDIR)/dosdp -type f -name '*.yaml)" ]; then cp ... fi
```

With that, the shell will no longer try to expand the pattern (it’s in single quotes), so it will not matter whereas there are `*.yaml` files or not – the `find` command will work all the same, but it will only output something if there are some files.

Actually we can even get rid of the first test on the existence of the `dosdp` directory if we use the following:

```sh
if [ -n "$$(find $(TMPDIR) -type f -path '$(TMPDIR)/dosdp/*.yaml')" ]; then cp ... fi
```

If `find` returns anything at all, it will mean both that the `dosdp` directory exists and that it contains at least one `*.yaml` file.

anitacaron:
Thank you, @gouttegd, for the explanations.

gouttegd (approved):

Issue #743, opened 2022-11-18
Random failure of the `test-dosdp-matches` check


gouttegd:
Over the last few days, we have observed repeated failures of the CI suite. The test that fails is always `test-dosdp-matches`, and the error is always in the RDF/XML validation:

```
Checking RDF/XML file tdm.owl...
  LightRDF: Unexpected EOF during reading Comment
```

While the failure is always the same _when it occurs_, it does not always occur; sometimes the check goes just fine.

I have managed to reproduce the failure just once on my local machine. I attach to this ticket the corresponding `tdm-bad.owl` file. Indeed, trying to parse that file with LightRDF consistently yields the “Unexpected EOF during reading Comment”, even though I can’t see anything wrong with that file (all other RDF/XML-consuming tools have no trouble to read it).

I also upload a `tdm-good.owl` file, obtained after a test run that didn’t produce any error. The only differences between the two is that a handful of `oboInOwl:hasDbXref` statements either have or have not a `rdf:datatype="http://www.w3.org/2001/XMLSchema#string"` qualifier.

> **Problem 1**: How comes that the produced `tdm.owl` file is not reproductible from one run to the next? Why do some `oboInOwl:hasDbXref` statements have a `rdf:datatype` qualifier while some others have not, and why does that change between runs?

The LightRDF error seems to be caused by anyone of the statements like this one:

```xml
<oboInOwl:hasDbXref rdf:datatype="http://www.w3.org/2001/XMLSchema#string">PRO:DAN</oboInOwl:hasDbXref>
```

Removing the `rdf:datatype` qualifier altogether clears the error. Changing the datatype to anything else than `XMLSchema#string` (even something that is not a valid XMLSchema type) clears the error. Keeping the qualifier as it is but removing the colon within the value (e.g., changing the value to `PRODAN`, without a colon) clears the error.

> **Problem 2**: What the hell is LightRDF doing here?

Suggestions:
* Investigate what clearly seems to be a hideous bug in LightRDF (or the underlying Rust parser). Try to produce a minimal file that reproduces the error so that we can report the bug to LightRDF upstream.
* In the meantime, disable the LightRDF-based RDF/XML validation check. Checking that our produced RDF/XML files are correct is sure nice, but not if we cannot trust the validator for not barking on what appears to be a perfectly correct RDF/XML file!
* Meditate on the value of a test suite that does not produce consistently reproductible results.
* Meditate on the brittleness of our tools.

[...]

PR #745, opened 2022-11-18
Disable the LightRDF-based RDF/XML validation check.

gouttegd:
We cannot rely on LightRDF to not incorrectly flag correct RDF/XML files as being unparseable, so we disable the LightRDF-based check by defult.

We remove the `extra_rdfxml_checks` ODK configuration parameter (which has never been in a public release of the ODK anyway). We only keep the `ensure_valid_rdfxml` parameter instead, which is now OFF by default (as it triggers the Jena-based and RDFLib-based checks, which are time-consuming).

The `check_rdfxml` script now accept `--lightrdf`, `--rdflib`, and `--jena` options to enable the respective checks, with RDFLib and Jena being enabled by default (but the script itself is only called as part of the standard workflow if the `ensure_valid_rdfxml` ODK parameter is ON). This will allow for easy changing of the validators to used by default in the future (including re-enabling the LightRDF validator, if/when it is fixed).

related to #743

### DIFF
--- a/odk/odk.py
+++ b/odk/odk.py
@@ -576,11 +576,8 @@ class OntologyProject(JsonSchemaMixin):
     robot_report : Dict[str, Any] = field(default_factory=lambda: ReportConfig().to_dict())
     """Block that includes settings for ROBOT report, ROBOT verify and additional reports that are generated"""
 
-    ensure_valid_rdfxml : bool = True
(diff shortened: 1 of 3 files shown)


matentzn (approved):
This is pretty amazing. Thanks for that.

PR #748, opened 2022-12-01
Remove JQ from ODK

### LINKED ISSUE #696: Check if time consuming jq pipeline is still necessary
In the last ODK release, we switched to ROBOT 1.9.0 which comes with the latest OBO Graphs, which should be entirely immutable with stable serialisation. `jq` should no longer be necessary, we should try to remove it.

matentzn:
Fixes #696


gouttegd (approved):
Looks fine, but I worry about the (unrelated) failure to install `kgx` when building the image… Won’t have time to investigate that until next week…

PR #754, opened 2022-12-05
WIP: editing obsoletes.sparql

### LINKED ISSUE #499: Automate mapping of deprecated terms in use
This is, IMO, a major missing component of the ODK

* MVP - generate a report of mappings for obsoleted terms every time imports are updated.  We already have suitable SPARQL for finding mappings `src/sparql/obsoletes.sparql`.   
   * This  should be extended to: 
      * Report label and comments on obsoleted term if present
      * Report label of mapped term
      * Report if ID of obsolete term is the value of an OboInOWL#alt_id  axiom (this indicates obsoletion was due to OBO merge. 
   * Potential problem: lack of range restrictions or checks on the value of replaced_by & consider axioms could cause problems for mapping to term referred to.  We need to support the following possibilities:
      * OBO curie - type string
      * IRI (typed)
      * short_form ID (?) 
        
* Complete functionality
    * ODK option to auto-update editors file & template tsvs for every obsolete term for which there is a *single* replacement mapped via 'replaced_by'.  Automated update should be accompanied by a report of all mappings.

matentzn:
Fixes #499 

Hey @anitacaron 

During branch clean-up I noticed this.. 

Should this be merged or discarded?

### DIFF
--- a/template/src/sparql/obsoletes.sparql
+++ b/template/src/sparql/obsoletes.sparql
@@ -2,13 +2,20 @@ prefix xsd: <http://www.w3.org/2001/XMLSchema#>
 PREFIX owl: <http://www.w3.org/2002/07/owl#>
 PREFIX rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>
 PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#>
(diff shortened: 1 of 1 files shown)


matentzn:
No problem, I wont delete it! I will just leave it here as a draft.

PR #755, opened 2022-12-05
Make sure run.sh.env is always deleted.

gouttegd:
To prevent people from inadvertently committing the run.sh.env file, we make sure it is always deleted when the run.sh script terminates. We do that by deleting the file in a EXIT trap, which is systematically executed when the script terminates for any reason.

Not all shells execute the EXIT trap when the script termimates upon being interrupted by a signal though, as this is not a POSIX-mandated behaviour (Bash, Dash, and seemingly Zsh do so). For stricter POSIX-compliant shells, we also trap the SIGINT signal, so that the cleanup call will be executed if the user stops a long-running invocation with Ctrl-C.

### DIFF
--- a/template/src/ontology/run.sh.jinja2
+++ b/template/src/ontology/run.sh.jinja2
@@ -57,6 +57,9 @@ if [ -n "$ODK_BINDS" ]; then
 fi
 
 # Gather environment variables into a single file
+# (and make sure that file is deleted no matter what)
+trap "rm -f run.sh.env" EXIT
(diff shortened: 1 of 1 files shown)


matentzn (approved):
I have never see thin before. This looks great, I assume you tried this, so lets just merge it in and test

gouttegd:
Tried on GNU/Linux and macOS with Ash, Bash, Dash, and Zsh, which should cover enough of our user base.

PR #756, opened 2022-12-06
Update run.sh.jinja2 (source 

matentzn:
This was a bug uncovered by @souzadevinicius

Tested on ARCH linux.. Will this cause issues?


gouttegd (approved):
Interesting. I never realised `source` was searching in the PATH. Good to know. And no, no issues foreseen here, all to the contrary.

PR #768, opened 2023-01-06
Use UID/GID of the calling user inside the container

gouttegd:
This PR is another attempt at fixing #682.

Instead of calling `docker run` with `-u $(id -u):$(id -g)` (which does the job but introduces various side-effects, mostly related to the fact that the user ID that is used has no home directory and no entry in `/etc/passwd`), we use an [entrypoint script](https://docs.docker.com/engine/reference/builder/#entrypoint) that will be executed when we “enter” the container. That script creates a dedicated `odkuser` account, which is given the UID and GID of the calling user (passed in the `run.sh` script as new environment variables). The entrypoint script then executes the command to run with the privileges of the newly created user account.

This allows to simultaneously ensure:
* that the command we run inside the container runs with the privileges of the user who invoked the `run.sh` script, thereby ensuring that any files created by that command have correct ownership;
* that the user account used to run the command does have a home directory (`/home/odkuser`) and a `/etc/passwd` entry.

The latter point will allow to run commands such as `pip install` to install supplementary Python packages in the `odkuser` user’s home directory.

As for the previous attempt to fix that problem (#684), a consequence of that PR is that running commands that require _root_ privileges inside the container (such as `apt-get install` to install supplementary packages not part of the image) will no longer be possible. If people really need that we can add a ODK-level option like `run_as_root` or similar.

### DIFF
--- a/docker/odklite/Dockerfile
+++ b/docker/odklite/Dockerfile
@@ -32,7 +32,8 @@ RUN apt-get update && \
         curl \
         jq \
         rsync \
-        time
+        time \
+        sudo
 
 # Install Python environment (copied over from the builder image).
 COPY --from=obolibrary/odkbuild:latest /staging/lite /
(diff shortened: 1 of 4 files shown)


gouttegd:
Superseded by #769

PR #769, opened 2023-01-08
Use dedicated user account with UID/GID of the calling user inside the container

gouttegd:
This PR is another attempt at fixing #682.

Instead of calling `docker run` with `-u $(id -u):$(id -g)` (which does the job but introduces various side-effects, mostly related to the fact that the user ID that is used has no home directory and no entry in `/etc/passwd`), we use an [entrypoint script](https://docs.docker.com/engine/reference/builder/#entrypoint) that will be executed when we “enter” the container. That script creates a dedicated `odkuser` account, which is given the UID and GID of the calling user (passed in the `run.sh` script as new environment variables). The entrypoint script then executes the command to run with the privileges of the newly created user account.

This allows to simultaneously ensure:
* that the command we run inside the container runs with the privileges of the user who invoked the `run.sh` script, thereby ensuring that any files created by that command have correct ownership;
* that the user account used to run the command does have a home directory (`/home/odkuser`) and a `/etc/passwd` entry.

The latter point will allow to run commands such as `pip install` to install supplementary Python packages in the `odkuser` user’s home directory.

As for the previous attempt to fix that problem (#684), a consequence of that PR is that running commands that require _root_ privileges inside the container (such as `apt-get install` to install supplementary packages not part of the image) will no longer be possible. If people really need that we can add a ODK-level option like `run_as_root` or similar.


gouttegd:
@matentzn I’ve updated the PR with the changes we’ve briefly discussed in the Tech Support meeting yesterday:

People who need their workflow to be run with root privileges can explicitly set the `ODK_USER_ID` variable to 0 (best place to do so would be in the `run.sh.conf` file, though setting it as environment variable would work as well). Then the entrypoint script would directly execute the workflow while staying under the identity of the superuser, instead of creating an unprivileged `odkuser` account and sudo-ing into it.

docker/odklite/Dockerfile
```diff
@@ -76,6 +77,14 @@ COPY scripts/check-rdfxml.sh /tools/check-rdfxml
 # Force Git to accept working on a repository owned by someone else
 RUN echo "[safe]\n    directory = /work" >> /etc/gitconfig
 
+# Make sure we run under an existing user account with UID/GID
+# matching the UID/GID of the calling user
+COPY scripts/entrypoint.sh /usr/local/sbin/odkuser-entrypoint.sh
+RUN chmod 755 /usr/local/sbin/odkuser-entrypoint.sh
+ENTRYPOINT ["/usr/local/sbin/odkuser-entrypoint.sh"]
```
matentzn, line 84: Will this have any consequences to GitHub actions using ODK and people that use the docker container without the run.sh script?

gouttegd: Same consequences as for all other users: the processes that run within the container will run under the identify of an unprivileged user.

If they need those processes to run with root privileges and can’t use the `run.sh` script for any reason, then it’s up to them to explicitly pass the `ODK_USER_ID=0` environment variable when they invoke Docker.

gouttegd:
Not sure I understand why the test suite was not triggered on the last three commits. Has anything changed in our CI configuration recently?

matentzn:
Nope, I think this is the problem: https://github.com/INCATools/ontology-development-kit/blob/master/.github/workflows/build-and-test.yml#L8

probably needs "edited" or "updated" or some such?

gouttegd:
Indeed… And I am the one who came up with that, actually: https://github.com/INCATools/ontology-development-kit/commit/3a82f6cd897110f2f4f5f916d113d9a2fde2ceba 🤦

gouttegd:
> Configure the build-and-test workflow to be automatically triggered only when a PR is opened, not everytime it is updated with new commits (to save runtime credits)

This was a good idea at the time, but this was before we added the extra check to prevent the test suite from being triggered when a PR only affects “inconsequential“ files (typically documentation). Now that we have this extra check, I’d be inclined to restore triggering the test suite upon any addition of new commits.

matentzn:
I think a trick to work around that in case you prefer to reduce QC load: "close pr" -> "reopen". But I am inclined to do whatever you think is best!

gouttegd:
Nice trick indeed, but if we ask PR authors to do something like that to manually trigger the test suite when they update their PRs, we might as well ask them to explicitly trigger the action in the “Actions” tab (as we currently do, it’s just that I had forgotten about it), rather than overloading the meaning of closing/re-opening a PR.

I’d suggest to leave things as they are for now.

matentzn:
This is not ideally, but we could mitigate the uncontrolled running of QC somewhat with https://stackoverflow.com/questions/68349031/only-run-actions-on-non-draft-pull-request?

gouttegd:
This seems to me like just another attempt at overloading an action (here the act of marking a PR as “draft” or “non-draft”) with another, unrelated meaning (“run the test suite on that PR“). Not keen to do that.

If we want a way for submitters to avoid triggering the test suite, I think it should be explicit. Maybe a checkbox at the end of the issue template:

```
[ ] No-CI: Do _not_ run the test suite on this PR until this checkbox is unchecked
```

and then have the GitHub action verify the checkbox status and abort without doing anything if it is checked?

gouttegd:
Ultimately though, I think the best way to reduce our overlong CI check is to separate the toolbox from the workflow (#877).

With the toolbox and the workflow maintained together as they currently are, a single change to the workflow (e.g. a change in `template/src/ontology/Makefile.jinja2`) causes the test suite to rebuild the entire ODK, even if there were no changes affecting the contents of the ODK image. That’s very inefficient.

matentzn:
> If we want a way for submitters to avoid triggering the test suite, I think it should be explicit. Maybe a checkbox at the end of the issue template:

Love that! Can we still impose that the PR cannot be merged until CI passes? This would perhaps be complicated..

gouttegd:
> Can we still impose that the PR cannot be merged until CI passes?

Not sure, but I would assume it must be possible to somehow explicitly _error out_ in a GitHub Action.

That is, if the checkbox is ticked (meaning the test suite should not be run), instead of normally terminating the action (which would be interpreted by GitHub as a successful run, and would satisfy the condition required to enable merging), we should terminate the action _abnormally_ (the equivalent of returning a non-zero error code in a command-line program), so that the PR is marked as having failed the CI check (even though in reality we skipped the entire test suite). If the “CI checks must pass before merging is enabled” rule is active, this would prevent from merging the PR until the checkbox is unchecked and the CI checks pass.

But!

Remember that our current test suite is already configured not to run if a PR changes some “unimportant“ files (e.g., documentation). I am not sure how the “CI checks must pass before merging is enabled” rule would interact with that. Such a rule could force us to run the test suite even on changes where we know it is completely pointless to run it, which is exactly what we have been trying to avoid.

Overall, I’m pretty skeptical of the benefits of the “CI checks must pass before merging is enabled” rule. CI is good, but in the end I trust human review and approval much, much more than passing a CI check. A “approval required before merging is enabled“ rule would be much more useful, in my opinion. If only because a human reviewer can always explicitly require that the test suite be run — by refusing to grant approval otherwise (“your changes look quite good. Uncheck the No-CI box to enable the test suite. Once it passes, I’ll approve.“).

gouttegd:
> A “approval required before merging is enabled“ rule would be much more useful, in my opinion.

Of note, since we currently do not have such a rule, it’s only my natural respect for your authority that is preventing me from merging #853. :p

matentzn (approved):

PR #782, opened 2023-01-25
Update requirements.txt.full

joeflack4:
- Added `tabulate` package requirement. This will be used by `mondo-ingest` to convert data frames into markdown tables.


matentzn (approved):
Thanks, good package to add!

PR #786, opened 2023-02-03
New parameter `mirror_type`

### LINKED ISSUE #632: Add new parameter: mirror_type: custom, none
Right now, when doing a module_type: custom, we still get the mirror goal, which is often just wrong, for example when the custom module does not need a mirror. (ontofox, wikidata API, etc)

anitacaron:
Fixes #632

### DIFF
--- a/odk/odk.py
+++ b/odk/odk.py
@@ -151,7 +151,7 @@ class ImportProduct(Product):
     """if use_gzipped is true, try use the base IRI instead of normal one to mirror from."""
 
     mirror_type: Optional[str] = None
-    """If module_type is custom, define the type of the mirror. Supported: custom, no_mirror."""
+    """Define the type of the mirror for your import. Supported: base, custom, no_mirror."""
 
 @dataclass_json
 @dataclass
--- a/odk/odk.py
+++ b/odk/odk.py
@@ -151,7 +151,7 @@ class ImportProduct(Product):
     """if use_gzipped is true, try use the base IRI instead of normal one to mirror from."""
 
     mirror_type: Optional[str] = None
-    """If module_type is custom, define the type of the mirror. Supported: custom, none."""
+    """If module_type is custom, define the type of the mirror. Supported: custom, no_mirror."""
 
 @dataclass_json
 @dataclass
--- a/odk/odk.py
+++ b/odk/odk.py
@@ -130,7 +130,7 @@ class ImportProduct(Product):
     """if large, ODK may take measures to reduce the memory footprint of the import."""
     
     module_type : Optional[str] = None
-    """Module type. Supported: slme, minimal, custom"""
+    """Module type. Supported: slme, minimal, custom, mirror"""
     
     module_type_slme : str = "BOT"
     """SLME module type. Supported: BOT, TOP, STAR"""
@@ -149,6 +149,9 @@ class ImportProduct(Product):
     
     use_gzipped: bool = False
     """if use_gzipped is true, try use the base IRI instead of normal one to mirror from."""
+
+    mirror_type: Optional[str] = None
+    """If module_type is custom, define the type of the mirror. Supported: custom, none."""
 
 @dataclass_json
 @dataclass
--- a/template/src/ontology/Makefile.jinja2
+++ b/template/src/ontology/Makefile.jinja2
@@ -444,7 +444,7 @@ $(IMPORTDIR)/{{ ont.id }}_import.owl: {% if project.import_group.use_base_mergin
 		remove $(patsubst %, --term %, $(ANNOTATION_PROPERTIES)) -T $(IMPORTDIR)/{{ ont.id }}_terms_combined.txt --select complement --select "classes individuals annotation-properties" \
 		$(ANNOTATE_CONVERT_FILE); fi
 {% elif ('custom' == ont.module_type) or (ont.module_type is none and 'custom' == project.import_group.module_type) %}
-$(IMPORTDIR)/{{ ont.id }}_import.owl: {% if project.import_group.use_base_merging %}$(MIRRORDIR)/merged.owl{% elif 'none' != ont.mirror_type %}$(MIRRORDIR)/{{ ont.id }}.owl{% endif %}
+$(IMPORTDIR)/{{ ont.id }}_import.owl: {% if project.import_group.use_base_merging %}$(MIRRORDIR)/merged.owl{% elif 'no_mirror' != ont.mirror_type %}$(MIRRORDIR)/{{ ont.id }}.owl{% endif %}
 	echo "ERROR: You have configured your default module type to be custom; this behavior needs to be overwritten in {{ project.id }}.Makefile!" && false
 {%- endif %}
 {%- endif %}
@@ -535,14 +535,12 @@ IMP_LARGE=true # Global parameter to bypass handling of large imports
 
 {% for ont in project.import_group.products %}
 
-{% if ont.mirror_type != 'none' -%}
 ## ONTOLOGY: {{ ont.id }}
-{% endif -%}
 {% if ont.description  -%}
 ## {{ ont.description }}
 {% endif -%}
 
-{%- if ont.mirror_type != 'none' -%}
+{%- if ont.mirror_type != 'no_mirror' -%}
 .PHONY: mirror-{{ ont.id }}
 .PRECIOUS: $(MIRRORDIR)/{{ ont.id }}.owl
 {%- endif -%}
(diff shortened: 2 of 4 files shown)


template/src/ontology/Makefile.jinja2
```diff
@@ -444,7 +444,7 @@ $(IMPORTDIR)/{{ ont.id }}_import.owl: {% if project.import_group.use_base_mergin
 		remove $(patsubst %, --term %, $(ANNOTATION_PROPERTIES)) -T $(IMPORTDIR)/{{ ont.id }}_terms_combined.txt --select complement --select "classes individuals annotation-properties" \
 		$(ANNOTATE_CONVERT_FILE); fi
 {% elif ('custom' == ont.module_type) or (ont.module_type is none and 'custom' == project.import_group.module_type) %}
-$(IMPORTDIR)/{{ ont.id }}_import.owl: {% if project.import_group.use_base_merging %}$(MIRRORDIR)/merged.owl{% else %}$(MIRRORDIR)/{{ ont.id }}.owl{% endif %}
+$(IMPORTDIR)/{{ ont.id }}_import.owl: {% if project.import_group.use_base_merging %}$(MIRRORDIR)/merged.owl{% elif 'none' != ont.mirror_type %}$(MIRRORDIR)/{{ ont.id }}.owl{% endif %}
```
matentzn, line 447: So what should happen here if mirror type is none? Does this mean there  _is no mirror_? If so its risky to make this default!

odk/odk.py
```diff
@@ -150,6 +150,9 @@ class ImportProduct(Product):
     use_gzipped: bool = False
     """if use_gzipped is true, try use the base IRI instead of normal one to mirror from."""
 
+    mirror_type: Optional[str] = None
```
matentzn, line 153: is `None` here not something else then the string `'none'` you look for later?

template/src/ontology/Makefile.jinja2
```diff
@@ -535,13 +535,17 @@ IMP_LARGE=true # Global parameter to bypass handling of large imports
 
 {% for ont in project.import_group.products %}
 
+{% if ont.mirror_type != 'none' -%}
 ## ONTOLOGY: {{ ont.id }}
+{% endif -%}
```
matentzn, line 540: Is this "endif" at the right place? Why you you skop the ONTOLOGY comment but not the "description" comment?

template/src/ontology/Makefile.jinja2
```diff
@@ -566,6 +570,10 @@ mirror-{{ ont.id }}: | $(TMPDIR)
 		$(ROBOT) convert -i $(MIRRORDIR)/{{ ont.id }}.owl.gz -o $@.tmp.owl{%- if ont.make_base %} && \
 		$(ROBOT) remove -i $@.tmp.owl {% if ont.base_iris is not none %}{% for iri in ont.base_iris %}--base-iri {{iri}} {% endfor %}{% else %}--base-iri $(URIBASE)/{{ ont.id.upper() }} {% endif %}--axioms external --preserve-structure false --trim false -o $@.tmp.owl{% endif %} &&\
 		mv $@.tmp.owl $(TMPDIR)/$@.owl; fi
+{%- elif 'custom' == ont.module_type and 'custom' == ont.mirror_type %}
+mirror-{{ ont.id }}: 
+	echo "ERROR: You have configured your default mirror type to be custom; this behavior needs to be overwritten in {{ project.id }}.Makefile!" && false
+{%- elif 'custom' == ont.module_type and 'none' == ont.mirror_type -%}
```
matentzn, line 576: Why do you need to couple the `module_type` and the `mirror_type` here so tightly? Couldnt the be configured separately?

matentzn (changes requested):
@anitacaron excellent thinking of designing a `mirror_type`; this is the most flexible and appropriate way to solve many of our problems, and I never though of it. Super! It will take a few rounds to get this right, hope you bear with me. Thanks!
 
- [x] The use of 'none' string is too scary in a makefile, because its not clear if this is a string of "undefined" (no variable set). Can we rename to `no_mirror`?
 - [x] The idea with mirror_type is great. I don't think you need to couple it with module_type though (by couple I mean, check them both). The mirror can be custom while the module_type is fixed, and the other way around. 
 - [ ] The natural continuation would be to implement `mirror_type: base` - don't remove `make_base: TRUE` but the same logic could apply here.
 - [ ] This PR should have no effect at all on (1) any currently implemented pipeline anywhere, and (2) on the imports goals.

odk/odk.py
anitacaron: No. It's the meaning in the python. It's an optional string value, then the default is None. Same as `module_type`.

template/src/ontology/Makefile.jinja2
anitacaron: Yes, none means no mirror. Can you detail it more?

anitacaron: This is the main reason for the issue we want to fix, no?

template/src/ontology/Makefile.jinja2
anitacaron: I thought again, and it's better if the comments stay there with the comment below when there isn't a mirror.
I skipped only the phony/precious part.

template/src/ontology/Makefile.jinja2
anitacaron: If there isn't a mirror, it won't add the mirror file.

anitacaron:
- I've changed the value `none` to `no_mirror`.

odk/odk.py
matentzn: Ok

template/src/ontology/Makefile.jinja2
```diff
@@ -396,7 +396,7 @@ $(IMPORTDIR)/%_import.owl: {% if project.import_group.use_base_merging %}$(MIRRO
 		query --update ../sparql/inject-subset-declaration.ru --update ../sparql/inject-synonymtype-declaration.ru --update ../sparql/postprocess-module.ru \
 		$(ANNOTATE_CONVERT_FILE); fi
 {% elif 'custom' == project.import_group.module_type %}
-$(IMPORTDIR)/%_import.owl: {% if project.import_group.use_base_merging %}$(MIRRORDIR)/merged.owl{% else %}$(MIRRORDIR)/%.owl{% endif %}
+$(IMPORTDIR)/%_import.owl: {% if project.import_group.use_base_merging %}$(MIRRORDIR)/merged.owl{% elif 'custom' == ont.mirror_type %}$(MIRRORDIR)/%.owl{% endif %}
```
matentzn, line 399: I dont like this.. ODK should never change the default behaviour! Even if its `custom` module type, the mirror can be the default normal. I think the condition here should be 

```
{% elif 'no_mirror' != ont.mirror_type %}
```

matentzn: Correct me if I am wrong though!

matentzn (changes requested):
- [x] Also add a row in CHANGELOG.md

template/src/ontology/Makefile.jinja2
anitacaron: I removed the check for both parameters.

template/src/ontology/Makefile.jinja2
anitacaron: You're right. This should be in sync with the other import part.

template/src/ontology/Makefile.jinja2
```diff
@@ -540,19 +540,21 @@ IMP_LARGE=true # Global parameter to bypass handling of large imports
...
 
+{%- if ont.mirror_type != 'no_mirror' -%}
 .PHONY: mirror-{{ ont.id }}
 .PRECIOUS: $(MIRRORDIR)/{{ ont.id }}.owl
+{%- endif -%}
 
 {%- if ont.mirror_from %}
 mirror-{{ ont.id }}: | $(TMPDIR)
-	if [ $(MIR) = true ] && [ $(IMP) = true ]{% if ont.is_large %} && [ $(IMP_LARGE) = true ]{% endif %}; then $(ROBOT) convert -I {{ ont.mirror_from }} -o $@.tmp.owl{%- if ont.make_base %} && \
+	if [ $(MIR) = true ] && [ $(IMP) = true ]{% if ont.is_large %} && [ $(IMP_LARGE) = true ]{% endif %}; then $(ROBOT) convert -I {{ ont.mirror_from }} -o $@.tmp.owl{%- if ont.make_base or 'base' == ont.mirror_type %} && \
```
matentzn, line 550: I am assuming this is the only time where `if ont.make_base` was needed..

matentzn: Oh foregt it.

matentzn (approved):

PR #788, opened 2023-02-06
Fix memory issue when import is large and uses module type mirror

### LINKED ISSUE #785: The combination is_large: TRUE does not work with make_base: TRUE

anitacaron:
Fixes #785


matentzn (approved):
Looks good, thank you!

PR #789, opened 2023-02-07
New parameter: `sparql_test_on`

### LINKED ISSUE #565: sparql_test: make ontology configurable
Should be able to use base/full instead of src.

anitacaron:
Fixes #565


matentzn (changes requested):
Can you add a testcase for this? Else it looks good.

matentzn (approved):

PR #790, opened 2023-02-08
Apply sssom_options

anitacaron:


matentzn (changes requested):
Move options after the I put argument?

matentzn (approved):
Thanks for catching the path bug!!

PR #793, opened 2023-02-14
Use a pre-built Konclude binary on arm64.

gouttegd:
When building odkfull for x86_64, we use a binary version of Konclude provided directly by upstream. But when building for arm64, there is no binary available and we have to build Konclude from source (we do that in the odkbuild image).

Building Konclude from source is time-consuming (up to several hours), and doing that every time one builds a ODK image is a waste since Konclude has not changed since its last release in 2021.

This PR therefore updates the arm64 build to use a custom build of Konclude that has only been built once and for all by myself. This considerably speeds up the build for arm64.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -51,9 +51,9 @@ RUN DEBIAN_FRONTEND="noninteractive" apt-get install -y --no-install-recommends 
 COPY --from=obolibrary/odkbuild:latest /staging/full /
 
 # Install Konclude.
-# On x86_64, we get it from a pre-built release; on arm64, we
-# have built it in the builder image, now we need to install
-# the run-time dependencies (Qt5).
+# On x86_64, we get it from a pre-built release from upstream; on arm64,
+# we use a custom pre-built binary to which we just need to add the
+# run-time dependencies (the binary is not statically linked).
 ARG TARGETARCH
 RUN test "x$TARGETARCH" = xamd64 && ( \
         wget -nv https://github.com/konclude/Konclude/releases/download/v0.7.0-1138/Konclude-v0.7.0-1138-Linux-x64-GCC-Static-Qt5.12.10.zip \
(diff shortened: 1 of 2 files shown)


gouttegd:
@matentzn You might want to try this to make building the multi-arch images faster. On my machine, building Konclude for arm64 is always what takes the most time, using a pre-built binary makes a significant difference.

Dockerfile
```diff
@@ -64,7 +64,13 @@ RUN test "x$TARGETARCH" = xamd64 && ( \
         rm Konclude.zip \
     ) || ( \
         DEBIAN_FRONTEND="noninteractive" apt-get install -y --no-install-recommends \
-            libqt5xml5 libqt5network5 libqt5concurrent5 \
+            libqt5xml5 libqt5network5 libqt5concurrent5 && \
+        wget -nv https://incenp.org/files/softs/konclude/0.7/Konclude-v0.7.0-1138-Linux-arm64-GCC.zip \
```
matentzn, line 68: I dont oppose this at all, but I would prefer to have this being uploaded to Zenodo instead of your own site - else we may have random outages due to you discontinuing your domain. Also, I think this is just for safety a bit better.

gouttegd: Very reluctant to upload to Zenodo, for several reasons:

* I would need to check whether this is in line with Zenodo’s policy, given that I am _not_ a Konclude author.
* Even if Zenodo happens to allow anyone to upload anything even if it’s not their own software (maybe they do), that does not feel like a correct thing to do. Right now, Konclude is _not_ on Zenodo, and I don’t know if it’s because its authors simply didn’t think about putting it there or if it’s because they don’t want to do it; either way, uploading it myself behind their back feels just wrong, even if it’s technically allowed by the free software license.
* I have only compiled Konclude for the benefits of the ODK, and I don’t plan to advertise the availability of the archive on my website. If I put the archive on a high-visibility website such as Zenodo, I will feel that I have now a responsibility towards anyone who’ll use the arm64 binary build of Konclude, and I don’t want that.
* I disagree about the safety bit. using Zenodo would be _adding_ one more intermediate (Konclude → me → Zenodo → ODK instead of Konclude → me → ODK), so this would actually _increase_ the attack surface.

Ultimately, the best solution here would be to ask the Konclude developers to provide the arm64 build on their GitHub repository, as they do for the x86_64 builds. Putting a custom build on an obscure website like mine is merely a workaround until something like that can happen.

As an aside, my own website has been up since 2009 and Zenodo only exists since 2013, so I can legitimately say that I have a better track record than them in terms of durability. :p

matentzn: Alright, lets pair this with an issue about asking Konclude devs to upload the arm64 dev and I will try to communicate with them to do it!

PR #800, opened 2023-02-20
Add pull request basic rules and minor/major release SOP to contributing

matentzn:


CONTRIBUTING.md
```diff
@@ -257,6 +257,20 @@ The examples folder serves for both unit test and documentation purposes.
 
 TODO
 
+## Pull request rules
+
+1. One PR per feature.
+2. Each PR must link to one or more existing issues.
+3. There should be one commit per logical change. This is important so we can be more effective at cherry picking for patch releases.
```
gouttegd, line 264: If we want to enforce that, the logical next step is to have a rule that when a PR has gone through several cycles of reviews/new-commits (where new commits fix mistakes that were highlighted in previous reviews), then it should not be merged (it’s no longer possible to have atomic commits), but instead its author should submit a new “clean” PR.

Do we want to do that? I would welcome it personally, but from experience it could be very difficult to convince people that it is a good use of their time.

gouttegd (approved):
No objection, but please consider the comment I left. Without a clear policy about what to do when a PR no longer has atomic commits (accept anyway or ask for a re-write), the rule about having one logical change per commit is pointless.

CONTRIBUTING.md
matentzn: This is precisely why I advocate squash merges. Combined with the rule "one logical change per PR", squash merges will achieve exactly what you are suggesting!

gouttegd: No! Squash merges only do that for _simple_ pull requests, when the entire change proposed by the PR could fit in a single logical commit.

For complex PR, when the proposed change touches many files and involves several independent smaller changes, squashing does _not_ achieve what re-submitting a clean PR would do!

Look at PR #769 for an example. It’s not a very complex one, and it is really about a single feature. And yet it is made of several small changes (adding `sudo` to the image, adding the `entrypoint` script, etc.) that should _not_ be squashed into a single commit.

gouttegd: Please do _not_ advocate the use of squash merges except for the simplest of PRs!

If people don’t want to re-submit “clean” PRs once review is over, fine, I’d rather have the PR merged “as is” with dozens of broken commits rather than have it merged as a single huge “do everything” commit.

gouttegd: And it’s not “one logical change per PR”, it’s “one _feature_ per PR”. That’s not the same thing.

A single feature (which should have its own PR) can require dozens of single logical changes (which should have their own commits).

matentzn: Ok :) I never made that distinction (feature and logical change), but I see your point! We will decide how to deal with this on a case by case basis then.

PR #807, opened 2023-02-25
Allow passing configuration options to the OWL API.

### LINKED ISSUE #778: Add environment defaults for OWL API parsing behaviour
For context read this thread: https://github.com/protegeproject/protege/issues/1096#issuecomment-1387292671

We should consider to make it easy in ODK to set some environment variable by default. Can we somehow do this in the Makefile? like `BANNED_PARSERS`?

gouttegd:
Update the run.sh script to check for environment variables with a name starting with `OWLAPI_`. All such variables are converted to OWLAPI Java options appended to the `ODK_JAVA_OPTS` variable, which is passed to the container.

For example, an environment variable `OWLAPI_INDENT_SIZE=2` will be converted to
`-Dorg.semanticweb.owlapi.model.parameters.ConfigurationOptions.INDENT_SIZE=2`.

No validation of the option is performed; it is up to the user to ensure that the name after OWLAPI_ corresponds to a valid OWLAPI configuration parameter.

closes #778

### DIFF
--- a/template/src/ontology/run.sh.jinja2
+++ b/template/src/ontology/run.sh.jinja2
@@ -40,6 +40,14 @@ ODK_TAG=${ODK_TAG:-latest}
 ODK_JAVA_OPTS={% if project.robot_java_args is defined and project.robot_java_args|length %}${ODK_JAVA_OPTS:-{{ project.robot_java_args }}}{% else %}${ODK_JAVA_OPTS:--Xmx8G}{% endif %}
 ODK_DEBUG=${ODK_DEBUG:-no}
 
+# Convert OWLAPI_* environment variables to the OWLAPI as Java options
+# See http://owlcs.github.io/owlapi/apidocs_4/org/semanticweb/owlapi/model/parameters/ConfigurationOptions.html
+# for a list of allowed options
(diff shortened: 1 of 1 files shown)


matentzn (commented):
Will this work in all standard shell/bash combinations?

gouttegd:
I just knew you were going to ask that. :D 

Yes, it will work in a standard Posix-compliant shell. The patch is bashism-free. :) The string substitution syntax (`${var#...}` and $`{var%...}`) is specified in [POSIX 1003.1-2008](https://pubs.opengroup.org/onlinepubs/9699919799.2008edition/utilities/V3_chap02.html#tag_18_06_02).

gouttegd:
By the way, while the #778 issue was raised primarily to offer the possibility to ban the TriX parser when using the ODK, this is actually no longer needed since that parser had already been [banned in OWLAPI 4.5.25](https://github.com/owlcs/owlapi/commit/4df61b3a07031f5e92f209914d2ed6a8e2bfe4e0).

But the feature is still useful beyond that specific use case.

matentzn (approved):
Lets test this then as early as possible! :) Thanks!

Issue #771, opened 2023-01-09
Revert change of -u in ODK 1.3.2


matentzn:
Until we fix #769

PR #808, opened 2023-02-25
Do not set UID/GID in seed-via-docker.sh.

gouttegd:
PR #771 partially reverted a temporarily introduced change in the run.sh script (the use of docker's -u option), but it did not revert a similar change in seed-via-docker.sh. As a result, the test suite was still run under the identity of a non-root user, which caused at least one test to fail (the sssom test, because the sssom tool uses the bioregistry module, which caches some data in the home directory).


matentzn (approved):
Thanks for catching!

gouttegd:
Note that it is still not enough for the `sssom` test to pass, because of a completely unrelated issue specific to that test: #809.

PR #810, opened 2023-02-26
Finalise definition of base release

matentzn:
This tiny PR may be one of the most consequential in the history of ODK, so let's get it 100% right.

What we are doing here is enshrining in stone 2 years worth of debates on how to define base modules for proper ontology dependency management. The debates happened, for better or worse, behind closed doors - it took just four of us (@cmungall, @balhoff, @dosumis and me) roughly 2 years to grudgingly agree on a middle ground. 

The full specification can be found here: https://github.com/OBOAcademy/obook/pull/387/

What does this PR do?

1. The definition of the "base" file is now changed. Previously, the base was simply "all the axioms the editors have provided", without any concerns for reasoning or redundancy. The old base reflects the structure of the editors file with no instantiation of axioms inferred by a reasoner. The new base reflects the structure of the ontology release file, including (a non-redundant set of) whatever instantiated reasoner-inferred axioms the authors of the ontology have chose to include in the release. This may include inferred subClassOf axioms between named classes and materialised existential restrictions. You can check out the exact sequence of ROBOT commands in this pull request to get the picture.
2. The old base release (all the axioms managed by the ontology editors, as they are, with no subclass-graph reduction and no reasoning) is now called "baselite" release
3. The "base merging" approach in ODK now has an option, switched on by default, to remove equivalent class axioms prior to extracting a module. This way externally provided classifications are protected again axiom injection ("trust the source" principle)


template/src/ontology/Makefile.jinja2
```diff
@@ -621,7 +621,7 @@ ALL_MIRRORS = $(patsubst %, $(MIRRORDIR)/%.owl, $(IMPORTS))
 MERGE_MIRRORS = true
 
 $(MIRRORDIR)/merged.owl: $(ALL_MIRRORS)
-	if [ $(IMP) = true ] && [ $(MERGE_MIRRORS) = true ]; then $(ROBOT) merge $(patsubst %, -i %, $^) -o $@; fi
+	if [ $(IMP) = true ] && [ $(MERGE_MIRRORS) = true ]; then $(ROBOT) merge $(patsubst %, -i %, $^) {% if project.import_group.base_merge_drop_equivalent_class_axioms %}remove --axioms equivalent --preserve-structure false {% endif %}-o $@; fi
```
matentzn, line 624: This ensures that when we use the base merging strategy for imports, we can drop equivalent class axioms prior to extracting a module. This protects the canonical classification provided by the external ontology against reclassifications due axiom interactions.

template/src/ontology/Makefile.jinja2
```diff
@@ -912,20 +912,11 @@ ROBOT_RELEASE_IMPORT_MODE_BASE=$(ROBOT) remove --input $< --select imports --tri
...
-
-{% if 'baseplus' in project.release_artefacts or project.primary_release == 'baseplus' -%}
-# baseplus: A highly experimental variant of base, UNSTABLE. Only to be used experimentally (may disappear in future ODK releases).
-$(ONT)-baseplus.owl: $(EDIT_PREPROCESSED) $(OTHER_SRC) $(IMPORT_FILES)
+# base: A version of the ontology that does not include any externally imported axioms.
+$(ONT)-base.owl: $(EDIT_PREPROCESSED) $(OTHER_SRC) $(IMPORT_FILES)
 	$(ROBOT_RELEASE_IMPORT_MODE) \{% if project.release_use_reasoner %}
-	reason --reasoner {{ project.reasoner }} --equivalent-classes-allowed {{ project.allow_equivalents }} --exclude-tautologies {{ project.exclude_tautologies }} \{% endif %}
+	reason --reasoner {{ project.reasoner }} --equivalent-classes-allowed {{ project.allow_equivalents }} --exclude-tautologies {{ project.exclude_tautologies }} \{% if project.release_materialize_object_properties is defined and project.release_materialize_object_properties %}
+	materialize {% for iri in project.release_materialize_object_properties %}--term {{iri}} {% endfor %} \{% endif %}{% endif %}
```
matentzn, line 919: This part adds materialization as an option. Only of the user specifically specifies object properties to materialise in `release_materialize_object_properties`, will they be materialized. There is no way to "materialize everything".

template/src/ontology/Makefile.jinja2
```diff
@@ -935,9 +926,9 @@ $(ONT)-baseplus.owl: $(EDIT_PREPROCESSED) $(OTHER_SRC) $(IMPORT_FILES)
 		{% if project.release_date -%} --annotation oboInOwl:date "$(OBODATE)" {% endif -%}--output $@.tmp.owl && mv $@.tmp.owl $@
 {% endif -%}
 
-{% if 'editors' in project.release_artefacts or project.primary_release == 'editors' -%}
-# editors: All the axioms as they are editted by the editors, excluding reasoning. This is currently the same as "base". Only to be used experimentally (may disappear in future ODK releases).
-$(ONT)-editors.owl: $(EDIT_PREPROCESSED) $(OTHER_SRC)
+{% if 'baselite' in project.release_artefacts or project.primary_release == 'baselite' -%}
+# baselite: All the axioms as they are editted by the editors, excluding reasoning. This is currently the same as "base". Only to be used experimentally (may disappear in future ODK releases).
+$(ONT)-baselite.owl: $(EDIT_PREPROCESSED) $(OTHER_SRC)
```
matentzn, line 931: Due to request by @dosumis I renamed the `editors` release to `baselite`. This release corresponds to the editor provided axioms "as they are", without any processing, reasoning, redundancy stripping etc.

template/src/ontology/Makefile.jinja2
```diff
@@ -947,7 +938,8 @@ $(ONT)-editors.owl: $(EDIT_PREPROCESSED) $(OTHER_SRC)
 # Full: The full artefacts with imports merged, reasoned.
 $(ONT)-full.owl: $(EDIT_PREPROCESSED) $(OTHER_SRC) $(IMPORT_FILES)
 	$(ROBOT_RELEASE_IMPORT_MODE) \
-		reason --reasoner {{ project.reasoner }} --equivalent-classes-allowed {{ project.allow_equivalents }} --exclude-tautologies {{ project.exclude_tautologies }} \
+		reason --reasoner {{ project.reasoner }} --equivalent-classes-allowed {{ project.allow_equivalents }} --exclude-tautologies {{ project.exclude_tautologies }} \{% if project.release_materialize_object_properties is defined and project.release_materialize_object_properties %}
+		materialize {% for iri in project.release_materialize_object_properties %}--term {{iri}} {% endfor %} \{% endif %}
```
matentzn, line 942: I also added the option to "materialise" to the full release, including imports merged in.

tests/test-release-materialise.yaml
```diff
@@ -0,0 +1,24 @@
...
+title: "Release Test A Ontology"
+github_org: obophenotype
+repo: rta
+report_fail_on: ERROR
+release_diff: FALSE
+namespaces: 
+  - http://purl.obolibrary.org/obo/PTT_
+  - http://purl.obolibrary.org/obo/PATO_
+release_materialize_object_properties:
+  - http://purl.obolibrary.org/obo/BFO_0000050
```
matentzn, line 10: This is what is being tested here.

dosumis:
> "base" axioms are axioms that use at least one "base" entity. This is the current definition in ROBOT https://github.com/ontodev/robot/pull/570, and it is not ideal; consider the case EXTERNAL:A subClassOf EXTERNAL:R some BASE:B - we really don't want this in our base release.

From discussion on call, I believe base-plus moved to using axiom tagging.  If so, could this be made clear?  If the interests of progress, we could decide not to make this a blocker, but fixing is *very* urgent.  This potentially pulls massive numbers of axioms from imports into base, including axiom injections (axioms that change the definition of base entities).  This goes against the whole rationale of having a base.

More minor concern -  language used to describe the changes in doc above and in the makefile options:

```make
    release_use_reasoner : bool = True
    """If set to True, the reasoner will be used during the release process."""
```

Use the reasoner is ambiguous.  There are many potential uses of a reasoner.  I think this means - assert inferred subClassOf axioms between named classes.  If so, text associated with the option should make that clear (probably var name should too)

> Now, it is defined as "the ontology with all non-redundant base axioms". 

We need clearer doc than this.  I don't think this communicates the nature of the change - the old base reflects the structure of the editors file with no instantiation of axioms inferred by a reasoner.  The new base reflects the structure of the ontology release file, including (a non-redundant set of) whatever instantiated reasoner-inferred axioms the authors of the ontology have chose to include in the release.  This may include inferred subClassOf axioms between named classes and materialised existential restrictions.

template/src/ontology/Makefile.jinja2
dosumis: Are the materialise and reason options on the full release and base generation coupled (via the same var) or can they be independently set?  I think we should do the former given that the aim of this new strategy is to have base reflect the release.

matentzn:
> From discussion on call, I believe base-plus moved to using axiom tagging. If so, could this be made clear? If the interests of progress, we could decide not to make this a blocker, but fixing is very urgent. This potentially pulls massive numbers of axioms from imports into base, including axiom injections (axioms that change the definition of base entities). This goes against the whole rationale of having a base.

I forgot to update the text in here, we had already made some progress on this during our last call:

> "base" axioms are axioms that use at least one "base" entity. This is the current definition in ROBOT [pr](https://github.com/ontodev/robot/pull/570). We have added issue #819 - but we have been using this process for a while now without any issues. During our last Monarch Ontology Tools call, we have reviewed the functionality briefly and determined that it "looks correct".

I have documented the next step for dealing with this (and a vote) in this ticket here: https://github.com/INCATools/ontology-development-kit/issues/819

odk/odk.py
```diff
@@ -567,10 +570,16 @@ class OntologyProject(JsonSchemaMixin):
...
     release_artefacts : List[str] = field(default_factory=lambda: ['full', 'base'])
-    """A list of release artefacts you wish to be exported."""
+    """A list of release artefacts you wish to be exported. Supported: base, full, baselite, simple, non-classified, 
+    simple-non-classified, basic."""
     
     release_use_reasoner : bool = True
-    """If set to True, the reasoner will be used during the release process."""
+    """If set to True, the reasoner will be used during the release process. The reasoner is used for three operations:
+    reason (the classification/subclassOf hierarchy computaton); materialize (the materialisation of simple existential/
+    object property restrictions); reduce (the removal of redundant subclassOf axioms)."""
```
matentzn, line 579: From @dosumis:

> More minor concern - language used to describe the changes in doc above and in the makefile options:
    release_use_reasoner : bool = True
    """If set to True, the reasoner will be used during the release process."""
> Use the reasoner is ambiguous. There are many potential uses of a reasoner. I think this means - assert inferred subClassOf axioms between named classes. If so, text associated with the option should make that clear (probably var name should too)

@matentzn has updated the description to be clear that using the reasoner means: reason, materialize and reduce.

dosumis (approved):
I'm now satisfied with the description of the new behaviour and of (roughly) how this is achieved. It would be good to get a more detailed review of the series of robot commands from others.

cmungall (approved):

PR #817, opened 2023-03-01
Add SPARQL to find uses of deprecated DC

cthoyt:
This is a follow-up to https://github.com/oborel/obo-relations/pull/692


anitacaron (changes requested):
Actually, we moved all sparql queries to the jinja template. You can add after [this line](https://github.com/INCATools/ontology-development-kit/blob/85703fa39c33a7e39448edb7fc2b79f5a48dacaa/template/_dynamic_files.jinja2#L613) following the pattern from [line 602](https://github.com/INCATools/ontology-development-kit/blob/85703fa39c33a7e39448edb7fc2b79f5a48dacaa/template/_dynamic_files.jinja2#L602)

anitacaron:
I added this QC as default.

matentzn (commented):
I think this check may be a bit to extreme for a default check. This will break a lot of ontologies out there.. 

At the very least, the check needs to be constrained to classes from the base namespace, but even then, I would feel safer if, instead of messing with the entire namespace at once, we enumerate the specific properties we want to standardise explicitly, like creator, contributor, license and description. What do you think?

cthoyt:
@matentzn fair enough, but it's also the case that this would only apply to anyone creating a new ontology from now, right? And it's reasonable to ask people starting now to follow higher standards.

Though, I do agree that we should slice the check to only apply to terms from the current ontology. Ontology developers aren't typically proactively contributing back to improve upstream ontologies they import. But, I'm not really sure if the impact of choosing one way or another is a really big deal. From the ontologies I've upgraded from DCE to DCTERMS, there have been only a small number of fixes necessary, and never from imported triples

template/_dynamic_files.jinja2
```diff
@@ -641,6 +641,20 @@ SELECT DISTINCT ?term ?property ?value WHERE
...
+^^^ src/sparql/dc-properties-violation.sparql
+# The purpose of this violation is to make sure people update
+# from using the deprecated DC Elements 1.1 namespace (http://purl.org/dc/elements/1.1/)
+# to using the recommended DC Terms namespace (http://purl.org/dc/terms/)
+# See also discussion on https://github.com/oborel/obo-relations/pull/692
+
+SELECT ?term ?predicate WHERE {
+  ?term ?predicate ?value .
+  FILTER(STRSTARTS(STR(?predicate), "http://purl.org/dc/elements/1.1/"))
+  FILTER(isIRI(?term) && ({% if project.namespaces %}{% for ns in project.namespaces %}STRSTARTS(str(?term), "{{ ns }}"){% if not loop.last %} || {% endif %}{% endfor %}{% else %}STRSTARTS(str(?term), "{{ project.uribase }}/{{ project.id.upper() }}_"){% endif %}))
```
matentzn, line 654: 😱 `if not loop.last` !!!!!!! coooooool. Never new that existed!!

matentzn (approved):
I assume you tested this - it looks good to me but I didn't try it!

template/_dynamic_files.jinja2
anitacaron: You said the same the first time I did that 😅

anitacaron: and yes, I tested it. 🙂

anitacaron (approved):

cthoyt:
@matentzn @anitacaron this is cool, but like many things in this ecosystem, I think it has become so complicated that nobody except you two can make a contribution :/

matentzn:
We are working on a way to replacing all of the QC system with a much more simple approach. We are even thinking about doing this soon. For now, you will have to collaborate with is if you want to know how to get these kinds of fixes in.. This PR does not even seem that bad. What specifically fo you find complicated?

Issue #809, opened 2023-02-25
SSSOM test fails because of undeclared prefixes in RO import


gouttegd:
The test suite currently fails at the `test-sssom.yaml` test:

```
sssom parse tmp/modmini-ontc.obographs.json -I obographs-json  -o ../mappings/modmini-ontc.sssom.tsv
WARNING:root:Wikipedia is used in the SSSOM mapping set but it does not exist in the prefix map
WARNING:root:Reactome is used in the SSSOM mapping set but it does not exist in the prefix map
Traceback (most recent call last):
  File "/usr/local/bin/sssom", line 8, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.10/dist-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.10/dist-packages/click/core.py", line 1657, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python3.10/dist-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python3.10/dist-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/sssom/cli.py", line 210, in parse
    parse_file(
  File "/usr/local/lib/python3.10/dist-packages/sssom/io.py", line 112, in parse_file
    doc.clean_prefix_map()
  File "/usr/local/lib/python3.10/dist-packages/sssom/util.py", line 170, in clean_prefix_map
    raise ValueError(
ValueError: ['Wikipedia', 'Reactome'] are used in the SSSOM mapping set but it does not exist in the prefix map
make: *** [Makefile:456: ../mappings/modmini-ontc.sssom.tsv] Error 1
```

This is because the _ro_ import module contains cross-references to Wikipedia and Reactome. Those cross-references end up in the JSON version of the test ontology from which `sssom` later tries to extract the `modmini-ontc.sssom.tsv` mapping set. `sssom parse` does not know what to do with the `Wikipedia:` and `Reactome:` prefixes and errors out as shown above.

Possible solutions:

[...]

PR #822, opened 2023-03-07
Fix test ontologies in ODK

### LINKED ISSUE #821: Test case imports should all be two fixed files in ODK
Step 1: all mirror dependencies should be mirror_from to raw URLs in ODK repo
Step 2: maybe in the future consider a local file solution.

matentzn:
Fixes #821 #809 

As discussed in the above issues, we decided to fix the test suite to specific ontologies which I have downloaded now. There is one case I could not fix, but I made a comment in that specific config. This pertains to a feature that would require a bit of hustle to implement, which I don't think is worth it right now (basically, `use_base` checks for a specific BASE purl, and downloads it from there, you cant change the URL location without, well, not - testing that feature)


gouttegd (approved):
A small step for mankind, a giant leap towards a test suite that we can actually rely on. Thank you!

matentzn:
Haha thanks.!

PR #829, opened 2023-03-18
Do not create individual import modules when use_base_merging is enabled

### LINKED ISSUE #828: Seeding creates individual import modules even when `use_base_merging` is used
When a repository is initialized (or updated with `update_repo`), the seeding process automatically creates a `src/ontology/imports/{imp.id}_import.owl` file (containing a boilerplate RDF/XML header) for each _id_ import declared in the ODK configuration file.

Problem is that it does that even when `use_base_merging` is _True_, even though in that case those individual import modules are actually never needed (what’s needed instead is a single `src/ontology/imports/merged-import.owl` module).

The `_dynamic_files.jinja2` template should be updated to take into account the `use_base_merging` parameter when creating the placeholder import modules, so that it creates only the `merged-import.owl` module when `use_base_merging` is _True_.

gouttegd:
When `use_base_merging` is enabled, the seeding process should only create a single `merged_import.owl` module, instead of one import module for each declared import.

Likewise, the auto-generated -edit file should contain a single import instruction for the single merged module.

closes #828

### DIFF
--- a/template/_dynamic_files.jinja2
+++ b/template/_dynamic_files.jinja2
@@ -23,9 +23,13 @@
 {% if project.edit_format == "obo" %}
 format-version: 1.2
 {% if project.import_group is defined -%}
+{% if project.import_group.use_base_merging %}
+import: {{ project.uribase }}/{% if project.uribase_suffix is not none %}{{ project.uribase_suffix }}{% else %}{{ project.id }}{% endif %}/imports/merged_import.owl
+{% else %}
 {% for imp in project.import_group.products %}
 import: {{ project.uribase }}/{% if project.uribase_suffix is not none %}{{ project.uribase_suffix }}{% else %}{{ project.id }}{% endif %}/imports/{{ imp.id }}_import.owl
 {% endfor %}
+{% endif %}
 {% endif %}
 {%- if project.use_dosdps %}
 import: {{ project.uribase }}/{% if project.uribase_suffix is not none %}{{ project.uribase_suffix }}{% else %}{{ project.id }}{% endif %}/patterns/definitions.owl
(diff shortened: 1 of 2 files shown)


template/src/ontology/catalog-v001.xml.jinja2
```diff
@@ -2,10 +2,15 @@
 <catalog prefer="public" xmlns="urn:oasis:names:tc:entity:xmlns:xml:catalog">
 <group id="Folder Repository, directory=, recursive=false, Auto-Update=false, version=2" prefer="public" xml:base="">
 {% if project.import_group is defined %}
+  {% if project.import_group.use_base_merging %}
+  <uri name="{{project.uribase}}/{% if project.uribase_suffix is not none %}{{ project.uribase_suffix }}{% else %}{{ project.id }}{% endif %}/imports/merged_import.owl" uri="imports/merged_import.owl"/>
+  <uri name="{{project.uribase}}/{% if project.uribase_suffix is not none %}{{ project.uribase_suffix }}{% else %}{{ project.id }}{% endif %}/imports/merged_import.obo" uri="imports/merged_import.obo"/>
```
matentzn, line 7: obo omports not needed, nowhere. I think this can be scrapped ones and for all also from the `else` part.

template/_dynamic_files.jinja2
```diff
@@ -180,6 +188,28 @@ imports/%_import.owl: mirror/%.owl
 	echo "ERROR: You have configured your default module type to be custom; this behavior needs to be overwritten in {{ project.id }}.Makefile!" && touch $@
 {% endif %}
 {%- endif %}
+{% if project.import_group.use_base_merging %}
```
matentzn, line 191: This reveals another mistakes, which is that the initial ODK sets up imports in RDFXML format, while the default format for the respective ROBOT pipeline is functional syntax..

matentzn (commented):
Not really related to your PR, but two comments that reveal mild general inconsistencies of this part of the makefile.

template/src/ontology/catalog-v001.xml.jinja2
gouttegd: Yes, I noticed that. I decided to keep the OBO entry for consistency with the individual imports, but happy to go the other way and instead remove all OBO catalog entries for OBO imports.

template/_dynamic_files.jinja2
gouttegd: It does not really matter, though, does it? The placeholder import module will be replaced by a ROBOT-generated module in the expected functional syntax.

matentzn: If you are a fetishist of git history cleanliness, then the very first commit (which is automatically done by the ODK) will be an RDF XML file, and the second one a functional syntax file. Its not a big deal, just saying..

matentzn: (In any case, I am happy with this PR, feel free to merge and either make follow up issues on the OBO and syntax stuff, or leave it!)

gouttegd: You know my position about how a PR should be about only one problem at a time :) so I’ll merge this one now and make a new one to fix the existing inconsistencies.

PR #830, opened 2023-03-20
Fix placeholder import modules

gouttegd:
This PR fixes two small inconsistencies in the seeding process that were found during the review of #829.

* The automatically generated placeholder import modules should be in the same format as the one used by the import pipeline (i.e. OWL Functional Syntax rather than RDF/XML).
* There’s no need for catalog entries for OBO variants of import modules, since we never generate OBO-formatted import modules anyway.

### DIFF
--- a/template/_dynamic_files.jinja2
+++ b/template/_dynamic_files.jinja2
@@ -215,7 +215,7 @@ Ontology(<{{ project.uribase }}/{% if project.uribase_suffix is not none %}{{ pr
 #}
 {% for imp in project.import_group.products %}
 ^^^ src/ontology/imports/{{ imp.id }}_import.owl
-Prefix(:=<{{ project.uribase }}/{% if project.uribase_suffix is not none %}{{ project.uribase_suffix }}{% else %}{{ project.id }}{% endif %}/imports/merged_import.owl>)
+Prefix(:=<{{ project.uribase }}/{% if project.uribase_suffix is not none %}{{ project.uribase_suffix }}{% else %}{{ project.id }}{% endif %}/imports/{{ imp.id }}_import.owl>)
 Prefix(obo:=<http://purl.obolibrary.org/obo/>)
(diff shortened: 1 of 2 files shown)


template/_dynamic_files.jinja2
```diff
@@ -217,20 +215,18 @@ imports/%_import.owl: mirror/%.owl
...
-     xml:base="{{ project.uribase }}/"
-     xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
-     xmlns:owl="http://www.w3.org/2002/07/owl#"
-     xmlns:oboInOwl="http://www.geneontology.org/formats/oboInOwl#"
-     xmlns:xml="http://www.w3.org/XML/1998/namespace"
-     xmlns:xsd="http://www.w3.org/2001/XMLSchema#"
-     xmlns:rdfs="http://www.w3.org/2000/01/rdf-schema#"
-     xmlns:obo="http://purl.obolibrary.org/obo/">
-    <owl:Ontology rdf:about="{{ project.uribase }}/{% if project.uribase_suffix is not none %}{{ project.uribase_suffix }}{% else %}{{ project.id }}{% endif %}/imports/{{ imp.id }}_import.owl"/>
+Prefix(:=<{{ project.uribase }}/{% if project.uribase_suffix is not none %}{{ project.uribase_suffix }}{% else %}{{ project.id }}{% endif %}/imports/merged_import.owl>)
```
matentzn, line 218: What does this translate to in the normal case? Example?

matentzn: It should be: 

```
http://purl.obolibrary.org/obo/{imp.id}/imports/somimport_import.owl.
```

matentzn: Yes! If this is the case, than its fine!

gouttegd: It is now. :)

PR #833, opened 2023-03-20
Update `constraints.txt` fixes #831

### LINKED ISSUE #831: Make sure the latest semsql is installed by `update-constraints.sh`
Here is the action:
https://github.com/INCATools/ontology-development-kit/blob/master/.github/workflows/constraints.yml#L18

And what it runs:
https://github.com/INCATools/ontology-development-kit/blob/master/Makefile#L170

Right now we end up with an old semsql version whenever we refresh this!

hrshdhgd:
fixes #831

### DIFF
--- a/constraints.txt
+++ b/constraints.txt
@@ -18,7 +18,7 @@ beautifulsoup4==4.12.0
 bidict==0.22.1
 bioregistry==0.6.99
 bleach==6.0.0
-bmt==1.0.3
+bmt==1.0.4
 cachetools==5.3.0
 certifi==2022.12.7
 cffi==1.15.1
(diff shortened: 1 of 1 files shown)


matentzn (approved):

gouttegd:
That’s not how it is supposed to work…

The `constraints.txt` should _never_ be updated manually; it’s strictly the output of PIP’s dependency resolution algorithm.

If there’s a Python package for which you would like to enforce some specific constraints, that should be done in the `requirements.txt.full` file, e.g.:

```
semsql >= 0.3
```

the constraint will then be picked up by the `update-constraint` action next time it is run (it can be run manually if there’s an urgent need to to create an ODK image with the new semql version), which will then update the `constraints.txt` file, taking into account the specific constraint on the package.

gouttegd:
My bad.

But in that case it’s strange that the previous execution of the action earlier today didn’t update semsql, but the manual invocation a few hours later did…

matentzn:
Indeed. I think what happened is that an OAK release today dealt with a bad dependency; not sure though

PR #846, opened 2023-04-11
Add safety check for constraints update

### LINKED ISSUE #845: Generating constraints.txt should fail if no constraints.txt is generated
in this run #844 some ridiculous error happened:

https://github.com/INCATools/ontology-development-kit/actions/runs/4657721662/jobs/8242594679#step:5:880

The PR should not have gone ahead. At least, we should add a `test` statement to the shell script?

gouttegd:
This PR

* updates the `update-constraints.sh` script to ensure it returns a non-zero code upon any failure condition (including the case where all commands execute successfully but the generated `constraints.txt` file is empty);
* updates the `constraints.yml` workflow to prevent it from creating a PR if the `update-constraints.sh` script returns a non-zero code.

closes #845

### DIFF
--- a/.github/workflows/constraints.yml
+++ b/.github/workflows/constraints.yml
@@ -18,6 +18,7 @@ jobs:
         run: sh update-constraints.sh
       - name: Create Pull Request
         uses: peter-evans/create-pull-request@v3
+        if: ${{ success() }}
         with:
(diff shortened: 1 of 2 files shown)


matentzn (approved):
Excellent thank you!

Issue #848, opened 2023-04-14
mkdocs auto deploy doesnot work.


vyasakhilesh:
While running below command from the host machine
```bash
sh run.sh make update_docs
````
Getting this error 
```bash
mkdocs gh-deploy --config-file ../../mkdocs.yaml
INFO     -  Cleaning site directory
INFO     -  Building documentation to directory: /work/site
WARNING  -  A relative path to 'odk-workflows/ManageAutomatedTest.md' is included in the 'nav' configuration,
            which is not found in the documentation files
WARNING  -  Documentation file 'odk-workflows/index.md' contains a link to
            'odk-workflows/ManageAutomatedTest.md' which is not found in the documentation files.
INFO     -  Documentation built in 0.35 seconds
INFO     -  Copying '/work/site' to 'gh-pages' branch and pushing to GitHub.
The authenticity of host 'github.com (::ffff:140.82.121.4)' can't be established.
ED25519 key fingerprint is SHA256:+DiY3wvvV6TuJJhbpZisF/zLDA0zPMSvHdkr4UvCOqU.
This key is not known by any other names
Are you sure you want to continue connecting (yes/no/[fingerprint])? yes
Warning: Permanently added 'github.com' (ED25519) to the list of known hosts.
Connection reset by 140.82.121.4 port 22
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
Traceback (most recent call last):
  File "/usr/local/bin/mkdocs", line 8, in <module>
    sys.exit(cli())
  File "/usr/local/lib/python3.10/dist-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.10/dist-packages/click/core.py", line 1657, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python3.10/dist-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python3.10/dist-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/mkdocs/__main__.py", line 280, in gh_deploy_command
    gh_deploy.gh_deploy(
[...]

PR #853, opened 2023-04-17
Forward the host SSH agent socket into the container

### LINKED ISSUE #852: Make possible to use user’s SSH keys from within the ODK container
Cool people don’t use Git with passwords, they use it with a SSH key pair instead. For those people, it should be possible to use Git commands that require a SSH key from within the container.

Assuming most people using a SSH key pair are also using a SSH agent, all we would have to do is to forward the agent’s socket into the container:

```
$ docker run -v $SSH_AUTH_SOCK:/ssh-auth.sock -e SSH_AUTH_SOCK=/ssh-auth.sock [other docker run options and arguments…]
```

This works fine at least on GNU/Linux. Unfortunately it doesn’t work on macOS, where [a special workaround](https://github.com/docker/for-mac/issues/410) exists instead:

```
$ docker run -v /run/host-services/ssh-auth.sock:/run/host-services/ssh-auth.sock -e SSH_AUTH_SOCK=/run/host-services/ssh-auth.sock [other docker run options and arguments…]
```

A simpler option is to just bind the user’s `~/.ssh` directory, but this will only work for simple setups where an agent is not necessary – and many users may not like the idea of exposing their SSH private key to the ODK container, not matter how much they trust the ODK…

gouttegd:
This PR:

* binds the SSH agent socket (as found in `$SSH_AUTH_SOCK`) from the host to a predictable location inside the container;
* sets the `SSH_AUTH_SOCK` variable in the container to point to said predictable location.

This allows processes running inside the container to communicate with the host's SSH agent and therefore to use the user's SSH keys without having to expose the keys directly to the container.

closes #852


matentzn:
Thanks @gouttegd I will look at this after biocuration unless there is a reason to rush?

gouttegd:
No emergency as far as I am concerned. This may solve #848, if the user here is using a SSH agent (which they probably do).

gouttegd:
Ping @matentzn. This is very simple change that does not alter the behaviour of the ODK, except that it makes it possible to use SSH with agent-provided keys from within the container.

matentzn (approved):
Lets merge and try, I can understand the general outline but without testing, no way for me to provide a proper review!

PR #871, opened 2023-05-15
Fix build on M1 architecture.

gouttegd:
It seems that in recent versions of Docker, the `--platform` option of `docker build` no longer accepts a single architecture name as its argument (as in `--platform arm64`). Instead, a tuple specifying both the operating system and the architecture (e.g. `linux/arm64`) should be used.

Using `--platform arm64` now causes Docker to fail building the ODK on arm64 Apple machines because it can’t find the arm64 variant of Ubuntu (our base image) anymore.

So this PR updates all the Makefiles to make sure they only use the new `os/arch` syntax.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,7 @@ EMAIL_ARGS=
 
 CACHE=
 
-ARCH=$(shell uname -m | sed 's/x86_64/amd64/')
+ARCH=linux/$(shell uname -m | sed 's/x86_64/amd64/')
 PLATFORMS=linux/amd64,linux/arm64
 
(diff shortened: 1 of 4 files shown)


matentzn:
I don't know anything about these things, but why is it not necessary to specifically add something like mac/arm64?

gouttegd:
The OS part of the platform refers to the system we run _in_ the image, not the system the image runs _on_. So for the ODK, it’s always `linux`. :)

gouttegd:
I admit it is confusing, because the architecture part does refer to the architecture the image is supposed to run on!

PR #873, opened 2023-05-23
Automatically check whether the repository needs to be updated

### LINKED ISSUE #856: Warn when `update_repo` should be run
Possible improvement: The ODK should warn users when the `$(ONT)-odk.yaml` file has been modified but `make update_repo` has _not_ been run.

One way this could work: When generating the standard Makefile, we compute a hash of the `$(ONT)-odk.yaml` file, and we store it in the Makefile. Then we add a rule in the Makefile that also computes a hash of the config file, and compares it with the stored hash. If the hashes are different, we emit a warning (“your ODK configuration has changed since the last time this Makefile was generated; run update_repo to make sure your changes take effect”). And we make that rule the first dependency of all other rules, so that the check is always performed.

gouttegd:
This PR implements the behaviour suggested in #856.

That is, when a repository is initialised or updated, the ODK configuration file is hashed and the hash value is baked in the generated Makefile. A new `config_check` rule (called as a dependency to the `all_odk` rule) then compares the stored hash with a freshly computed hash on the current configuration file, and prints a message warning the user that their repository is out-of-date if the two hashes do not match.

closes #856

### DIFF
--- a/odk/odk.py
+++ b/odk/odk.py
@@ -22,6 +22,7 @@ import subprocess
 import shutil
 from shutil import copy, copymode
 import logging
+from hashlib import sha256
 
 logging.basicConfig(level=logging.INFO)
 TEMPLATE_SUFFIX = '.jinja2'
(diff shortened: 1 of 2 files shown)


matentzn (commented):
From just reading this it is super amazing; I am missing a bit the config_hash parameter and default value in the dataclass https://github.com/INCATools/ontology-development-kit/blob/master/odk/odk.py#L452,

This is also used to generate the documentation

gouttegd:
`config_hash` is _not_ a parameter. It is not supposed to be set by the user. Its value is automatically generated by the seeding script. I don’t see the point of setting a “default value” in the data class – a default value would make no sense, it would always be overwritten by the seeding script.

Likewise I don’t think it’s useful for this variable to be documented. Users should not have to care about it. They don’t even have to know it exists. It’s purely an internal variable to make the up-to-date check work.

matentzn (approved):
NOW I get it! genius! Sorry I didnt read it correctly, I missed this row:

https://github.com/INCATools/ontology-development-kit/pull/873/files#diff-ee525c96b3878b3e521023193d220a8e146ff4d1fe8f9348b75690d97af54d8fR723

And got confused.

Super!

PR #880, opened 2023-06-04
Default component goal should be restricted to owl

matentzn:
As it turns out, having a general goal for files in the components directory was a mistake for various reasons (see https://github.com/monarch-initiative/mondo-ingest/pull/299; in a nutshell, you cant overwrite a wildcard goal in make with another more specific wildcard goal - only with a concrete goal without wildcards). 

This PR should:

- Check of a component with the file extension `.owl` exists (i.e. other file formats are unaffected).
- Only if it does not exist, create an empty file (this way, the makegoal also does not needlessly cause all downstream goals to update)

cc @joeflack4

### DIFF
--- a/template/src/ontology/Makefile.jinja2
+++ b/template/src/ontology/Makefile.jinja2
@@ -538,9 +538,9 @@ recreate-%:
 no-mirror-recreate-%:
 	$(MAKE) COMP=true IMP=false IMP_LARGE=false MIR=false PAT=true $(COMPONENTSDIR)/$*.owl -B
 
-$(COMPONENTSDIR)/%: | $(COMPONENTSDIR)
+$(COMPONENTSDIR)/%.owl: | $(COMPONENTSDIR)
 	touch $@
-.PRECIOUS: $(COMPONENTSDIR)/%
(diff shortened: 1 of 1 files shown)


gouttegd (approved):

joeflack4:
Thanks for your attention to this @matentzn ! Were you able to test it out? Assuming it fixes the issue, is this something I can fix with `update_repo`, or do I have to wait for the next ODK release or?

gouttegd:
> is this something I can fix with update_repo, or do I have to wait for the next ODK release

Both! You will need to run `update_repo` to get the fix, but first you need to wait until the fix has made it into a release. For now the fix is only in the `master` branch, running `update_repo` would have no effect.

The `update_repo` command does _not_ get the latest stuff straight from the ODK repository. It only updates your own repository against the latest ODK image that is locally available in your Docker setup.

matentzn:
The fix is in :dev now: https://hub.docker.com/r/obolibrary/odkfull/tags!

@gouttegd thank your for your explanations. In the Repo @joeflack4 and I work on (not a real ontology pipeline), we use `dev` image for everything because we constantly need the latest tools, makefile fixes etc.

PR #882, opened 2023-06-06
Clean up uses of catalog file in Makefile

matentzn:
- replaces hardcoded uses of catalog file with variable
- removes unnecessary dependency of sparql_test command on catalog file (literally all goals depend on the catalog file, no need to have it in one)
- removes redundant --catalog declaration in verify command which already has it in the ROBOT variable.


anitacaron:
I've tested the commands where they use the `catalog-v001.xml`, and it seems to continue working as expected.

matentzn:
Thank you @anitacaron 

The reason this PR failed is a bit disconcerting.. Seems the PROLOG libs we depend on where 503. I will try again tomorrow!

anitacaron:
Yes, I noticed. That's why I didn't approve the PR yet.

anitacaron (approved):

PR #886, opened 2023-06-15
Add Perl module Business::ISBN.

### LINKED ISSUE #885: Request for a new Perl package in the ODK image
FlyBase ontologies are using for some of their quality control checks, among other things, a Perl script that requires the `Business::ISBN` module (to validate `isbn:` cross-references). We install manually prior to running the script.

Any objection to the idea of adding that Perl module directly to the ODK? (We already have a full Perl distribution there, so one more module is not adding much.)

gouttegd:
This PR adds the `Business::ISBN` Perl module to the system-wide Perl distribution.

closes #885

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,6 +30,7 @@ RUN apt-get update && DEBIAN_FRONTEND="noninteractive" apt-get install -y --no-i
     dos2unix \
     sqlite3 \
     libjson-perl \
+    libbusiness-isbn-perl \
     pkg-config \
     xlsx2csv \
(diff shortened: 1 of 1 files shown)


matentzn (approved):
No problem IMO!

gouttegd:
Thanks!

(Remember when I said how my PR #769 should not cause any problem to anyone, except to those who are using `apt` to install custom packages as part of their workflows – something that will no longer possible with that PR? When, I had forgotten that _I_ was one of those people… :D )

PR #889, opened 2023-06-22
Fix bug that docs workflow is created during seeding

matentzn:


matentzn:
@anitacaron did you mean to approve this?

anitacaron:
The problem is the `docs` GitHub action that is created [as default](https://github.com/INCATools/ontology-development-kit/blob/master/template/_dynamic_files.jinja2#L1769) and the `mkdocs.yml` file is created with `documentation` option.

anitacaron (changes requested):
Need to change the code in line 1769 or change the default value.

```
{% if (('github_actions' in project.ci) and ('docs' in project.workflows)) %}
```

anitacaron:
I can update the PR with your decision.

anitacaron:
The default workflow should be 'qc' instead of 'docs'.

anitacaron:
Or no default at all.

matentzn:
Ok, I will let you decide how to move this forward :) Thank you!

anitacaron (approved):

PR #890, opened 2023-06-27
Run the test suite when new commits are added to a PR.

gouttegd:
Update the configuration of the main CI workflow to trigger a CI check not only when a PR is initially opened, but also when new commits are pushed to it ("synchronize" event).

The reason for not triggering new CI checks when a PR is updated was to avoid too many runs of the test suite, but this was before we added a path-based ignore rule that prevents the test suite from being run on changes that do not affect any running code. With such a rule in place, avoiding automatically triggered CI checks is no longer warranted.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -6,6 +6,7 @@ on:
       - master
     types:
       - opened
+      - synchronize
     paths-ignore:
       - '**.md'
       - 'CITATION.cff'


gouttegd:
Test suite run explicitly cancelled since the change in `build-and-test.yml` does not touch the test suite itself – only _when_ it is triggered. This is not a change affecting anything the CI check can test.

gouttegd:
See [this comment](https://github.com/INCATools/ontology-development-kit/pull/769#issuecomment-1609860127) for more details on the rationale for re-enabling automatic triggering of the test suite.

.github/workflows/build-and-test.yml
```diff
@@ -6,6 +6,7 @@ on:
       - master
     types:
       - opened
+      - synchronize
```
matentzn, line 9:
```suggestion
      - synchronize
      - reopened
```

Which is the same as leaving it blank:
https://frontside.com/blog/2020-05-26-github-actions-pull_request/#how-do-workflows-trigger-on-pull_request

gouttegd: Not sure I see the point of re-running the test suite upon re-opening a closed PR…

If you re-open the test suite without having pushed any new commits, then the last CI check performed on the last pushed commit before the PR was closed is still valid; and after re-opening, if you change the PR by pushing new commits, the `synchronize` setting will trigger a new run.

The only thing that running the test suite immediately upon re-opening the PR would test for is the following scenario:

* A PR is submitted while the master branch is in state _t0_. Let’s say the CI checks pass.
* The PR is closed for whatever reasons.
* Other, unrelated changes happen on the master branch, which is now in state _t1_.
* Later, the PR is re-openened. The PR still applies cleanly against master’s _t1_ state, but somehow, because of the changes between _t0_ and _t1_, the PR now fails the CI checks.

This is no different from the sitation where a PR passes the CI checks at a given moment, but then is left (without being closed) without being merged for a while, and then is finally approved and merged – by the time the PR is merged, maybe the changes that happened in the meantime on the master branch would actually cause the PR to fail, but we do not test that.

matentzn: Ok! Lets follow your lead!

matentzn (approved):

PR #905, opened 2023-08-02
Run under root by default unless ODK_USER_ID is explicitly set.

gouttegd:
This PR changes the entry point script so that it defaults to run the inner processes as root, unless the `ODK_USER_ID` variable is explicitly set to something else than zero.

This avoids breaking things for users who for some reasons do not pass the `ODK_USER_ID` variable when they invoke the ODK – which includes Windows users, since the `run.bat` script has never been updated to pass that variable.


matentzn (approved):
THANK YOU!

I will create a development snapshot right away

PR #918, opened 2023-09-01
Update 1.4.2 maintaince release

matentzn:
This PR:

- updates to latest constraints.txt to catch the last sssom releases
- Adds the libbusiness-isbn-perl package
- Updates to using the variables for versions of tools (previously, versions were hardcoded) in Dockerfiles
- Updates Changelog to reflect latest changes in constraints.txt

@gouttegd can you check if this is all ok (note that this PR is set to merge into the 1.4 maintenance PR).


gouttegd:
Note that the test suite does not run by default on branches other than the main branch. You need to trigger it manually if you want it to run.

matentzn:
I am running tests locally! At all times.

PR #919, opened 2023-09-02
Execute ammonite with java explicitly

matentzn:
This is not necessary when using the normal local build, but it is when running multiarch builds.

Should we merge this to master as well?


gouttegd (approved):
There should be nothing wrong with that.

However, I would still like to understand how you are even able to build a multi-arch image while your system does _not_ have a arm64 builder (referring to our discussion on Slack), because I suspect the answer might explain why this change here is necessary…

PR #920, opened 2023-09-03
Add v1.4.1 and v1.4.2 release information to Changelog

matentzn:
@gouttegd shouldnt this information be in the changelog, or is all the relevant information only kept under the v.1.4 section?


gouttegd:
Up to you. I personally think it’s clearer if the maintenance branch (and the versions derived from it) is kept completely out of the main branch: nothing goes from the maintenance branch to the main branch (not even the release notes), only the other way around (from master to maintenance).

matentzn:
I trust your judgement here!

PR #927, opened 2023-09-26
Remove APs from merged import

matentzn:
adds a feature to the merged import pipeline that strips away any non-declared APs from the merged import.


matentzn:
@anitacaron can you carefully think this through? It looks sane, but maybe I have not thought of all the implications

template/src/ontology/Makefile.jinja2
```diff
@@ -448,6 +448,7 @@ $(IMPORTDIR)/merged_import.owl: $(MIRRORDIR)/merged.owl $(IMPORTDIR)/merged_term
 	if [ $(IMP) = true ]; then $(ROBOT) merge -i $< \{% if project.import_group.exclude_iri_patterns is not none %}
 		{% for pattern in project.import_group.exclude_iri_patterns %}remove  --select "{{ pattern }}" {% endfor %} \{% endif %}
 		extract -T $(IMPORTDIR)/merged_terms_combined.txt --force true --copy-ontology-annotations true --individuals {{ project.import_group.slme_individuals }} --method {{ project.import_group.module_type_slme }} \
+		remove $(patsubst %, --term %, $(ANNOTATION_PROPERTIES)) -T $(IMPORTDIR)/merged_terms_combined.txt --select complement \
```
matentzn, line 451:
```suggestion
		remove $(patsubst %, --term %, $(ANNOTATION_PROPERTIES)) -T $(IMPORTDIR)/merged_terms_combined.txt --select complement --select annotation-properties \
```

matentzn: Even this does not work correctly. Too much is removed

anitacaron:
This is working, but not sure if it's how you expect. The only annotations in the merged_import.owl are listed in the merged_terms_combined.txt.

anitacaron:
Let me know if this is what you mean and I'll approve it.

matentzn:
Thank you @anitacaron. Indeed, it seems work as intended (even though I am a bit surprised --signature true is not needed, but, oh well)!

Ok, now last question to both of you @anitacaron @gouttegd :

I think this is a good change. This will drastically (drastically) reduce the size of our merged imports on the one side, while making it possible to the user to add any AP they like explicitly. On the other side (actually the same point), it _will remove a lot of annotation assertions_. People might be wondering: why cant I see the comments of this imported class.

image

Can you both provide an approving review if you agree with the tradeoffs and with my judgement of it?

gouttegd:
What’s the behaviour wrt to obsolete terms (since obsoletion is represented by an annotation property)?

Let’s say my ontology (_A_) imports a term B:1234 from ontology _B_. At some point, editors of B decides to obsolete B:1234, and they make a new release where B:1234 is now obsolete.

If `owl:deprecrated` is not explicitly listed among the annotation properties to preserve (which it is not, by default, from what I can tell), what will happen when I refresh the _B_ import in _A_? Will B:1234 correctly appear as obsolete?

anitacaron:
It didn't remove owl annotations, so the obsolete terms remained unchanged.

gouttegd:
OK, I’ll need to test locally, because right now I don’t understand which annotations exactly are supposed to be removed.

gouttegd:
In FBbt, this removes a lot of classes from CL and GO, but they are classes that probably should not have been there to begin with (they are not referenced by anything in FBbt). I am not sure I understand why.

But the resulting import file seems good, though.

gouttegd (approved):

gouttegd:
My bad, made a comparison on the wrong builds. The classes that were removed were due to other changes, not this one.

None of the annotations that are removed would be missed, I think, and if needed, listing the properties to preserve in `annotation_properties` does work as expected.

Looks good to me.

anitacaron (approved):

PR #929, opened 2023-09-26
Add option to include defined-by annotation in imports

### LINKED ISSUE #796: Add rdfs:isDefinedBy to imports
ROBOT now has this cool option with merge to add provenance:

```
robot merge -I http://purl.obolibrary.org/obo/pato.owl -I http://purl.obolibrary.org/obo/ro.owl --annotate-defined-by true -o defined-by.owl
```

We should consider adding this by default to ODK.

anitacaron:
Fixes #796

There're two cases:

1. When `use_base_merging: true`, the option is added when merging all mirrors, so then we can get properly where the terms come from.
2. When `use_base_merging` not defined, the option is added during the release process.


anitacaron:
Can you clarify, @matentzn, please?

matentzn:
This should be configurable at least on import_group level, because the file size will blow up massively with that feature, and we want to at least give some hope that it can be switched off for people for which size matters (github file size limits etc).

Also,  what about the case that you are not using base merging, but want to use this feature when creating the full.owl release? Isnt this another case where we may want this (we could use the same flag from import_group.annotate_defined_by)?

template/src/ontology/Makefile.jinja2
```diff
@@ -915,7 +915,7 @@ $(ONT).owl: $(ONT)-{{ project.primary_release }}.owl
 {% endif -%}
 
 # ROBOT pipeline that merges imports, including components.
-ROBOT_RELEASE_IMPORT_MODE={% if project.use_edit_file_imports -%}$(ROBOT) merge --input $< {% else -%}
+ROBOT_RELEASE_IMPORT_MODE={% if project.use_edit_file_imports -%}$(ROBOT) merge --input $< --annotate-defined-by true {% else -%}
```
anitacaron, line 918: @matentzn 
> Also, what about the case that you are not using base merging, but want to use this feature when creating the full.owl release? Isnt this another case where we may want this (we could use the same flag from import_group.annotate_defined_by)?

I did this in the variable `ROBOT_RELEASE_IMPORT_MODE`.

anitacaron:
Ok, I'll add the configuration for this.

anitacaron:
@matentzn is ready for review again

template/src/ontology/Makefile.jinja2
```diff
@@ -915,7 +915,7 @@ $(ONT).owl: $(ONT)-{{ project.primary_release }}.owl
 {% endif -%}
 
 # ROBOT pipeline that merges imports, including components.
-ROBOT_RELEASE_IMPORT_MODE={% if project.use_edit_file_imports -%}$(ROBOT) merge --input $< {% else -%}
+ROBOT_RELEASE_IMPORT_MODE={% if project.use_edit_file_imports -%}$(ROBOT) merge --input $< {% if project.import_group is defined and project.import_group.use_base_merging is false and project.import_group.annotate_defined_by %}--annotate-defined-by true{% endif -%}{% else -%}
```
matentzn, line 918: Is this conditional needed? `and project.import_group.use_base_merging is false`? seems redundant. If you try to annotate an already annotated class, nothing happens!

matentzn: (but please test this if true) :D

matentzn: Ok! Looks good then

matentzn:
From now on for changes that affect the pipeline, we should have two independent reviews. To minimise mistakes..

gouttegd (approved):

PR #930, opened 2023-09-26
Update yaml to use canonical form for true/false in tests dir

### LINKED ISSUE #674: Add warning to next release to use canonical form for booleans in config files
- [x] Update all test files in ODK repo
- [ ] Add warning to 1.3.2 release changelog

See also https://github.com/OBOAcademy/obook/pull/258

anitacaron:
Fixes #674


matentzn (approved):
If @gouttegd cant think of a reason this is problematic - I cant!

gouttegd (approved):
I won’t start ranting about how JSON/YAML are insane for accepting so many values as booleans in the first place. Happy to see that YAML 1.2 at least has a bit more sense.

PR #932, opened 2023-09-27
Update `illegal-date-violation.sparql` to accept `xsd:dateTime`

anitacaron:
Related to https://github.com/obophenotype/cell-ontology/pull/2066

Also adds the property `oboInOwl:creation_date` to the list of properties to check.


matentzn (approved):
I don't have an issue with this suggestion, but shouldn't we try to get rid of oboInOwl:creationDate?

anitacaron:
We can get rid of the property in another moment, but it's being used in CL.

matentzn:
cool

PR #934, opened 2023-09-27
Make cargo use the "sparse" protocol to download the creates index.

gouttegd:
The default protocol that Cargo is using to download its index of crates is somehow too resource-consuming when the ODK is built under BuildKit (to build multiarch image). So we force Cargo to use the new HTTP-based "sparse" protocol, which is supposedly much more efficient and does not seem to cause issue with BuildKit.

This is the same as #933, but applied to the 1.4 branch.


gouttegd:
@matentzn Given that the 1.4.3 release was built off that branch, I assume you approve of it. :)

PR #939, opened 2023-10-05
Re-enable the LightRDF-based RDF/XML check.

gouttegd:
(This is the same as #928, but targeting the `dev` branch for testing purposes.)

The ODK used to have a validation check that ensured that RDF/XML files were valid. The check was based on LightRDF, whichwas fast enough to make it possible to enable the check by default.

We had to disable that check because of a bug in LightRDF that caused the library to sometimes fail to parse perfectly valid RDF/XML files (#745). That bug has been fixed, so now we can re-enable the check by default.

The check-rdfxml script is now invoked by default, and by default is uses the fast LightRDF-based check. A new ODK option is added (extra_rdfxml_checks): when enabled, it instructs the check-rdfxml script to *also* perform the Jena- and RDFLib-based checks (which are not enabled by default as they are more time-consuming).

### DIFF
--- a/odk/odk.py
+++ b/odk/odk.py
@@ -622,8 +622,11 @@ class OntologyProject(JsonSchemaMixin):
     robot_report : Dict[str, Any] = field(default_factory=lambda: ReportConfig().to_dict())
     """Block that includes settings for ROBOT report, ROBOT verify and additional reports that are generated"""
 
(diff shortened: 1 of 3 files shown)


gouttegd:
@matentzn Merging without review here since you already explicitly approved merging this to the `dev` branch for testing.

PR #940, opened 2023-10-06
Collect per-command resource usage data.

gouttegd:
When running under ODK_DEBUG=yes, instruct `make` to run every command through a small wrapper around the actual shell to capture resource usage (CPU time and memory consumption) for each command, in addition to capturing resource usage for the entire run. The resource usage data is written into `tmp/debug.log`.


gouttegd:
@matentzn @anitacaron I’ve done only limited testing of this for now. I’ll do more when I’m back next week.

matentzn:
This looks great! 

@anitacaron when you have an opportunity can you test this?

PR #958, opened 2023-11-25
Add SSSOM-Java to odkfull.

gouttegd:
This PR adds SSSOM-Java to the full image: both the `sssom-cli` command line tool, and the ROBOT plugin.

The command-line tool is installed so that it is readily reachable from the PATH. The ROBOT plugin is installed in a directory dedicated to ROBOT plugins (`/tools/robot-plugins`), but *without* configuring ROBOT to look for plugins in that directory. That is left to any workflow that require the plugin. This is so workflow engineers can still choose to download a more recent version of the plugin if they want/need.


matentzn (approved):
Thank you!

PR #968, opened 2023-12-11
Deliver ROBOT plugins in ODK

### LINKED ISSUE #909: Deliver ROBOT plugins in ODK - whats the best way?
We need to decide on a way to make ROBOT plugins easily configurable with ODK. 

There are basically two completely different approaches (at least)

1. We pass in a plugins dir to run.sh (as a volume) and require all custom plugins to be present
1. We add custom plugins into the build process. Basically, we depend on them the same way we depend on data artefacts, download them into a _local_ plugin directory in the project, and point `ROBOT_PLUGINS_DIRECTORY` at that. For example:

Makefile contents:

```
export ROBOT_PLUGINS_DIRECTORY=$(PWD)/tmp/plugins

$(ROBOT_PLUGINS_DIRECTORY)/monarch.jar:
	mkdir -p $(ROBOT_PLUGINS_DIRECTORY)
	wget ... -O $@

hp-eq.owl: | $(ROBOT_PLUGINS_DIRECTORY)/monarch.jar
	$(ROBOT) monarch:upheno-augment -i hp-edit.owl --relation has_phenotype_affecting  -o $@
```

While I appreciate the cleanliness of approach 1, it puts too much of a burden on most of the people I work with - I would just like to fold it into the general make process somehow, but I am open to add both alternatives.

Of course you could sort of do a hybrid: pass in the ~/.robot/plugins dir as a volume _and_ download dependencies as necessary. But I would like to keep dependencies on local file structures as minimal as possible.

gouttegd:
This PR proposes a method to manage ROBOT plugins in the ODK. The aims are:

1) allow to easily use any plugin that is already provided by the ODK itself (currently, the ODK provides only the SSSOM plugin, but more plugins may be added in the future), both in the standard ODK-generated Makefile and in a custom Makefile;
2) allow to easily add any extra plugin that is _not_ provided by the ODK;
3) allow to easily _override_ a plugin that is provided by the ODK (for example, if the ODK provides version 0.6.2 of the SSSOM plugin, a ODK user must be able to locally use version 0.7.0 without having to wait for the next ODK release with the latest version).

The core principle is that we define a _runtime plugins directory_, which is `$(TMPDIR)/plugins`. The `$(ROBOT_PLUGINS_DIRECTORY)` variable is defined to point to that directory and exported to the environment, so that ROBOT can find any plugin located in that directory.

Once that principle is posed, the rest is only a matter of ensuring that the plugins end up being placed in that directory. This is done by the `all_robot_plugins` target. Any rule (whether it is a standard rule in the ODK-generated Makefile or a custom rule in the custom Makefile) that requires the use of a plugin should depend on that target, to ensure that the plugins are ready to use.

There are several ways for a user to use extra plugins and/or to override the ODK-provided plugins:

**A) Putting the plugin(s) they want to use in a top-level `plugins` directory.** The default rule to copy plugins into the runtime directory will look into that top-level `plugins` directory in addition to looking into the ODK’s `/tools/robot-plugins` directory, so any plugin found there will be automatically copied as if it had been provided by the ODK. The top-level `plugins` directory takes precedence over the ODK’s `/tools/robot-plugins` directory.

**B) Using custom rules in the custom Makefile.** To override a ODK-provided plugin, one may override the default installation rule with a specific rule. For example, to force the use of version 0.7.0 of the SSSOM plugin (instead of whatever version is provided in the ODK), one may do the following:

```Makefile
$(ROBOT_PLUGINS_DIRECTORY)/sssom.jar:
        curl -L -o $@ https://github.com/gouttegd/sssom-java/releases/download/sssom-java-0.7.0/sssom-robot-plugin-0.7.0.jar
```

To install a _supplementary_ plugin (instead of overriding a default one), one can override the `custom_robot_plugins` target, which is itself a dependency of `all_robot_plugins`. For example, to add the KGCL plugin:

```Makefile
$(ROBOT_PLUGINS_DIRECTORY)/kgcl.jar:
	curl -L -o $@ https://github.com/gouttegd/kgcl-java/releases/download/kgcl-0.2.0/kgcl-robot-plugin-0.2.0.jar

custom_robot_plugins: $(ROBOT_PLUGINS_DIRECTORY)/kgcl.jar
```

**C) Explicitly listing the supplementary plugins in the ODK configuration.** This PR adds a new possible configuration section in the ODK config file, allowing to list extra plugins to install. For example, another way of adding the KGCL plugin would be to amend the ODK configuration as follows:

```yaml
robot_plugins:
  plugins:
    - name: kgcl
      mirror_from: https://github.com/gouttegd/kgcl-java/releases/download/kgcl-0.2.0/kgcl-robot-plugin-0.2.0.jar
```

The generated Makefile will then contain the required code to automatically download that plugin and place it into the runtime plugin directory.

If no `mirror_from` key is specified, then the ODK-generated Makefile will contain a dummy rule instead – it will be up to the user to override that rule in the custom Makefile.

closes #909


template/src/ontology/Makefile.jinja2
```diff
@@ -141,6 +141,47 @@ config_check:
...
+custom_robot_plugins:
+
+{% if project.robot_plugins is defined %}
+.PHONY: extra_robot_plugins
+extra_robot_plugins: {% for plugin in project.robot_plugins.plugins %} $(ROBOT_PLUGINS_DIRECTORY)/{{ plugin.name }}.jar {% endfor %}
+{% endif %}
+
+# Install all ROBOT plugins to the runtime plugins directory
+.PHONY: all_robot_plugins
+all_robot_plugins: $(foreach plugin,$(notdir $(wildcard /tools/robot-plugins/*.jar)),$(ROBOT_PLUGINS_DIRECTORY)/$(plugin)) \
```
matentzn, line 162: I am assuming that $(wildcard /tools/robot-plugins/*.jar) does not throw an error if /tools/robot-plugins/ does not exist;

template/src/ontology/Makefile.jinja2
```diff
@@ -141,6 +141,47 @@ config_check:
 $(TMPDIR) $(REPORTDIR) $(MIRRORDIR) $(IMPORTDIR) $(COMPONENTSDIR) $(SUBSETDIR):
 	mkdir -p $@
 
+# ----------------------------------------
```
matentzn, line 144: Why did you choose to not have `{% if project.robot_plugins is defined %}` wrapping the entire section? Few people will need robot custom plugins, so there is an argument here to be made about makefile hygiene. (a weak argument).

matentzn (approved):
great!! I will test this a bit later, but the design looks good.

matentzn (commented):
Sorry meant to say "comment". Approval _after_ testing !

template/src/ontology/Makefile.jinja2
gouttegd: Because we still need most of this section even if no custom plugins are defined in the spec, to install (1) the plugins that are provided by the ODK and (2) any plugin found in the top-level `plugins` directory.

template/src/ontology/Makefile.jinja2
gouttegd: It doesn’t.

gouttegd:
Of note, I am open to the argument that this may be _too_ flexible: maybe the new ODK configuration section is not needed, and it would actually be enough if supplementary plugins can only be installed through custom rules in the custom Makefile, without the possibility of declaring them in the `-odk.yaml` config file — presumably anyone knowledgeable enough to know that they need a plugin would know how to add a rule in the custom Makefile.

template/src/ontology/Makefile.jinja2
matentzn: I am thinking about the case where the user does not want any of this (custom plugins to be installed from any source). My assumption is that this is 80% of the  ODK setups. In this case, there is really no reason for any extra line of code in the Makefile, no?

Your argument seems to be that "independent of what the user configures, if they put plugins in their user dir for plugins, they should be able to use the ODK makefile to pull them into the project". I think that any plugin the user intends to use should be mentioned in the config file, which means that unless `{% if project.robot_plugins is defined %}` there is no need to provide any code that allows handling plugins. But maybe I am not seeing something here!

matentzn:
FYI: Why did you use a fork for this PR? Makes it a (tiny) bit more cumbersome for me to test (not much just a bit)

template/src/ontology/Makefile.jinja2
gouttegd: > I am thinking about the case where the user does not want any of this (custom plugins to be installed from any source). My assumption is that this is 80% of the ODK setups. In this case, there is really no reason for any extra line of code in the Makefile, no?

What about the plugins provided by the ODK (such as, for now, the SSSOM plugin)?

We want those plugins to be readily usable without users having to do anything special – that’s kind of the entire point of bundling them in the ODK, otherwise we would just ask people who want to use those plugins to configure them in the `plugins` section.

As for the top-level `plugins` directory, I believe it provides useful flexibility that would be more cumbersome to achieve if the only way to use custom plugins was through the config file. For example:

1) What if I want to _temporarily_ override a ODK-provided plugin (e.g., for testing or debugging purposes, I want to use a specific version of `sssom.jar` that is different from the one provided in the ODK)? All I have to do is to put the version I want in the top-level `plugins` directory (and remove it when I no longer need it). If the only way to install extra plugins was through the ODK config file, I would need to either (a) change the configuration and re-generate the Makefile, or (b) override the generated rule to install the plugin in the custom Makefile. Doable, sure, but more cumbersome.

2) What if I wrote a very specialised plugin that only makes sense for the particular ontology I am working on? I can simply put it into the top-level `plugins` folder (where I can optionally commit it to the repository). If the only way to install an extra plugin was through the ODK configuration file, I would need to either make sure the plugin is downloadable from somewhere, or to have a custom rule that copies it from somewhere on the local disk to the runtime plugins directory. Doable again, but again more cumbersome.

gouttegd: As for the “extra lines of code in the Makefile”: those lines would only ever get executed _if a rule declares itself as being dependent on the `all_robot_plugins` target_ (be it a standard rule from the generated Makefile or a custom rule). If your workflow does not include any such rule, those extra lines of code will never run.

gouttegd:
> Why did you use a fork for this PR?

Didn’t do that on purpose, I just forgot that on my local copy the `origin` remote points to my fork instead of here.

template/src/ontology/Makefile.jinja2
gouttegd: If you want _less_ flexibility, I’d rather remove the possibility of declaring extra plugins through the config file and only leave the possibility of having custom installation rules. The config file is the least flexible option of all, so better for it to go instead of making it the only option.

matentzn: Nope, I am fine with the argumentation. Lets say we want to deliver some standard functionality through a ROBOT plugin, there should be a way to deliver it regardless of wether the option is defined or not. Thanks for your explanations!

matentzn (dismissed):
As this is tested by you, I am included to leave at my line by line review and not test it (I am a bit out of breath). If you feel like I _should_ test this, please merge this into the `dev` branch and I will do it next week.

gouttegd:
I am happy with the amount of testing I’ve done myself.

In any case the PR can’t break an existing standard workflow, since there are currently _no_ standard workflows to deal with plugins.

What’s missing now is documenting the plugin delivery mechanism – such a mechanism would not be very useful if people don’t know how to use it (or even that it exists)… Since we agree on the mechanism itself (so it won’t be changed), I’ll do that shortly.

matentzn (approved):
I cant see any other obvious problem. I tested three pipelines and couldn't see an obvious issue.

PR #976, opened 2024-01-06
Fix and refresh schema docs

matentzn:
@hkir-dev this does not work perfectly anymore (the nesting does not work), see changes to schema_documentation.py.

### DIFF
--- a/docs/project-schema.md
+++ b/docs/project-schema.md
@@ -1,45 +1,17 @@
 
 ## ODK Project Configuration Schema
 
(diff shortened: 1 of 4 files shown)


matentzn:
@hkir-dev in case you did not see this yet - If I could get your recommendation on how to fix, I will gladly do it!

hkir-dev:
It appears that the recent update to jsonschema2md is the root cause of the issue. I've applied a bug fix, but I believe your expertise could help identify any lingering problems more efficiently, @matentzn. Could you please review and inform me of any remaining issues?

matentzn:
Thank you @hkir-dev for taking a look and adding a fix :) Looks ok for now, will cycle back on this later

PR #979, opened 2024-01-08
Remove hard coded ODK version in github action container

### LINKED ISSUE #733: Hard coded ODK version in dynamic jinja?
The dynamic jinja files contain conditionals like:

```
{% if env is defined -%}{{env['ODK_VERSION'] or "v1.2.25" }}{%- else %}v1.2.25{% endif %}
```

in various places. This is weird, especially since `v1.2.25` is an ancient version of ODK. Should we replace this case by `latest`?

anitacaron:
Fixes #733


matentzn (approved):
LGTM

PR #986, opened 2024-01-25
Update relationgraph to 2.3.2

matentzn:
@balhoff Do you expect any compatibility issues with semsql etc?


balhoff:
I don't expect any problems.

anitacaron (approved):

PR #987, opened 2024-01-26
Update printout during debug mode

matentzn:
`IMAGE` parameter does not exist anymore.

### DIFF
--- a/template/src/ontology/run.sh.jinja2
+++ b/template/src/ontology/run.sh.jinja2
@@ -74,7 +74,7 @@ if [ x$ODK_DEBUG = xyes ]; then
     # If you wish to change the format string, take care of using
     # non-breaking spaces (U+00A0) instead of normal spaces, to
     # prevent the shell from tokenizing the format string.
-    echo "Running ${IMAGE} with ${ODK_JAVA_OPTS} of memory for ROBOT and Java-based pipeline steps."
+    echo "Running obolibrary/${ODK_IMAGE}:${ODK_TAG} with ${ODK_JAVA_OPTS} of memory for ROBOT and Java-based pipeline steps."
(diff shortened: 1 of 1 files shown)


template/src/ontology/run.sh.jinja2
```diff
@@ -74,7 +74,7 @@ if [ x$ODK_DEBUG = xyes ]; then
     # If you wish to change the format string, take care of using
     # non-breaking spaces (U+00A0) instead of normal spaces, to
     # prevent the shell from tokenizing the format string.
-    echo "Running ${IMAGE} with ${ODK_JAVA_OPTS} of memory for ROBOT and Java-based pipeline steps."
+    echo "Running obolibrary/${ODK_IMAGE}:${ODK_TAG} with ${ODK_JAVA_OPTS} of memory for ROBOT and Java-based pipeline steps."
```
gouttegd, line 77: While you’re at it, you should also change the phrasing, because `${ODK_JAVA_OPTS} of memory` results in something like `-Xmx12G of memory`, which does not make much sense. And people could always use `ODK_JAVA_OPTS` to pass other Java options unrelated to memory.

I’d suggest simply `with ${ODK_JAVA_OPTS} options for ROBOT and…`.

gouttegd (approved):
Good, but please see comment about (my) bad phrasing.

template/src/ontology/run.sh.jinja2
matentzn, line 77:
```suggestion
    echo "Running obolibrary/${ODK_IMAGE}:${ODK_TAG} with '${ODK_JAVA_OPTS}' as options for ROBOT and other Java-based pipeline steps."
```

sic?

gouttegd (approved):

anitacaron (approved):

PR #988, opened 2024-01-29
OAK Validation workflow

### LINKED ISSUE #610: Consider replacing some of the SPARQL QC checks with OAK validate

anitacaron:
Fixes #610


anitacaron:
I used CL as tester in PR https://github.com/obophenotype/cell-ontology/pull/2274

matentzn (commented):
This is sooo exciting! EDIT https://github.com/obophenotype/cell-ontology/pull/2274

I didnt see your comment! I now seeee it! Cool!

matentzn (changes requested):
- Add example yaml file
- Make tested ontology file configurable (noy hardcoded to "simple")
- Make it possible to switch between open and closed-world validation (this needs to be done in oak - you can make and issue there and assign it to Harshad)

matentzn:
We can close it - I will resume it whenever I think it is relevant!

PR #990, opened 2024-02-08
Make qc.yml ODK managed by default

### LINKED ISSUE #797: Switch qc.yml to ODK managed
Right now, qc.yml GitHub actions is user managed. The idea was to make it customisable. The truth is this was a bad idea - I keep having to manually migrate this file because many comments touched by the test keep needing more up to date ODK versions.

matentzn:
This PR:
- Makes qc.yml ODK-managed by default
- Makes it possible to despense with the qc.yml from the beginning, by making it excludable in the config

Fixes #797

### DIFF
--- a/odk/odk.py
+++ b/odk/odk.py
@@ -586,7 +586,7 @@ class OntologyProject(JsonSchemaMixin):
     ci : Optional[List[str]] = field(default_factory=lambda: ['github_actions'])
     """continuous integration defaults; currently available: travis, github_actions, gitlab-ci"""
     
-    workflows : Optional[List[str]] = field(default_factory=lambda: ['docs'])
+    workflows : Optional[List[str]] = field(default_factory=lambda: ['docs', 'qc'])
     """Workflows that are synced when updating the repo. Currently available: docs, diff, qc, release-diff."""
     
     import_pattern_ontology : bool = False
--- a/template/_dynamic_files.jinja2
+++ b/template/_dynamic_files.jinja2
@@ -737,7 +737,7 @@ branches:
 notifications:
   email:
     - obo-ci-reports-all@groups.io
-{% endif -%}{% if 'github_actions' in project.ci %}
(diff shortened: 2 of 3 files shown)


gouttegd (approved):
Looks good overall, but we’ll need to warn users in the release notes of the following issue:

Suppose an ontology is using its own `qc.yml` file _and_ does not have an explicit `workflows` key in its `$(ONT)-odk.yml` configuration file. When `update_repo` is run, the default value of `workflows` will be used, and will instruct the seeding process to overwrite the ontology’s own `qc.yml` with the ODK-provided one – something the ontology maintainers probably don’t want.

Sure, it’s up to whoever runs `update_repo` to check what files are changed and to revert any unwanted change (so in this case, revert to the original `qc.yml` file), but still it’d be nice to warn them in advance.

If you have your own `qc.yml` file (and you don’t want it replaced by the ODK qc), then before running `update_repo`:
* rename your file to something other than `qc.yml`, or,
* explicitly set the `workflows` variable in your `$(ONT)-odk.yml` and make sure it does _not_ contain `qc`.

gouttegd:
Or, maybe all ODK-maintained workflows should use a naming convention that makes it clear that they are ODK-maintained? Like `qc-odk.yml`, `docs-odk.yml`, etc.? This way users could have their own workflows without running the risk of a name clash with a ODK-maintained one.

matentzn:
> Or, maybe all ODK-maintained workflows should use a naming convention that makes it clear that they are ODK-maintained? Like qc-odk.yml, docs-odk.yml, etc.? This way users could have their own workflows without running the risk of a name clash with a ODK-maintained one.

I think this would have been better, but now, what would happen is that all pipelines that update their system get a second QC activated.. I think its best to just roll with this for now, and make it very clear on the announcements. Probably though, the next actions should consider to be named odk-x.yaml - this is, indeed, much cleaner.

PR #998, opened 2024-02-24
Add convenience check if customised ROBOT report config is out of date and update default config

### LINKED ISSUE #552: We do not sync customised ROBOT reports
We allow people to customise their ROBOT reports, but we have no facilitate to update the profiles when new checks are added to ROBOT report (i.e. the profile.txt once it is customised, it is no longer in sync). Maybe provide a utility method to pull the latest profile from robot and merge in new tests.

matentzn:
Fixes #552 

This PR:

- Adds a convenience check if customised ROBOT report config is out of date 
- updates the default config which we have not done in years.. No drive now to make this dynamic, but created an issue: https://github.com/INCATools/ontology-development-kit/issues/999

```
sh run.sh make check_for_robot_updates
....
Missing tests:
equivalent_class_axiom_no_genus
illegal_use_of_built_in_vocabulary
invalid_entity_uri
missing_subset_declaration
missing_synonymtype_declaration
misused_replaced_by
multiple_equivalent_class_definitions
```

### DIFF
--- a/docker/odklite/Dockerfile
+++ b/docker/odklite/Dockerfile
@@ -44,7 +44,7 @@ RUN wget -nv $ROBOT_JAR \
     wget -nv https://raw.githubusercontent.com/ontodev/robot/v$ROBOT_VERSION/bin/robot \
         -O /tools/robot && \
     chmod 755 /tools/robot &&\
-    wget -nv https://raw.githubusercontent.com/ontodev/robot/v$ROBOT_VERSION/src/main/resources/report_profile.txt \
+    wget -nv https://raw.githubusercontent.com/ontodev/robot/v$ROBOT_VERSION/robot-core/src/main/resources/report_profile.txt \
         -O /tools/robot_report_profile.txt
 
 # Install DOSDPTOOLS.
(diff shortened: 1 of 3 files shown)


gouttegd:
I don’t understand the point of this.

Let’s say a new test is added to the ROBOT’s master branch. What good does it do to ODK users to know that a new test is available in a version of ROBOT that is not the version that is included in the ODK, and that may not even be published yet?

matentzn:
Hmm, it's true that there could be a small lag between master and release, didn't think of that. But I was hoping to make it easier for people who use a custom profile to get notified if ROBOT supports new tests. Already many ontologies I work with use custom profile.txt files, and are 2-6 tests behind what robot has to offer.. maybe you have a better idea of how to make it easy for a user to understand when there are new tests available?

gouttegd:
Given the current frequencies of ROBOT releases, the lag is likely to be quite substantial (a few months at least).

What I would suggest is to bundle the ROBOT default profile _at build time_. That is, we store somewhere in the ODK a copy of the ROBOT profile (let’s say in `/tools/robot/default_profile.txt`), and we compare the profile actually used in the ontology to that.

No need to download anything from the Internet, and this will let people know if there are any tests that are supported by the version of ROBOT available in the ODK that their custom profile does not include.

Of course, when we update ROBOT in the ODK (as we do in point releases), we update the profile accordingly, so that `/tools/robot/default_profile` always contain the up-to-date list of all the tests that are supported by the version of ROBOT provided with the ODK.

gouttegd:
> Is this what you mean?

Yes, exactly.

gouttegd (approved):

PR #1000, opened 2024-02-25
Add babelon toolkit to requirements.txt

matentzn:
`babelon` is a library to manage multi-lingual ontologies.

The babelon schema is used already by ontologies like [HPO](https://github.com/obophenotype/human-phenotype-ontology/blob/master/src/ontology/hp.Makefile) and [MP](https://github.com/mgijax/mammalian-phenotype-ontology/blob/04ba2825644ec494d0991d8fad16594ab4192e08/src/ontology/mp.Makefile#L112)

The recent babelon release (my pet project last week) allows removing all this boilerplate we needed to hack the linkml conversion process, and also provides some basic functionalities to translate ontologies automatically.

### DIFF
--- a/requirements.txt.full
+++ b/requirements.txt.full
@@ -2,6 +2,7 @@
 sssom
 oaklib
 linkml
+babelon
 dosdp
 semsql
 
--- a/requirements.txt.lite
+++ b/requirements.txt.lite
@@ -15,3 +15,4 @@ j2cli[yaml]
 j2cli
 lightrdf
 sssom
+babelon


matentzn:
@gouttegd I fixed the previous issues with the constraints.txt, this is ready for review.

requirements.txt.full
```diff
@@ -2,6 +2,7 @@
 sssom
 oaklib
 linkml
+babelon
```
gouttegd, line 5: No objection to adding that.

Just, have you already tried building the ODK with it? We just need to be sure this addition does not create yet another mess with Python dependencies (e.g. if Babelon depends on on _foo_ > 1.2.3 and we have another Python packages that requires _foo <= 1.1.1).

gouttegd (approved):

requirements.txt.lite
```diff
@@ -15,3 +15,4 @@ j2cli[yaml]
 j2cli
 lightrdf
 sssom
+babelon
```
gouttegd, line 18: On second thought, not sure this belongs to ODKLite.

The rationale for ODKLite is that it should be the minimal ODK image needed to run the standard ODK workflows. As long as Babelon is not used in any of said workflows, I’d say it belongs to ODKFull only.

gouttegd (changes requested):

requirements.txt.full
matentzn: Yes I did and I had to make a new release! Our QC actually takes this into account.

requirements.txt.lite
matentzn: Transcribing right now the "$(ONT)-international.owl" pipeline from HPO and MP which will allow people to add an internationalised release. Testing this right now, will make PR shortly.

gouttegd: You mean that the HPO and MP internationalisation rules will get upstreamed into the ODK to become part of the standard workflows?

OK then, in that case no objection.

PR #1004, opened 2024-02-25
Fix the cleaning up rule.

### LINKED ISSUE #1002: Cleaning up rule errors out when a directory does not exist
The standard `clean` rule raises an error when one of the directories it is supposed to clean up does not exist.

For example, when the `target` directory (`$(UPDATEREPODIR)`) does not exist (which would be the case if `make update_repo` has never been called yet in the lifetime of the repo):

```sh
sh run.sh make clean 
[ -n "mirror" ] && [ mirror != "." ] && [ mirror != "/" ] && [ mirror != ".." ] && [ -d ./mirror ] && rm -rf ./mirror/*
[ -n "tmp" ] && [ tmp != "." ] && [ tmp != "/" ] && [ tmp != ".." ] && [ -d ./tmp ] && rm -rf ./tmp/*
[ -n "target" ] && [ target != "." ] && [ target != "/" ] && [ target != ".." ] && [ -d ./target ] && rm -rf ./target/*
make: *** [Makefile:571: clean] Error 1
```

This is because of this:

```sh
... && [ -d ./target ] && ...
```

If the directory does not exist, then `[ -d ./target ]` returns 1, the rest of the line is not evaluated, and the return value of the entire command is 1 (which the shell interprets – correctly – as an error).

Several ways to fix that.

A) Just adding ` || true` at the end of the command:

```sh
[ -n "target" ] && [ target != "." ] && [ target != "/" ] && [ target != ".." ] && [ -d ./target ] && rm -rf ./target/* || true
```

so that if any of the tests fail, the entire command is still considered successful.

B) Enclose all the tests in a `if ... then ... fi` construct, and only run the `rm` command if none of the tests failed:

```sh
if [ -n "target" ] && [ target != "." ] && [ target != "/" ] && [ target != ".." ] && [ -d ./target ] ; then
   rm -rf ./target/*
fi
```

C) Same but using a single test command with `-a`-chained conditions instead of multiple `&&`-chained commands:

```sh
if [ -n "target" -a target != "." -a target != "/" -a target != ".." -a -d ./target ] ; then
   rm -rf ./target/*
fi
```

D) Overhaul the entire command.

The clean-up command tries to ensure that we don’t inadvertently delete files that shouldn’t be deleted, if for example `TMPDIR` is set to an unexpected value (by human error or malicious intent). It does a poor job of that, however. It only tries to catch _some_ unexpected values. If we are serious about protecting users against inadvertent/malicious deletion, we can do better.

For example, something like this:

```make
clean:
        for dir in $(MIRRORDIR) $(TMPDIR) $(UPDATEREPODIR) ; do \
                reldir=$$(realpath --relative-to=$$(pwd) $$dir) ; \
                case $$reldir in .*) ;; *) rm -rf $$reldir/* ;; esac \
        done
        rm -f $(CLEANFILES)
```

which only run the `rm` command if the directory to clean up is stricly under the `src/ontology` directory.

gouttegd:
The existing rule to cleanup the mirroring, temporary, and update directories suffers from two distinct issues:

1) it errors out if the directory to clean up does not exist;
2) it does not try hard enough to ensure that we are not deleting something that we shouldn’t, in the event that one of the `*DIR` variables point to something unexpected.

This PR fixes both issues by replacing the existing checks by a check that:

1) computes the pathname of the directory to clean up relative to the current directory (`src/ontology`);
2) only cleans up the directory if the pathname computed above does not start with `.`, indicating that the directory is indeed located below `src/ontology`.

closes #1002

### DIFF
--- a/template/src/ontology/Makefile.jinja2
+++ b/template/src/ontology/Makefile.jinja2
@@ -1203,14 +1203,17 @@ update_docs:
 	mkdocs gh-deploy --config-file ../../mkdocs.yaml
 {%- endif %}
 
-# Note to future generations: prepending ./ is a safety measure to ensure that 
-# the environment does not malicously set `CLEANFILES` to `\`.
+# Note to future generations: computing the real path relative to the
+# current directory is a way to ensure we only clean up directories that
+# are located below the current directory, regardless of the contents of
+# the *DIR variables.
 .PHONY: clean
 clean:{% if project.use_dosdps %}
 	$(MAKE) pattern_clean{%- endif %}
-	[ -n "$(MIRRORDIR)" ] && [ $(MIRRORDIR) != "." ] && [ $(MIRRORDIR) != "/" ] && [ $(MIRRORDIR) != ".." ] && [ -d ./$(MIRRORDIR) ] && rm -rf ./$(MIRRORDIR)/*
-	[ -n "$(TMPDIR)" ] && [ $(TMPDIR) != "." ] && [ $(TMPDIR) != "/" ] && [ $(TMPDIR) != ".." ] && [ -d ./$(TMPDIR) ] && rm -rf ./$(TMPDIR)/*
-	[ -n "$(UPDATEREPODIR)" ] && [ $(UPDATEREPODIR) != "." ] && [ $(UPDATEREPODIR) != "/" ] && [ $(UPDATEREPODIR) != ".." ] && [ -d ./$(UPDATEREPODIR) ] && rm -rf ./$(UPDATEREPODIR)/*
+	for dir in $(MIRRORDIR) $(TMPDIR) $(UPDATEREPODIR) ; do      \
+		reldir=$$(realpath --relative-to=$$(pwd) $$dir) ;    \
+		case $$reldir in .*|"") ;; *) rm -rf $$reldir/* ;; esac \
+	done
 	rm -f $(CLEANFILES)
 
 .PHONY: help


template/src/ontology/Makefile.jinja2
```diff
@@ -1203,14 +1203,17 @@ update_docs:
...
+# are located below the current directory, regardless of the contents of
+# the *DIR variables.
 .PHONY: clean
 clean:{% if project.use_dosdps %}
 	$(MAKE) pattern_clean{%- endif %}
-	[ -n "$(MIRRORDIR)" ] && [ $(MIRRORDIR) != "." ] && [ $(MIRRORDIR) != "/" ] && [ $(MIRRORDIR) != ".." ] && [ -d ./$(MIRRORDIR) ] && rm -rf ./$(MIRRORDIR)/*
-	[ -n "$(TMPDIR)" ] && [ $(TMPDIR) != "." ] && [ $(TMPDIR) != "/" ] && [ $(TMPDIR) != ".." ] && [ -d ./$(TMPDIR) ] && rm -rf ./$(TMPDIR)/*
-	[ -n "$(UPDATEREPODIR)" ] && [ $(UPDATEREPODIR) != "." ] && [ $(UPDATEREPODIR) != "/" ] && [ $(UPDATEREPODIR) != ".." ] && [ -d ./$(UPDATEREPODIR) ] && rm -rf ./$(UPDATEREPODIR)/*
+	for dir in $(MIRRORDIR) $(TMPDIR) $(UPDATEREPODIR) ; do      \
+		reldir=$$(realpath --relative-to=$$(pwd) $$dir) ;    \
```
matentzn, line 1214: Just because I have never seen this before. So lets say someone maliciously sets `MIRRORDIR` to `/` or `~`. Lets further assume this is run outside the docker container (just for the argument, not because I think this is likely; someone might do it mistakenly). Is there no risk with `--relative-to` to delete the root directory?

matentzn: I just read your issue https://github.com/INCATools/ontology-development-kit/issues/1002

still not understand exactly how the --relative-to works technically..

gouttegd: No. Let’s say `MIRRORDIR` is set to `/`. Then the call to `realpath` will set `reldir` to something like `../../../../` (the number of `../` dependent on how deep the current directory is compared to the root, e.g. if we are in `/home/users/alice/myontology/src/ontology/`, `reldir` will be `../../../../../../`, because the root is 6 directories above the current directory).

Then in the following line, we will fall into the first pattern of the `case..esac` construct: `case $reldir in .*|"") ;;` (because `reldir` will start with a dot), so the `rm` command will not get executed.

This generalizes to any scenario where `MIRRORDIR` is set to any directory that is _above_ the current directory (or that _is_ the current directory).

gouttegd: If you’re concerned about the call to `realpath`, a slightly simpler approach would be something like that:

```sh
for dir in $(MIRRORDIR) $(TMPDIR) $(UPDATEREPODIR) ; do
    case $dir in .*|/*|"") ;; *) rm -rf $dir/* ;; esac
done
```

where the first pattern of the `case..esac` construct will catch

* any case where `$dir` points to any directory above the current directory (because it starts with `.`);
* any case where `$dir` is an absolute pathname (because it starts with `/` ; this includes the case where `$dir` _is_ `/`)
* the case where `$dir` is empty (so that we do not run a `rm -rf /*` command).

But contrary to the version with `realpath`, this will _not_ catch cases where a malicious user hides the fact that he is trying to go upward in the filesystem hierarchy by:

* first going downward, for example by setting `$dir` to `tmp/../../../../../`;
* creating a symlink that points to a upward directory, and setting `$dir` to it.

Using `realpath` would catch those, because `realpath` (1) canonicalizes the pathname to remove any superfluous `../`), and (2) resolves any symbolic links along the way.

matentzn: I am not concerned. Just wondering about any risks as I can personally not reed this line of code - technically I mean. I trust your judgement!

matentzn (approved):
More trusting here than anything else, but your explanation fills me with confidence

PR #1006, opened 2024-02-25
Add international edition to ODK

matentzn:
This PR is a monster. It was fun to do it, and it took me more than 10 hours to write.. There will probably be issues here and there - a review should probably focus on the places where the change affects previous parts of the code. I did some extensive testing, but I am still on it a bit.

### DIFF
--- a/odk/odk.py
+++ b/odk/odk.py
@@ -189,7 +189,34 @@ class SSSOMMappingSetProduct(Product):
     sssom_tool_options: Optional[str] = ""
     """SSSOM toolkit options passed to the sssom command used to generate this product command"""
 
+@dataclass_json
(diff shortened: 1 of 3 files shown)


gouttegd (changes requested):
Not much to review since most of the actual work is done by Babelon, which I did _not_ review.

Two concerns: 1) reference to an `OPENAI_API_KEY` out of the blue (how are users supposed to communicate that key), and 2) theft that “auto translate” relies on LLM services must be explicitly documented.

gouttegd (dismissed):

matentzn:
@gouttegd sorry there was a small bug, its done now. :/ I will merge once QC passes.

PR #995, opened 2024-02-22
Update ROBOT plugins

gouttegd:
This PR:

* updates the ROBOT SSSOM plugin to version 0.7.5;
* adds the ROBOT KGCL plugin (in its version 0.3.2).

closes #992

PR #1009, opened 2024-02-27
Update the documentation on ROBOT plugins.

gouttegd:
This commit updates the page on the use of ROBOT plugins to: (1) add the KGCL plugin to the list of plugins that are bundled with the ODK (so that people know they don't need to manually configure that plugin), and (2) use the Uberon plugin rather than the KGCL plugin in examples about how to configure a non-bundled plugin, to avoid needless confusion.

(This should have been done as part of #995.)

### DIFF
--- a/docs/UsingRobotPlugins.md
+++ b/docs/UsingRobotPlugins.md
@@ -10,20 +10,20 @@ Before you can use plugins in a custom workflow, the ODK must be aware of those 
 
 Add a new `robot_plugins` section to your ODK configuration file (`src/ontology/ONT-odk.yaml`). That section should contain a single `plugins` key, which itself should contain the list of the plugins you want to use. Each entry in the list must contain at least a `name` key, which is the name under which the plugin will be available, and optionally a `mirror_from` key, pointing to an online location from which the plugin can be downloaded.
 
-For example, to use the [KGCL](https://github.com/gouttegd/kgcl-java/) plugin:
+For example, to use the [Uberon plugin](https://github.com/gouttegd/uberon-robot-plugin):
 
 ```yaml
 robot_plugins:
   plugins:
-    - name: kgcl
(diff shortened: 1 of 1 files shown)


matentzn (approved):
LGTM

gouttegd:
Any idea on what’s going on with the CI check? It has been stuck on “waiting for status to be reported” ever since I created the PR, and I can’t merge anything until it’s done.

anitacaron:
The conflict between not needing to run QC because it's just a change in an MD file and the need to pass QC to be able to merge?

gouttegd:
> The conflict between not needing to run QC because it's just a change in an MD file and the need to pass QC to be able to merge?

If so, that’s pretty dumb from GitHub. I would have expected that the workflow not running at all (because the PR doesn’t meet the criteria) would count as a pass. 🤦

gouttegd:
> If so, that’s pretty dumb from GitHub.

And yet [that is exactly the case](https://docs.github.com/en/actions/managing-workflow-runs/skipping-workflow-runs):

> If a workflow is skipped due to [path filtering](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore), [branch filtering](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpull_requestpull_request_targetbranchesbranches-ignore) or a commit message (see below), then checks associated with that workflow will remain in a "Pending" state. A pull request that requires those checks to be successful will be blocked from merging.

So basically a PR that only changes the documentation (and therefore won’t trigger the CI workflow) can’t be merged.

Am I supposed to push a commit with a dummy change in the “code” part of the repository just so that we can run a completely useless CI check just so that I can merge a documentation change? The level of stupidity is staggering.

gouttegd:
Fuck it, the doc will remain not updated until I have some unrelated code changes to push. _Not_ going to make a dummy commit just to make GitHub happy.

matentzn:
Why didn't you just merge this?

PR #1022, opened 2024-03-02
HOTFIX: Fix the qc.yml to not run mirrors

matentzn:
This is necessary to be pushed as a hotfix quickly, as some QC pipelines can run mirrors without failing. 

I didnt notice this because of chicken egg - I couldnt test qc.yml directly, and it failed because it was not able to find the v1.5 ODK release..

### DIFF
--- a/template/_dynamic_files.jinja2
+++ b/template/_dynamic_files.jinja2
@@ -793,7 +793,7 @@ jobs:
       - name: Run ontology QC checks
         env:
           DEFAULT_BRANCH: {{ project.git_main_branch }}
-        run: cd src/ontology && make ROBOT_ENV='ROBOT_JAVA_ARGS=-Xmx6G' test IMP=false PAT=false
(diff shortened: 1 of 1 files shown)


gouttegd (approved):
No objection to the fix. But I am not sure I understand why this is necessary.

Making the mirror-downloading rules independent of `IMP` was done to allow _custom_ workflows to depend on a mirror independently of the import pipeline.

But as far as I know, in the _standard_ workflows, the mirrors are only used to refresh the imports, so if we run with `IMP=false`, no refresh of the imports should be attempted and therefore no mirror-downloading rule should be invoked.

matentzn:
Maybe the pipelines that failed because of this failed because of custom goals. 

This seems to have been the case with OBA: https://github.com/obophenotype/bio-attribute-ontology/actions/runs/8125158836/job/22207444205

But with MAXO it _seems_ to me that mirror/merged.owl was being created for some reason during the QC. this i perhaps because `mirror-%` is PHONY and therefore executed no matter what?

PR #1026, opened 2024-03-11
Add correct default and casing to release_annotate_inferred_axioms

matentzn:
Previously, binary values were printed in upper case.


gouttegd (approved):
All good, but note that ROBOT accepts any casing for the value of boolean options (`true`, `TRUE`, `True`, `tRuE`, etc.), so lower casing herr is not strictly necessary.

matentzn:
Yes, this is not essential, but I like to be consistent. Thanks for checking!

PR #1042, opened 2024-04-06
Add mermaid 2 to ODK docs

matentzn:
This enables us to add nice diagrams like https://monarch-initiative.github.io/ontogpt/phenotype/, widely used in LinkML style docs.


matentzn:
I have merged this as this is restoring the previous state.

PR #1067, opened 2024-05-30
Ugrade base system to Ubuntu 24.04

### LINKED ISSUE #1008: Upgrade the base system to 24.04
The next Ubuntu LTS (24.04, “Noble Numbat”) is scheduled to be released in April 2024. This issue is to remind us to upgrade our images to this new base system when it is available.

gouttegd:
This PR updates the base system of all our images to the latest LTR version of Ubuntu, 24.04.

* Some system packages that were previously implicitly installed now need to be explicitly asked for (e.g. `libpcre3`, needed for Konclude, or `npm`, needed to install Obographviz).
* The OBO Dashboard must now be installed with `--break-system-packages`, otherwise `pip` will flatly refuse to install it system-wide.
* We can no longer have `virtualenv` in the builder image, as it messes with the installation of our Python packages.
* The `odk.py` script has some issues with Python 3.12 that need fixing. 

closes #1008

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -97,7 +97,7 @@ COPY scripts/obodash /tools
 RUN chmod +x /tools/obodash && \
     git clone --depth 1 https://github.com/OBOFoundry/OBO-Dashboard.git && \
     cd OBO-Dashboard && \
-    python3 -m pip install -r requirements.txt && \
+    python3 -m pip install -r requirements.txt --break-system-packages && \
     echo " " >> Makefile && \
     echo "build/robot.jar:" >> Makefile && \
     echo "	echo 'skipped ROBOT jar download.....' && touch \$@" >> Makefile && \
--- a/Dockerfile
+++ b/Dockerfile
@@ -40,6 +40,7 @@ RUN apt-get update && DEBIAN_FRONTEND="noninteractive" apt-get install -y --no-i
     xlsx2csv \
     gh \
     nodejs \
+    npm \
     graphviz \
     python3-psycopg2 \
     swi-prolog
--- a/Dockerfile
+++ b/Dockerfile
@@ -43,7 +43,8 @@ RUN apt-get update && DEBIAN_FRONTEND="noninteractive" apt-get install -y --no-i
     npm \
     graphviz \
     python3-psycopg2 \
-    swi-prolog
+    swi-prolog \
+    libpcre3
 
 # Install run-time dependencies for Soufflé.
 RUN DEBIAN_FRONTEND="noninteractive" apt-get install -y --no-install-recommends \
--- a/docker/builder/Dockerfile
+++ b/docker/builder/Dockerfile
@@ -27,8 +27,7 @@ RUN apt-get update && \
         rustc \
         cargo \
         python3-dev \
-        python3-pip \
-        python3-virtualenv
+        python3-pip
 
 # Build the Python packages.
 # On x86_64, most if not all of these packages should be available as
--- a/docker/builder/Dockerfile
+++ b/docker/builder/Dockerfile
@@ -44,6 +44,7 @@ COPY constraints.txt /build/constraints.txt
 RUN python3 -m pip install \
         -r /build/requirements.txt.lite \
         -c /build/constraints.txt \
+        --no-warn-script-location \
         --root /staging/lite
 # Then those needed by the odkfull image.
 # After installing those packages, we forcibly remove from the odkfull
@@ -53,6 +54,7 @@ RUN python3 -m pip install \
 RUN python3 -m pip install \
         -c /build/constraints.txt \
         -r /build/requirements.txt \
+        --no-warn-script-location \
         --root /staging/full && \
         cd /staging/lite && \
         find . -type f | while read f ; do rm -f /staging/full/$f ; done && \
--- a/docker/builder/Dockerfile
+++ b/docker/builder/Dockerfile
@@ -3,7 +3,7 @@
 # (typically because pre-compiled binaries don't exist for arm64) and
 # we don't want to build directly on the final image (to avoid
 # cluttering the image with build-time dependencies).
-FROM ubuntu:22.04
+FROM ubuntu:24.04
 WORKDIR /build
 
 # Software versions
--- a/odk/odk.py
+++ b/odk/odk.py
@@ -778,7 +778,7 @@ class ExecutionContext(JsonSchemaMixin):
     project : Optional[OntologyProject] = None
     meta : str = ""
 
-    
+
 @dataclass
 class Generator(object):
     """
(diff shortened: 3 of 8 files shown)


Dockerfile
```diff
@@ -95,7 +97,7 @@ COPY scripts/obodash /tools
 RUN chmod +x /tools/obodash && \
     git clone --depth 1 https://github.com/OBOFoundry/OBO-Dashboard.git && \
     cd OBO-Dashboard && \
-    python3 -m pip install -r requirements.txt && \
+    python3 -m pip install -r requirements.txt --break-system-packages && \
```
matentzn, line 100: > The OBO Dashboard must now be installed with --break-system-packages, otherwise pip will flatly refuse to install it system-wide.

I wonder if this is very suboptimal and we should, instead, install OBO dashboard as a pypi package and install it the usual way? Why is `--break-system-packages` not needed in `docker/builder/Dockerfile`?

odk/odk.py
```diff
@@ -787,7 +787,7 @@ class Generator(object):
     """
 
     ## TODO: consider merging Generator and ExecutionContext?
-    context : ExecutionContext = ExecutionContext()
+    context : ExecutionContext = field(default_factory=ExecutionContext)
```
matentzn, line 790: Is this because of the newer dataclass version in the newer python?

docker/builder/Dockerfile
```diff
@@ -27,8 +27,7 @@ RUN apt-get update && \
         rustc \
         cargo \
         python3-dev \
-        python3-pip \
-        python3-virtualenv
```
matentzn, line 31: > We can no longer have virtualenv in the builder image, as it messes with the installation of our Python packages.

What about pipelines based on ODK that require the ability to install a virtual environment before running system? Seems quite fundamental a thing to remove from the image.. How does it "mess"?

matentzn (commented):
> Some system packages that were previously implicitly installed now need to be explicitly asked for (e.g. libpcre3, needed for Konclude, or npm, needed to install Obographviz).

Ok. 

> The odk.py script has some issues with Python 3.12 that need fixing.

I assume this is primarity the dataclass field issue, right? So all fixed?

Dockerfile
anitacaron: What are the break system packages? I can update to the latest version, if possible.

Dockerfile
```diff
@@ -97,7 +97,7 @@ COPY scripts/obodash /tools
 RUN chmod +x /tools/obodash && \
```
anitacaron, line 97: > Fuck it, we are _not_ going to create a virtual environment just for the
OBO Dashboard. If the OBO Dashboard dependencies end up breaking the
rest of our Python packages, we might as well just create a separate
Docker image just for the dashboard.

Your commit messages are great, but it's important to avoid using offensive language in them. Thank you!

Dockerfile
gouttegd: > I wonder if this is very suboptimal and we should, instead, install OBO dashboard as a pypi package and install it the usual way?

We could, but then we would need actual _releases_ of the OBO Dashboard package whenever we want to update it in the ODK. For now, the OBO Dashboard is effectively working in a “rolling release” fashion: [it doesn’t do releases](https://github.com/OBOFoundry/OBO-Dashboard/releases) and we always install it from the tip of the master branch. That would need to change.

odk/odk.py
gouttegd: Yes.

The other fix that the `odk.py` script needed was the removal of a bogus escape sequence in one of the messages it displays to the user (`\@`).

docker/builder/Dockerfile
gouttegd: We are talking about the *builder* image here! This does not remove anything from the final ODKFull!

Having `virtualenv` installed in the builder image creates a mess because

> the Ubuntu package for virtualenv automatically installs platformdirs version 2.5.1, which then prevents us from installing the platformdirs 4.x that we need as a dependency for some of our packages.

Dockerfile
gouttegd: Sorry but I won’t take lessons in how to write commit messages from folks who can’t be bothered to write decent commit messages themselves.

If people complain about an occasional profanity in one of my commit messages, at least I will know that there _are_ some people who read them.

Dockerfile
anitacaron: > That would need to change.

I can create a release right now.

Dockerfile
matentzn: Correct me if I am wrong but I dont think this was meant as a lesson, more like a statement of affectedness. 

I also prefer to avoid strong language in the spirit of conduct (as the Covenant Code of Conduct suggests), I had endless discussions of misconduct in some other projects which just distract from our common goals.

Dockerfile
gouttegd: > Why is `--break-system-packages` not needed in `docker/builder/Dockerfile`?

Because in `docker/builder/Dockerfile` we are _not_ installing Python packages to the system-wide directory: we are installing them to a staging area, from where we copy them to the ODKLite/ODKFull images.

Dockerfile
gouttegd: > Correct me if I am wrong

My pleasure.

> I dont think this was meant as a lesson, more like a statement of affectedness.

A “statement of affectedness” would have been something like, “I felt offended by the presence of that F-word, I don’t like reading such things when I am reviewing a PR. I’d appreciate if you abstained to do that“ or similar.

“It’s important to avoid using offensive language in [commit messages]” is absolutely a lesson in how one should write a commit message – one that I find hard to accept coming from someone who I can’t remember having ever seen write a single decent commit message that actually explains what they are trying to do.

> as the Covenant Code of Conduct suggests

The ODK does _not_ have a code of conduct. Maybe it should (though I’d much rather have a “code on how to write efficient commit messages so that future contributors don’t have to guess what the f... the developers had in mind when they committed that change”, but it’s OK, we can have different priorities), but currently it does not.

Even if it had, according to the most popular code of conducts in use out there, the correct response to the occasional use of inappropriate language is a “*private*, written warning from community leaders, providing clarity around the nature of the violation and an explanation of why the behavior was inappropriate” – not a public rebuke.

> I had endless discussions of misconduct in some other projects which just distract from our common goals

Fine. I hereby publicly apologise for having expressed my state of mind, as to why I believe installing the OBO Dashboard in a virtual environment just to circumvent `pip`’s default behaviour would be a unnecessary complication, using a 6-letters profanity instead of a more elaborate, professional, and polite phrasing. I did that for the sake of efficiency, but I didn’t consider that indeed people might not want to read a F-word when they are busy reviewing a PR.

I hope this will end the current distraction, and I will temporarily withdraw for all ODK maintenance to be sure you are not distracted again.

PR #1075, opened 2024-06-24
Revert "Update to newest owltools to align OWL API with ROBOT 1.9.6"

matentzn:
This reverts commit 716a6b6b9e8c6b5c05b834c6c3abb027df747912.

For now, it its better for the stability of pipelines that depend on owltools such as Mondo, Uberon to stick with the older version of owltools. These pipeline require to process owtools generated obo format for example with perl scripts and many other "ad hoc" scripts, and even small serialisation changes cause us problem now.

@balhoff we would prefer to freeze the owltools in time, and move everything to ROBOT now. Could you make an issue that shows the "garbage encountered in the GO pipeline" so we can think of a better solution for this?


gouttegd (approved):

balhoff (approved):
@matentzn this is fine for now, although I have made a few more changes to owltools to correct the problems we saw. But I don't have time this week to thoroughly test it. I expect we will need the owltools update for GO. We can use our own copy if need be.

PR #1076, opened 2024-06-24
Upgrade base system to Ubuntu 24.04

### LINKED ISSUE #1008: Upgrade the base system to 24.04

gouttegd:
This PR updates the base system of all our images to the latest LTR version of Ubuntu, 24.04.

* Some system packages that were previously implicitly installed now need to be explicitly asked for (e.g. `libpcre3`, needed for Konclude, or `npm`, needed to install Obographviz).
* The OBO Dashboard must now be installed with `--break-system-packages`, otherwise `pip` will by default refuse to install it system-wide (the idea is that only APT tools are allowed to install anything system-wide).
* We can no longer have `virtualenv` in the builder image, as it messes with the installation of our Python packages.
* The `odk.py` script has some issues with Python 3.12 that need fixing.

closes #1008


gouttegd:
(For those keeping track, this the same PR as #1067 modulo the wording of some commit messages and freshly updated Python constraints.)

matentzn:
I was almost done testing, when I encountered some issues with some incompatible python packages. Can you instruct me again how to update constraints.txt locally? Just running the update script locally causes errors like 

```
Collecting matplotlib_venn (from -r requirements.txt.full (line 20))
  Using cached matplotlib-venn-1.0.0.tar.gz (34 kB)
  Installing build dependencies ... done
  Getting requirements to build wheel ... error
  error: subprocess-exited-with-error
  
  × Getting requirements to build wheel did not run successfully.
  │ exit code: 1
  ╰─> [18 lines of output]
      Traceback (most recent call last):
        File "/Users/matentzn/ws/ontology-development-kit/tmpdir/lib/python3.11/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 353, in <module>
          main()
        File "/Users/matentzn/ws/ontology-development-kit/tmpdir/lib/python3.11/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 335, in main
          json_out['return_val'] = hook(**hook_input['kwargs'])
                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        File "/Users/matentzn/ws/ontology-development-kit/tmpdir/lib/python3.11/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 118, in get_requires_for_build_wheel
          return hook(config_settings)
                 ^^^^^^^^^^^^^^^^^^^^^
        File "/private/var/folders/vj/ks1_0k8x3t9ftrwcr0t9vjwr0000gn/T/pip-build-env-1rlkx9mk/overlay/lib/python3.11/site-packages/setuptools/build_meta.py", line 327, in get_requires_for_build_wheel
          return self._get_build_requires(config_settings, requirements=[])
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        File "/private/var/folders/vj/ks1_0k8x3t9ftrwcr0t9vjwr0000gn/T/pip-build-env-1rlkx9mk/overlay/lib/python3.11/site-packages/setuptools/build_meta.py", line 297, in _get_build_requires
          self.run_setup()
        File "/private/var/folders/vj/ks1_0k8x3t9ftrwcr0t9vjwr0000gn/T/pip-build-env-1rlkx9mk/overlay/lib/python3.11/site-packages/setuptools/build_meta.py", line 313, in run_setup
          exec(code, locals())
        File "<string>", line 14, in <module>
      ModuleNotFoundError: No module named 'setuptools.command.test'
      [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
error: subprocess-exited-with-error

× Getting requirements to build wheel did not run successfully.
│ exit code: 1
╰─> See above for output.

note: This error originates from a subprocess, and is likely not a problem with pip.
```

And I am not sure if this is due to my local setup or something else. I also tried to do this from within the freshly build ODK container, but the error seems the same..

Dockerfile
matentzn, line 1: When using old, legacy odk wrappers I get a (non-terminating) error message we may want to consider:

run.sh:

```
docker run --memory=${MEMORY_GB}g  -v $PWD/../../:/work -w /work/src/ontology -e ROBOT_JAVA_ARGS=${MEMORY_JAVA} -e JAVA_OPTS=${MEMORY_JAVA} --rm -ti obolibrary/$IMAGE $TIMECMD "$@"
```

Results in:

```
sh run.sh robot query -i mondo-tags-reasoned.owl --query ../sparql/qc/mondo/qc-preferred-external-no-equivalent.sparql out.tsv
Error response from daemon: manifest for obolibrary/odkfull:v1.6 not found: manifest unknown: manifest unknown
```

gouttegd:
> Can you instruct me again how to update constraints.txt locally? 

We don’t have a standard procedure for that when we update the base system. The normal procedure (`update-constraints.sh`) is not appropriate in this situation.

The idea is to create a Ubuntu-based Docker images with the same packages (Ubuntu packages, not Python packages) as in the builder image, then within that image try to install all the Python packages _without the current constraints_. Then freeze the resulting Python environment to produce a whole new `constraints.txt` file. Working on that now.

> And I am not sure if this is due to my local setup or something else.

I reproduce the error here. From what I can tell, this is due to `setuptools` 72.0.0 (https://setuptools.pypa.io/en/stable/history.html#v72-0-0pip) which has removed a command that has been deprecated for something like 5 years, but that somehow Python package developers were (and _are_) still using. That’s the Python ecosystem for you.

Dockerfile
gouttegd: So maybe don’t use old, legacy odk wrappers?

gouttegd: Does the wrapper you used in the command above happen to include a `docker pull` command, by any chance?

update-constraints.sh
matentzn, line 1: right now I am relying a lot on the commandline tool j2 in various pipelines.

```
j2 --format=json config/workshops.md.jinja2 tmp/workshops.json > ../../docs/community/workshops.md
Traceback (most recent call last):
  File "/usr/local/bin/j2", line 5, in <module>
    from j2cli import main
  File "/usr/local/lib/python3.12/dist-packages/j2cli/__init__.py", line 10, in <module>
    from j2cli.cli import main
  File "/usr/local/lib/python3.12/dist-packages/j2cli/cli.py", line 8, in <module>
    import imp, inspect
ModuleNotFoundError: No module named 'imp'
make: *** [hp.Makefile:538: ../../docs/community/workshops.md] Error 1
(.venv) ➜  ontology git:(workshop-pages) ✗ 
```

This is a known, seemingly wide ranging issue: 
https://stackoverflow.com/questions/77401730/modulenotfounderror-no-module-named-imp

I wonder if we can somehow decouple these python 12 problems from the base system migration because that, all by itself, seems to be fine to me after a weekend of running various tests..

Dockerfile
matentzn: Oooooops. Say no more:

```
MEMORY_GB=${MEMORY_GB:-60}
IMAGE=${IMAGE:-odkfull:v1.6}
MEMORY_JAVA="-Xmx${MEMORY_GB}G"
ODK_DEBUG=${ODK_DEBUG:-no}

docker pull obolibrary/$IMAGE

echo "Running ODK with ${MEMORY_GB} GB of memory, using ${IMAGE}."

TIMECMD=
if [ x$ODK_DEBUG = xyes ]; then
    # If you wish to change the format string, take care of using
    # non-breaking spaces (U+00A0) instead of normal spaces, to
fi

docker run --memory=${MEMORY_GB}g  -v $PWD/../../:/work -w /work/src/ontology -e ROBOT_JAVA_ARGS=${MEMORY_JAVA} -e JAVA_OPTS=${MEMORY_JAVA} --rm -ti obolibrary/$IMAGE $TIMECMD "$@"
```

Comment closed.

update-constraints.sh
gouttegd: > right now I am relying a lot on the commandline tool j2 in various pipelines.

Well, maybe it’s time to find something else.

The last release of that tool [happened in 2019](https://pypi.org/project/j2cli/#history), and it is [officially no longer developed](https://github.com/kolypto/j2cli), so it will _never_ support Python 3.12.

(Aside:

> DeprecationWarning: the imp module is deprecated in favour of importlib and 
slated for removal in Python 3.12; see the module's documentation for 
alternative uses

What a surprise, another case of a Python package that is deprecated then removed, but that turns out to still be used by countless of packages… :head_banging_on_a_wall:

)

gouttegd: [Jinjanator](https://github.com/kpfleming/jinjanator) seems to be the only fork that is still active, so that may be your best shot.

Unclear to me whether it maintains compatibility with `j2cli` command line syntax, and therefore if it can be used as a drop-in replacement. If it does, its developers didn't think useful to mention it (which makes me suspect that it doesn't).

matentzn: This (https://github.com/kpfleming/jinjanator) is fine by me, just tried it and works! I will use that for now. Thanks for unearthing!

gouttegd: Ok, I'll update the PR to switch to that tool later today.

Since you tested it, do you know if it accepts the same CLI syntax as `j2`? If so I'd like to install a `j2` alias in the ODK, so as to avoid breaking existing pipelines that invoke `j2`.

matentzn (approved):
After 10 rebuilds and dozens of tests and runs and pipeline upgrades, I am personally convinced everything is ok with this change. I want to note a tiny wrinkle which is a bit un-Damien like, which is coupling the python migration with the image-base migration.

As it is likely that we need 1.5.3 before 1.6, there is a small but not non-existent risk that some interactions between the old base image and the new python packages could cause some problems (unless of course, you would personally judge the base image migration as a "minor" change, since it does not affect the repo setup).

I know you have thought all of this through! Good to go, great work and thank you!

gouttegd:
> I want to note a tiny wrinkle which is a bit un-Damien like, which is coupling the python migration with the image-base migration.

Not sure what is so “un-Damien like” in that. Python 3.12 is the version of Python that is provided with Ubuntu 24.04, while Ubuntu 22.04 (our previous base image) provided Python 3.10. The Python migration from 3.10 to 3.12 is merely a consequence of the base image migration, so of course it belongs to the same PR.

In other words, the base image and the Python version are already coupled by default. To change that, we’d need to actively uncouple them, which would be a completely different matter and would warrant a distinct PR just for that.

Uncoupling could be done by:

* Installing Python 3.10 through third-party package repositories such as [ppa:deadsnakes](https://launchpad.net/~deadsnakes/+archive/ubuntu/ppa). Not keen on that (not keen on using _any_ third-party repository as a general rule).
* Compiling Python 3.10 from source. Not keen on that either. I’ve seen people compiling Python themselves and ending up exposing some subtle bugs in various packages because their Python was built in an environment that didn’t match the expectations of the core Python developers (for example, a bug in the `subprocess` module because Python was built on a system where `/dev/shm` was not mounted at compile time). I’d rather avoid doing that unless there is a _very_ compelling reason to do so.
* Keeping the Python 3.12 from the base system, but create a `venv` or `virtualenv` (or similar) environment where we install Python 3.10. AKA “why going through the hassle of managing ONE version of Python when we could be managing TWO versions instead?”

Overall, I believe uncoupling Python from the base system would be a needless complication. Having to migrate Python (which can include having to tweak the `odk.py` script or replacing some now obsolete packages such as `j2cli`) at the same time than we migrate the base system is but a minor hassle in comparison.

> unless of course, you would personally judge the base image migration as a "minor" change, since it does not affect the repo setup

I’d rather have a 1.6 before we need to make a 1.5.3 release, so as to avoid the question altogether, but it we do need to make a 1.5.3 release, yes, I would consider the base image migration a minor change. It should not impact existing instances that uses the standard workflows. Instances that use custom workflows MAY be impacted (e.g. when a custom workflow is using `j2`), but custom workflows are not in the scope of the “minor releases don’t require updating or tweaking your repository” promise.

matentzn:
> In other words, the base image and the Python version are already coupled by default.

I did not realise this, probably because I didnt read something carefully enough!

> I’d rather have a 1.6 before we need to make a 1.5.3 release, so as to avoid the question altogether, but it we do need to make a 1.5.3 release, yes, I would consider the base image migration a minor change. 

All good then!

Thanks for the clarifications!

PR #1079, opened 2024-06-29
run.sh: Check for spaces in the path to the current directory.

### LINKED ISSUE #1078: Spaces in path causes `docker run` error in `run.sh`
## Overview
I had an ODK repository with spaces in the path (https://github.com/monarch-initiative/mondo-ingest/issues/545), and this resulted in an error when using `run.sh`.

## The error
`docker: invalid reference format`

<details><summary>Log</summary>
<p>

```
joeflack4@MacBook-Pro ontology % sh run.sh make build-mondo-ingest
Running obolibrary/odkfull:dev with '-Xmx20G' as options for ROBOT and other Java-based pipeline steps.
docker: invalid reference format.
See 'docker run --help'.
``

</p>
</details> 

## Replicability
1. Try on an ODK rep with a path that includes a space in it, e.g. this example with dir `mondo builds/`, error:
```
joeflack4@MacBook-Pro ontology % pwd
/Users/joeflack4/Desktop/mondo builds/mondo-ingest-build-ordo-subset/src/ontology
oeflack4@MacBook-Pro ontology % sh run.sh make build-mondo-ingest
Running obolibrary/odkfull:dev with '-Xmx20G' as options for ROBOT and other Java-based pipeline steps.
docker: invalid reference format.
See 'docker run --help'.
```

2. Remove the space, e.g. in this example changing the dir name to `mondo-builds/`, and observe the issue disappear.
```
joeflack4@MacBook-Pro ontology % pwd
/Users/joeflack4/Desktop/mondo-builds/mondo-ingest-build-ordo-subset/src/ontology
joeflack4@MacBook-Pro ontology % sh run.sh make build-mondo-ingest
Running obolibrary/odkfull:dev with '-Xmx20G' as options for ROBOT and other Java-based pipeline steps.
mondo-ingest.Makefile:65: warning: overriding recipe for target 'components/omim.owl'
```

## Additional information
Original context:
- https://github.com/monarch-initiative/mondo-ingest/issues/545

### PR DESCRIPTION
**Pull request by gouttegd**
The shell wrapper script for the ODK is assuming that the directory containing the repository does not contain spaces in its path. This is required for the logic that binds the repository to the `/work` directory inside the container to work, and we may not change that without breaking the requirement that the wrapper script shall be POSIX compliant (supporting space-containing paths would require using array-typed variables, which are not specified by POSIX).

This commit adds an explicit check at the beginning of the wrapper script to ensure that the current working directory does not contain any spaces, and immediately abort if it does.

closes #1078

### CODE REVIEW

**Review by matentzn — APPROVED** [review]
Awesome, thank you! Looks good!

**Comment by gouttegd** [comment]
The fact that the repository must not be at the end of a space-containing path should be documented somewhere. Since `docs/CreatingRepo.md` redirects to https://oboacademy.github.io/obook/tutorial/setting-up-project-odk/, that page in the OBOOK is probably the most appropriate place to do so.

**Comment by matentzn** [comment]
Good idea. Feel free to do it yourself, if you dont have time, just drop an issue and assign to me! Thanks a lot!

**Comment by gouttegd** [comment]
Will do that next week. My work machine is off until Monday. :palm_tree:

File: template/src/ontology/run.sh.jinja2 [hunk 1]
```diff
@@ -16,6 +16,14 @@
 
 set -e
 
+# Check for spaces in the current directory path
+for item in $PWD ; do
+    if [ "$item" != "$PWD" ]; then
```
joeflack4, line 21: I don't get how this fixes the issue, but I trust that it does and I'm just misunderstanding something. Thanks for addressing #1078!

gouttegd: This is a way of checking whether the current directory (as stored in the `PWD` variable) contains spaces.

To illustrate how it works, let’s first suppose that `PWD` does _not_ contain any space (for example, let’s say it contains `/home/alice/myont`). The first line will be expanded as `for item in /home/alice/myont`. Because `/home/alice/myont` cannot be split by the shell (since it contains no spaces), the `for` loop will be executed only once, where the loop variable `item` will be set to `/home/alice/myont`.

Inside the loop, the `"$item" != "$PWD"` condition will be evaluated to _false_ (since `item` contains the same value as `PWD`), so we’ll leave the loop without having done anything.

But suppose that `PWD` contains (at least) one space (for example `/home/alice/my ont`). The first line will be expanded as `for item in /home/alice/my ont`. The shell will split `/home/alice/my ont` in two (notice that `$PWD` is _not_ enclosed in quotes), so the `for` loop will (in principle) be executed twice: first with `item` set to `/home/alice/my`, and then with `item` set to `ont`.

During the first turn in the loop, the `"$item" != "$PWD"` condition will now be evaluated to _true_, so the script will print the error message and terminate.

That’s a slightly convoluted way of doing `if echo $PWD | grep ' ' ; then echo "There are spaces in the current directory path ; fi`, with the advantage that we don’t need to spawn a `grep` process. :)

joeflack4: Wow, thanks for taking the time on the explanation!
I only needed to read up to:
>  The first line will be expanded as `for item in /home/alice/myont`.

I though that `$PWD` would get interpolated as a string, so I thought it would be iterating over characters int hat string.

PR #1081, opened 2024-07-04
Update minor version in Makefile to 1.6

matentzn:
Updating minor version in Makefile, primarily so that my regular `dev` snapshot releases are correctly tagged with 1.6-dev rather than 1.5-dev.


gouttegd (approved):
Usually it’s best to do that right after a release, for the very reason you mention in the message.

Since it was not done right after 1.5, obviously no objection to doing it now.

PR #1084, opened 2024-07-16
Adding = operator to ENV assignments

matentzn:
During the last runs of docker build I got a warning "- LegacyKeyValueFormat: "ENV key=value" should be used instead of legacy "ENV key value" format (line X)", which is due to the fact that the "equal-less form is deprecated and may be removed in the future" (thanks @gouttegd for the note). Here I add equal signs to all such cases across the project.


gouttegd (approved):
Thanks for the meaningful commit message, appreciated.

PR #1089, opened 2024-08-09
Update SSSOM-Java.

gouttegd:
This PR simply bumps the SSSOM-Java (ROBOT plugin and command line tool) to the latest 0.9.0 version.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,7 @@ ENV ODK_VERSION=$ODK_VERSION
 # Software versions
 ENV JENA_VERSION=4.9.0
 ENV KGCL_JAVA_VERSION=0.4.0
-ENV SSSOM_JAVA_VERSION=0.7.7
+ENV SSSOM_JAVA_VERSION=0.9.0
 
 # Avoid repeated downloads of script dependencies by mounting the local coursier cache:
 # docker run -v $HOME/.coursier/cache/v1:/tools/.coursier-cache ...


gouttegd:
The CI check failure has nothing to do with SSSOM and is instead caused by this incomprehensible bug when trying to install the Python packages:

```
ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/topics/dependency-resolution/#dealing-with-dependency-conflicts

The conflict is caused by:
    ipykernel 6.29.5 depends on debugpy>=1.6.5
    The user requested (constraint) debugpy==1.8.3
```

where Python seems to consider that `debugpy==1.8.3` (which we forcibly require) does _not_ satisfy `ipykernel`’s dependency on `debugpy>=1.6.5`. Because Python.

In local testing, this is “fixed” by bumping to `debugpy==1.8.5`, so I would expect the bug to automatically go away with the next “constraints update” on Monday.

gouttegd:
So the issue above is caused by `debugpy-1.8.3` having been simply deleted from PyPI. Kudos to PIP for not being able to detect that case and hallucinating a bogus error message about “conflicting dependencies” instead.

In that case it should indeed automatically be fixed at the next “constraints update” GH action, which should automatically pick up the new 1.8.5 version. No need to fix it now.

matentzn (approved):
Thank you!

PR #1090, opened 2024-08-11
Remove default user

gouttegd:
The Ubuntu 24.04 base image that we now use contains a predefined `ubuntu` user account with a UID of 1000. This could cause some confusion with the ODK’s own user account (`odkuser`), which by default also has a UID of 1000.

There is no need for that `ubuntu` account, so this PR simply removes it.

While we are at it, this PR also refactors the last section of the ODKLite `Dockerfile` (where we add the instruction to remove the `ubuntu` user account) to group several operations together and reduce slightly the number of layers in the ODKLite image.

### DIFF
--- a/docker/odklite/Dockerfile
+++ b/docker/odklite/Dockerfile
@@ -86,6 +86,8 @@ RUN chmod 755 /usr/local/sbin/odkuser-entrypoint.sh
 ENTRYPOINT ["/usr/local/sbin/odkuser-entrypoint.sh"]
 # secure_path is enabled by default in Debian/Ubuntu, disable it
 RUN sed -i '/secure_path/d' /etc/sudoers
+# Remove default user account
+RUN userdel -r ubuntu
 
 # Install a script that provides information about the ODK and its tools
 COPY scripts/odk-info.sh /tools/odk-info
--- a/docker/odklite/Dockerfile
+++ b/docker/odklite/Dockerfile
@@ -74,7 +74,7 @@ RUN wget -nv https://github.com/lihaoyi/Ammonite/releases/download/$AMMONITE_VER
     java -cp /tools/amm ammonite.AmmoniteMain /dev/null
 
 # Install RDF/XML validation script
-COPY scripts/check-rdfxml.sh /tools/check-rdfxml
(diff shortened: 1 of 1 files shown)


gouttegd:
CI check explicitly cancelled since it is bound to fail until the `debugpy` issue is fixed (which should happen the next time the “Update Python constraints” workflow is run).

matentzn:
> which should happen the next time the “Update Python constraints” workflow is run

FYI as admin of ODK you should feel free to simply run [the action](https://github.com/INCATools/ontology-development-kit/actions/workflows/constraints.yml) and merge the PR after an eyeball review (which only does one thing: checking that no tool decreased in version number, of if it does, it can be explained and dismissed).

gouttegd:
> FYI as admin of ODK you should feel free to simply run the action

Sure, but in this case, with the next automatic run scheduled to happen tomorrow, there was no real need to trigger an out-of-cycle run.

matentzn (approved):
I tested it with one ontology and I cant see any difference, so I am assuming its all good.

I trust you have, as usual, weighed the risks if there are any; I can only do a LGTM review here as I dont understand the consequences of deleting the default user. :)

gouttegd:
Deleting the default user should have no consequences at all.

On the other hand, leaving the default user in place (that is, without this PR) _could_ lead to confusion and possibly bugs (though I have not so far observed any), because we end up with two entries in `/etc/passwd` with the same UID but different home directories:

```
ubuntu:x:1000:1000:Ubuntu:/home/ubuntu:/bin/bash
odkuser:x:1000:1000:ODK User:/home/odkuser:/bin/bash
```

For standard workflows, this should not cause any issues because the standard workflows never refer to the user’s home directory anyway.

But if you have a custom workflow that somehow needs to refer to the home directory (for example, if you are use `runoak`, which may cache some data under `$HOME/.data`), then you _may_ run into some troubles if two different programs in the workflow don’t use the same logic to pick up the home directory from the conflicting entries in `/etc/passwd` – that is, if one program believes the correct home directory is `/home/ubuntu` while another one believes that it is `/home/odkuser`.

Again, I have not observed any such trouble yet, but the mere fact that the possibility exists is reason enough to remove the (useless) default user.

PR #1096, opened 2024-08-23
Make sure the SSH socket belongs to `odkuser`.

gouttegd:
When we run under a non-privileged user account (which is the default situation), we need to make the SSH authentication socket, if it exists, belongs to the `odkuser` account. Under some conditions, Docker may create that socket as belonging to root by default.


matentzn (approved):
I agree with the motivation, but cannot judge the solution. Howeber I have known you long enough to trust that it works as you intend.

PR #1108, opened 2024-10-18
Write $(SRCMERGED) in OFN format.

### LINKED ISSUE #1107: Failure of `-obo-report.tsv` induced by OBO serialisation quirks
This problem was encountered by Uberon on [this PR](https://github.com/obophenotype/uberon/pull/3374); see the CI failure report [here](https://github.com/obophenotype/uberon/actions/runs/11393321033/job/31701344085).

The main change in that PR is the addition of a new term that is disjoint from two imported GO terms:

```
[Term]
id: UBERON:0000001 ! gross anatomical part
name: gross anatomical part
disjoint_from: GO:0005623 ! cell
disjoint_from: GO:0110165 ! cellular anatomical entity
```

(Other tags in that frame removed for clarity, they don’t matter here. What matter are the two `disjoint_from` tags.)

At some point in the standard pipeline, the `-edit` file is processed to generate what we call `$(SRCMERGED)`. Basically, `imports` statement are removed and components and pattern-derived axioms are merged. Importantly, `$(SRCMERGED)` is in the same format as the edit file, so if the ontology is maintained in the OBO format (as is the case for Uberon), then `$(SRCMERGED)` is also written in the OBO format. `$(SRCMERGED)` is then used as input to `robot report` to generate the `reports/uberon-edit.obo-obo-report.tsv` report, which is the step that failed in the above PR.

The problem is that, when writing `$(SRCMERGED)`, the disjointness axioms that, in the -edit file, are part of the `UBERON:0000001` frame, are now written as if they were part of the frames of the referenced GO terms. That is, the fragment above becomes:

```
[Term]
id: GO:0005623
disjoint_from: UBERON:0000001 ! gross anatomical part

[Term]
id: GO:0110165
disjoint_from: UBERON:0000001 ! gross anatomical part

[Term]
id: UBERON:0000001
name: gross anatomical part
```

(I have not looked at the logic used by the OWLAPI to decide in which frame a `disjoint_from` tag should go, but I suspect the decision is made by lexicographically sorting the IRIs involved in the axiom, and putting the tag in the frame corresponding to the first IRI. In this example, `GO:0005623` would be sorted before `UBERON:0000001`, so the OWLAPI creates a frame for `GO:0005623` instead of writing the `disjoint_from` tag in the frame for `UBERON:0000001`.)

Now, _logically_, this does not change anything. _A DisjointWith: B_ is obviously the same as _B DisjointWith: A_. **But** the second form makes the `missing_label` test (part of the standard ROBOT report check) fail, because the `GO:0005623` and `GO:0110165` frames don’t have a label!

gouttegd:
Make sure the `$(SRCMERGED)` intermediate file is always written in OFN format, rather than in the same format than the -edit file.

If `$(SRCMERGED)` is written in OBO format, this can cause some checks made on that file to fail because of some peculiarities of the OBO serialisation that do not accurately reflect the contents of the original -edit file.

closes #1107

### DIFF
--- a/template/src/ontology/Makefile.jinja2
+++ b/template/src/ontology/Makefile.jinja2
@@ -196,7 +196,7 @@ $(ROBOT_PLUGINS_DIRECTORY)/{{ plugin.name }}.jar:
 MAIN_PRODUCTS = $(sort $(foreach r,$(RELEASE_ARTEFACTS), $(r)) $(ONT))
 MAIN_GZIPPED  = {% if project.gzip_main %}$(foreach f,$(FORMATS), $(ONT).$(f).gz){% endif %}
 MAIN_FILES    = $(foreach n,$(MAIN_PRODUCTS), $(foreach f,$(FORMATS), $(n).$(f))) $(MAIN_GZIPPED)
-SRCMERGED     = $(TMPDIR)/merged-$(SRC)
+SRCMERGED     = $(TMPDIR)/merged-$(ONT)-edit.ofn
 
 .PHONY: all_main
 all_main: $(MAIN_FILES)


gouttegd:
@matentzn Not going to merge this one without your input on it, even if it has to wait.

template/src/ontology/Makefile.jinja2
```diff
@@ -196,7 +196,7 @@ $(ROBOT_PLUGINS_DIRECTORY)/{{ plugin.name }}.jar:
 MAIN_PRODUCTS = $(sort $(foreach r,$(RELEASE_ARTEFACTS), $(r)) $(ONT))
 MAIN_GZIPPED  = {% if project.gzip_main %}$(foreach f,$(FORMATS), $(ONT).$(f).gz){% endif %}
 MAIN_FILES    = $(foreach n,$(MAIN_PRODUCTS), $(foreach f,$(FORMATS), $(n).$(f))) $(MAIN_GZIPPED)
-SRCMERGED     = $(TMPDIR)/merged-$(SRC)
+SRCMERGED     = $(TMPDIR)/merged-$(ONT)-edit.ofn
```
matentzn, line 199: Not sure how this change works:

```
$(SRCMERGED): $(EDIT_PREPROCESSED) $(OTHER_SRC)
	$(ROBOT) remove --input $< --select imports --trim false \
		merge  $(patsubst %, -i %, $(OTHER_SRC)) -o $@

$(EDIT_PREPROCESSED): $(SRC)
	$(ROBOT) convert --input $< --format ofn --output $@
```

EDIT_PREPROCESSED by default always converts to functional, regardless of the file ending, and the way SRCMERGED is defined, it should not really change the serialisation unless someone manually overwrites the goal (I mean, there is not `convert` step in there). Dont you have to at least at a convert -f ofn clause into the SRCMERGED goal?

gouttegd: Notice how the command that produces `$(SRCMERGED)` does not have an explicit `convert -f...` bit. Therefore, the format of the resulting file is solely determined by the extension of the file given to the last `-o` option.

With `$(SRCMERGED)` defined as `$(TMPDIR)/merged-$(SRC)`, that extension will be the same as the extension of the -edit file. So if the -edit file is in OBO, `$(SRCMERGED)` is in OBO as well -- regardless of whatever format is used for `$(EDIT_PREPROCESSED).

gouttegd: For the avoidance of doubt: ROBOT does _not_ automatically preserve the input format. Just because the input file of a pipeline is in OFN, it doesn’t mean the output file will be in OFN as well without an explicit `convert`. The format of the output file is always determined by the extension found in the argument to the `-o` option (or of course by the `-f` option, in the specific case of the `convert` command`).

matentzn: > For the avoidance of doubt: ROBOT does not automatically preserve the input format. Just because the input file of a pipeline is in OFN, it doesn’t mean the output file will be in OFN as well without an explicit convert. The format of the output file is always determined by the extension found in the argument to the -o option (or of course by the -f option, in the specific case of the convert command).

lol I didn't know this! I guess I never paused to think this might be the case.

matentzn (approved):
makes sense, this is in the spirit of what we did for EDITPREPROCESSED

PR #1109, opened 2024-10-22
Allow sharing a local OAK cache with the container

### LINKED ISSUE #1051: Facilitate or at least document how to share the OAK cache
The [Ontology Access Kit](https://incatools.github.io/ontology-access-kit/index.html) (OAK, aka `oaklib` from Python’s point of view, aka `runoak` from the command line’s point of view) is one of the tools/libraries provided by the ODK.

In fact the ODK is supposedly one of the easiest way for “non-technical” users to get access to OAK, because installing Python programs is still too difficult for many people.

When OAK is used to access online resources (for example with `-i sqlite:obo:uberon`, which accesses a pre-built SQLite version of Uberon), it attempts to cache a copy of those resources in the local filesystem, to avoid always re-downloading them upon each new call. The default location for the cache is `~/.data`, or the value of the `PYSTOW_HOME` environment variable if such a variable is set.

(As an aside, defaulting to such a generic name under the user’s home directory is a terrible move, but that’s a [deliberate decision](https://github.com/cthoyt/pystow/issues/10) that’s unlikely to ever change.)

Now when OAK is used from the ODK, the `~/.data` directory is within the Docker container. So any file that OAK is storing there will only exist for as long as the container itself exists. That means that when people are running several OAK commands like this:

```sh
sh run.sh runoak -i sqlite:obo:uberon command1 ...
sh run.sh runoak -i sqlite:obo:uberon command2 ...
sh run.sh runoak -i sqlite:obo:uberon command3 ...
```

none of these commands will benefit from the cache. They will all download a fresh copy of Uberon.

One workaround is of course to run a _shell_ within a container, instead of running `runoak` directly, and to then invoke `runoak` from that shell:

```sh
sh run.sh bash
odkuser@abe8c94b5e84:/work/src/ontology$ runoak -i sqlite:obo:uberon command1 ...
odkuser@abe8c94b5e84:/work/src/ontology$ runoak -i sqlite:obo:uberon command2 ...
odkuser@abe8c94b5e84:/work/src/ontology$ runoak -i sqlite:obo:uberon command3 ...
```

But that is not really a satisfying solution as it will still lead to Uberon being re-downloaded every time the user has to start working with it, even if maybe they already downloaded it the day before.

It is possible to configure the ODK to make the local cache visible from the container by “binding” the `~/.data` directory from the local filesystem to the `/home/odkuser/.data` directory within the container, by adding the following in the `src/ontology/run.sh.conf` file:

```
ODK_BINDS=~/.data:/home/odkuser/.data
```

(This will only work once #1050 will have been fixed.)

Another solution would be set the `PYSTOW_HOME` variable to a directory _within_ the repository (most likely somewhere under `src/ontology/tmp`), which is already bound to a mount point within the container. That would at least allow sharing the cache between ODK/OAK invocations that are run from within the same repository.

At the very least, the ODK should provide documentation on how to do that.

Should the ODK try to do that *automatically*? I am on the fence here. On one side, it’d be nice for users if the OAK cache could work “out of the box” without any extra configuration. On the other side, the ODK container is supposed to shield the local filesystem (except the actual repository) from any side-effects – everything that happens in the container stays in the container –, so it may not be a good idea to silently break a hole through the container’s wall: what if an interrupted download corrupts the cache? Users could expect that it would have no consequence, since the command was run inside a container – except that no, actually the cache is outside the container, so you’ve just corrupted your actual cache, oops!

Thoughts?

gouttegd:
This PR facilitates the sharing of a local OAK cache directory (the directory where OAK stores, among other things, the downloaded SQLiteDB files) with the ODK container.

First, it amends the entry point script to ensure that the `/home/odkuser/.data/oaklib` directory always belong to the `odkuser`. This allows to bind a local OAK cache to that directory (where it will be used by default by any OAK process started from within the container) while making sure the in-container directory is writable by the `odkuser` user account.

Second, it amends the `run.sh` script to recognise a `ODK_SHARE_OAK_CACHE` variable, which is supposed to point to the OAK cache directory the user wants to share with the container. That directory is automatically bound to the `/home/odkuser/.data/oaklib` directory within the container (or `/root/.data/oaklib`, if we run as the root user).

For convenience, `ODK_SHARE_OAK_CACHE=user` is interpreted as `ODK_SHARE_OAK_CACHE=~/.data/oaklib` (automatically sharing the user’s default OAK cache), and `ODK_SHARE_OAK_CACHE_repo` is interpreted as `ODK_SHARE_OAK_CACHE=$PWD/tmp/oaklib` (using a directory within the repo as the cache; for users who would want to share the cache across multiple ODK invocations in the same repo, but without touching their own cache in `~/.data/oaklib`).

closes #1051


template/src/ontology/run.sh.jinja2
```diff
@@ -90,6 +90,22 @@ rm -f tmp/debug.log
 VOLUME_BIND=$PWD/../../:/work$ODK_SSH_BIND
 WORK_DIR=/work/src/ontology
 
+# Support for OAK cache sharing
+if [ -n "$ODK_SHARE_OAK_CACHE" ]; then
+    case "$ODK_SHARE_OAK_CACHE" in
+    user)
+        # We assume the cache is in its default location; if it is not,
+        # ODK_SHARE_OAK_CACHE must point to the actual location.
+        ODK_SHARE_OAK_CACHE="$HOME/.data/oaklib"
```
gouttegd, line 99: We _could_ in fact not assume anything, and automatically find the user’s cache no matter where it is located (that’s what the ODKRunner does).

But it would be quite longer (not really complex, just longer), because there are many ways to change the default location of the cache and we’d need to check them all.

This would look like this:

```sh
if [ -n "$ODK_SHARE_OAK_CACHE" ]; then
    case "$ODK_SHARE_OAK_CACHE" in
    user)
        if [ -n "$OAKLIB_HOME" ]; then
            ODK_SHARE_OAK_CACHE="$OAKLIB_HOME"
        elif [ -n "$PYSTOW_HOME" ]; then
            ODK_SHARE_OAK_CACHE="$PYSTOW_HOME/oaklib"
        elif [ -n "$PYSTOW_USE_APPDIRS" ]; then
            if [ -n "$XDG_DATA_HOME" ]; then
                ODK_SHARE_OAK_CACHE="$XDG_DATA_HOME/oaklib"
            elif [ -d "$HOME/Library/Application Support" ]; then
                ODK_SHARE_OAK_CACHE="$HOME/Library/Application Support/oaklib";
            else
                ODK_SHARE_OAK_CACHE="$HOME/.local/share/oaklib"
            fi
        else
            ODK_SHARE_OAK_CACHE="$HOME/${PYSTOW_NAME:-.data}/oaklib"
        fi
        ;;
```

I am not sure whether it brings enough benefits to justify making the `run.sh` script so long.

scripts/entrypoint.sh
```diff
@@ -8,7 +8,11 @@ REAL_USER_ID=${ODK_USER_ID:-1000}
 REAL_GROUP_ID=${ODK_GROUP_ID:-1000}
 
 groupadd -o -g $REAL_GROUP_ID odkuser
-useradd -o -s /bin/bash -u $REAL_USER_ID -g $REAL_GROUP_ID -c "ODK User" -m odkuser
+useradd -o -s /bin/bash -u $REAL_USER_ID -g $REAL_GROUP_ID -c "ODK User" -M odkuser
+mkdir -p /home/odkuser
+chown odkuser:odkuser /home/odkuser
+[ -d /home/odkuser/.data ] && chown odkuser:odkuser /home/odkuser/.data
```
matentzn, line 14: I will say something super stupid now, please dismiss gently without too much prejudice:

If we bind the cache dir as volume to the container run, can the chmod change the user permissions on the cache directory on the host in a way that may adversely affect accessing the cache outside the container? 

Sorry, you deserve more expert reviewers.

template/src/ontology/run.sh.jinja2
matentzn: What is your concern with the length of the script: Unless the code introduces syntax that might cause some kinds of shells to choke (i.e. non POSIX) I don't see run.sh length as much of an issue anymore.

matentzn (approved):
I dont see anything wrong here. I did not run the code, but I trust you did!

scripts/entrypoint.sh
gouttegd: > can the chmod change the user permissions on the cache directory on the host

Not at all (but it’s not a stupid question).

When a directory from the host (let’s say `/home/damien/.data/oaklib`) is bound to a directory in the container (let’s say `/home/odkuser/.data/oaklib`), only the _last_ component of the host directory path is visible from the container (in this case, the `oaklib` directory).

Basically, the `oaklib` directory is the ”cut-off” point between the container and the host. Anything above it only exists in the container, while everything below it comes straight from the host:

```
/home/odkuser/.data/oaklib/something/in/my/cache
\_________________/ \__________________________/
  ONLY EXISTS IN      EXISTS IN THE HOST
  THE CONTAINER
```

Another way to put it is that, just because `/home/damien/.data/oaklib` on the host is bound to `/home/odkuser/.data/oaklib` in the container, it doesn’t mean that `/home/damien/.data` is bound to `/home/odkuser/.data`.

So, any `chmod` made on `/home/odkuser` or `/home/odkuser/.data` will have no effect on `/home/damien/.data/oaklib`.

Of note, this concern about not “adversely affecting the cache outside of the container” is precisely the reason for calling `chmod` twice, first on `/home/odkuser`, then on `/home/odkuser/.data`. We could instead have called `chmod` in recursive mode on `/home/odkuser` (`chmod -R odkuser:odkuser /home/odkuser`), which would have automatically taken care of `/home/odkuser/.data`. **But** this would also have impacted any files and directories below the “cut-off” point (files and directories from the host), which we want to avoid.

template/src/ontology/run.sh.jinja2
gouttegd: I am simply concerned that this script is turning into a huge piece of bloat, that is becoming harder and harder to maintain (for us) and to understand (for everybody else).

scripts/entrypoint.sh
matentzn: Awesome this makes total sense! Thank you!

PR #1111, opened 2024-10-23
Automatically set Java -Xmx to 90% of Docker memory.

### LINKED ISSUE #1074: Unifying max memory settings between Docker and ROBOT
Currently, users working with large ontologies might need to increase two different memory settings:

* the maximal amount of memory that a Docker container is allowed to use (typically configured in the configuration panel of _Docker Desktop_);
* the maximal amount of memory that ROBOT is allowed to use when running within a ODK container (typically configured with the `robot_java_args` option in the `-odk.yaml` configuration file).

Changing only one of those settings is not enough: allowing Docker to use 40GB of RAM will be of little help if ROBOT is still limited to use only 8GB (this will only help for ODK workflow steps that do not use ROBOT, e.g. OAK-dependent steps), and conversely allowing ROBOT to use 40GB of RAM will have no effect if the container itself is limited to a smaller amount.

It would be nice if users only had to perform _one_ configuration step (changing just _one_ setting) and not have to worry about making sure that all components of the ODK can benefit from the new setting.

One way to achieve that would be to make the default ROBOT setting dependent on the Docker setting: that is, the `run.sh` wrapper script should detect the maximal amount of memory that Docker is configured to use, and set the ROBOT Java options accordingly. E.g., if Docker is configured to use at most 40GB of RAM, we can allow ROBOT to use, say, 36GB (90% of the total amount of memory allocated to the container).

One way to get the Docker setting from the wrapper script:

```sh
docker system info | sed -nre 's,^ Total Memory: ,,p'
```

gouttegd:
When seeding/updating a repo, if the project does not explicitly set any `robot_java_args`, generate code in the run.sh script to automatically compute a default max memory setting that is dependent on the total amount of memory that Docker is allowed to use.

The code is used if (1) we are indeed using Docker (and not Singularity) and (2) no contradicting option is set in ODK_JAVA_OPTS (either in the environment or in a run.sh.conf file).

closes #1074

### DIFF
--- a/template/src/ontology/run.sh.jinja2
+++ b/template/src/ontology/run.sh.jinja2
@@ -63,7 +63,14 @@ if [ -n "$TAG_IN_IMAGE" ]; then
   ODK_IMAGE=$(echo $ODK_IMAGE | awk -F':' '{ print $1 }')
 fi
 ODK_TAG=${ODK_TAG:-latest}
-ODK_JAVA_OPTS={% if project.robot_java_args is defined and project.robot_java_args|length %}${ODK_JAVA_OPTS:-{{ project.robot_java_args }}}{% else %}${ODK_JAVA_OPTS:--Xmx8G}{% endif %}
+{% if project.robot_java_args is defined and project.robot_java_args|length %}
+ODK_JAVA_OPTS=${ODK_JAVA_OPTS:-{{ project.robot_java_args }}}
+{% else %}
+if [ -z "$USE_SINGULARITY" ]; then
+    {% raw %}DEFAULT_MAX_MEM=$(bc <<<"($(docker info --format={{.MemTotal}}) * .9) / (1024*1024*1024)")G{% endraw %}
+fi
+ODK_JAVA_OPTS=${ODK_JAVA_OPTS:--Xmx$DEFAULT_MAX_MEM}
+{% endif %}
 ODK_DEBUG=${ODK_DEBUG:-no}
 
 ODK_USER_ID=${ODK_USER_ID:-{% if project.run_as_root %}0{% else %}$(id -u){% endif %}}
--- a/template/src/ontology/run.sh.jinja2
+++ b/template/src/ontology/run.sh.jinja2
@@ -68,6 +68,8 @@ ODK_JAVA_OPTS=${ODK_JAVA_OPTS:-{{ project.robot_java_args }}}
 {% else %}
 if [ -z "$USE_SINGULARITY" ]; then
     {% raw %}DEFAULT_MAX_MEM=$(bc <<<"($(docker info --format={{.MemTotal}}) * .9) / (1024*1024*1024)")G{% endraw %}
+else
+    DEFAULT_MAX_MEM=8G
 fi
 ODK_JAVA_OPTS=${ODK_JAVA_OPTS:--Xmx$DEFAULT_MAX_MEM}
 {% endif %}


template/src/ontology/run.sh.jinja2
```diff
@@ -63,7 +63,14 @@ if [ -n "$TAG_IN_IMAGE" ]; then
   ODK_IMAGE=$(echo $ODK_IMAGE | awk -F':' '{ print $1 }')
 fi
 ODK_TAG=${ODK_TAG:-latest}
-ODK_JAVA_OPTS={% if project.robot_java_args is defined and project.robot_java_args|length %}${ODK_JAVA_OPTS:-{{ project.robot_java_args }}}{% else %}${ODK_JAVA_OPTS:--Xmx8G}{% endif %}
+{% if project.robot_java_args is defined and project.robot_java_args|length %}
```
matentzn, line 66: Never realised that `project.robot_java_args|length` condition is necessary to catch the case that someone sets the key to an empty value.. Makes sense though.

template/src/ontology/run.sh.jinja2
```diff
@@ -63,7 +63,14 @@ if [ -n "$TAG_IN_IMAGE" ]; then
...
+ODK_JAVA_OPTS=${ODK_JAVA_OPTS:-{{ project.robot_java_args }}}
+{% else %}
+if [ -z "$USE_SINGULARITY" ]; then
+    {% raw %}DEFAULT_MAX_MEM=$(bc <<<"($(docker info --format={{.MemTotal}}) * .9) / (1024*1024*1024)")G{% endraw %}
```
matentzn, line 70: I have not reviewed this, assuming this was tested.

template/src/ontology/run.sh.jinja2
```diff
@@ -63,7 +63,14 @@ if [ -n "$TAG_IN_IMAGE" ]; then
...
+fi
+ODK_JAVA_OPTS=${ODK_JAVA_OPTS:--Xmx$DEFAULT_MAX_MEM}
```
matentzn, line 72: Where is this DEFAULT_MAX_MEM coming from?

template/src/ontology/run.sh.jinja2
gouttegd: Tested on both my GNU/Linux and macOS machines.

(And before you ask: `bc` is not a GNUism, it’s a POSIX-specified command. :) )

Of note, this returns a value in gigabytes, _rounded to the lower gigabyte_. This means that if the maximum amount of memory allocated to Docker is barely above 1G (so that 90% of it is less than 1G), this will return `0G`.

I considered adding an explicit test for that condition, but decided it was not worth it. I don’t expect anyone to run the ODK with ~1G of memory anyway (not even sure it _can_ run with 1G) -- and if someone _really_ wants to try that, they can always explicitly set ODK_JAVA_OPTS.

template/src/ontology/run.sh.jinja2
gouttegd: Oops. It should have been set to the old default (`8G`) in the case where `USE_SINGULARITY` is set, my bad.

template/src/ontology/run.sh.jinja2
matentzn: Super, thanks for the explanation!

matentzn (approved):
Looks good, thank you!

joeflack4:
Glad to see this!

Issue #1105, opened 2024-10-17
Seeding process failing under macOS because of 'dubious ownership in repository' 


gouttegd:
While trying to replicate the issue #1104, I came across another problem: Trying to seed a repository under macOS (using the standard `sh seed-via-docker.sh -C myont-odk.yaml` seeding command) fails with the following error:

```
fatal: detected dubious ownership in repository at '/work/target/myont'
To add an exception for this directory, call:

	git config --global --add safe.directory /work/target/myont

Traceback (most recent call last):
  File "/tools/odk.py", line 1115, in <module>
    cli()
  File "/usr/local/lib/python3.10/dist-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.10/dist-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python3.10/dist-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python3.10/dist-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
  File "/tools/odk.py", line 1079, in seed
    runcmd("cd {dir}/src/ontology && make && git commit -m 'initial commit' -a && git branch -M {branch} && make prepare_initial_release && git commit -m 'first release'".format(dir=outdir, branch=project.git_main_branch))
  File "/tools/odk.py", line 1109, in runcmd
    raise Exception('Failed: {}'.format(cmd))
Exception: Failed: cd target/myont/src/ontology && make && git commit -m 'initial commit' -a && git branch -M main && make prepare_initial_release && git commit -m 'first release'
```

Interestingly, this happens on macOS (OS 15.01, Docker Desktop 4.34.3), but _not_ on GNU/Linux (Docker 27.3.1). This may explain why the problem was never caught before by our CI test suite, since said test suite runs on Ubuntu.

The problem lies [here](https://github.com/INCATools/ontology-development-kit/blob/ac92ce332125e4a585b72eb031bffc711cc005f1/docker/odklite/Dockerfile#L80):

```
# Force Git to accept working on a repository owned by someone else
RUN echo "[safe]\n    directory = /work" >> /etc/gitconfig
```

[...]

PR #1097, opened 2024-08-23
Make `odk.py seed` more robust

gouttegd:
The `seed` command of the `odk.py` script is quite fragile, in that it will easily and loudly crash out when it used incorrectly.

A) Called with no positional arguments (`odk.py seed`):

This causes a crash because the `repo` variable is then set to `None`, while the Jinja template expects a string (or at least something that has a `upper()` method):

```
Traceback (most recent call last):
  File "/tools/odk.py", line 1112, in <module>
    cli()
  File "/usr/local/lib/python3.12/dist-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tools/odk.py", line 1041, in seed
    tgts += unpack_files(tdir, mg.generate(srcf))
                               ^^^^^^^^^^^^^^^^^
  File "/tools/odk.py", line 800, in generate
    return template.render( project = self.context.project, env = {"ODK_VERSION": os.getenv("ODK_VERSION")})
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/jinja2/environment.py", line 1304, in render
    self.environment.handle_exception()
  File "/usr/local/lib/python3.12/dist-packages/jinja2/environment.py", line 939, in handle_exception
    raise rewrite_traceback_stack(source=source)
[...]

PR #1113, opened 2024-10-28
Prepare ODK 1.5.3

gouttegd:
This PR updates the `BRANCH-1.5-MAINTENANCE` branch to make it ready for a 1.5.3 point release.

It back ports several updates to the ODK tooling, including the update of the base image to Ubuntu 24.04 (which, as discussed [here](https://github.com/INCATools/ontology-development-kit/pull/1076#issuecomment-2259986362), is considered a _minor_ change, on the ground that it does not affect the standard ODK-generated workflows -- it does mean that we have to backport the J2cli -> Jinjanator replacement though, which in turns means that any custom workflow that uses `j2` will not work with the new image until the workflow is amended to use `jinjanate` instead).

It also backports a few fixes that only concern the image and not the workflows (#1096, #1097, #1105).


matentzn (approved):
Looks good, shall I make a release?

matentzn:
(or can you, I forgot)

gouttegd:
My humble desktop machine would rather have you do it. :D 

FYI, I’ve tested a locally built 1.5.3 this afternoon on both Uberon and FBbt without issue.

PR #1115, opened 2024-10-30
Relax SubClassOf axioms.

### LINKED ISSUE #1114: Relax SubClassOf axioms by default
ROBOT’s `relax` command has a [`--include-subclass-of` option](https://robot.obolibrary.org/relax#include-subclass-of) to relax _SubClassOf_ axioms where the superclass is an intersection, e.g. to turn

> CL:0000104 **SubClassOf:** CL:0000540 _and_ (RO:0000053 _some_ PATO:0070026)

into

> CL:0000104 **SubClassOf:** CL:0000540
> CL:0000104 **SubClassOf:** RO:0000053 _some_ PATO:0070026

That option should be enabled by default in the ODK-generated standard pipelines, given that (1) the two forms are exactly equivalent (contrary to the relaxing of **EquivalentTo** axioms, where the relaxed form is a weaker assertion), and (2) the non-relaxed form is known to cause downstream issues, for example when converting to the OboGraphs JSON or OBO formats.

gouttegd:
When running `robot relax`, enable the `--include-subclass-of` option, so that SubClassOf axioms where the superclass is an intersection are relaxed into equivalent separate simple SubClassOf axioms.

That behaviour is _not_ configurable (no option at the ODK config level), as it is deemed unnecessary: relaxing those axioms should be uncontroversial.

In the event that someone would want to restore the previous behaviour of _not_ relaxing SubClassOf axioms, they can do so by overriding the `RELAX_OPTIONS` variable in their custom Makefile. That should be enough.

closes #1114

### DIFF
--- a/template/src/ontology/Makefile.jinja2
+++ b/template/src/ontology/Makefile.jinja2
@@ -59,6 +59,7 @@ OBO_FORMAT_OPTIONS =        {{ project.obo_format_options }}
 SPARQL_VALIDATION_CHECKS =  {% for x in project.robot_report.custom_sparql_checks|default(['owldef-self-reference', 'iri-range', 'label-with-iri', 'multiple-replaced_by']) %}{{ x }} {% endfor %}
 SPARQL_EXPORTS =            {% for x in project.robot_report.custom_sparql_exports|default(['basic-report', 'class-count-by-prefix', 'edges', 'xrefs', 'obsoletes', 'synonyms']) %}{{ x }} {% endfor %}
 ODK_VERSION_MAKEFILE =      {% if env is defined %}{{env['ODK_VERSION'] or "Unknown" }}{% else %}"Unknown"{% endif %}
+RELAX_OPTIONS =             --include-subclass-of true
 
 TODAY ?=                    $(shell date +%Y-%m-%d)
 OBODATE ?=                  $(shell date +'%d:%m:%Y %H:%M')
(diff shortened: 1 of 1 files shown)


template/src/ontology/Makefile.jinja2
```diff
@@ -1130,7 +1131,7 @@ $(ONT)-non-classified.owl: $(EDIT_PREPROCESSED) $(OTHER_SRC) $(IMPORT_FILES)
 $(ONT)-simple.owl: $(EDIT_PREPROCESSED) $(OTHER_SRC) $(SIMPLESEED) $(IMPORT_FILES)
 	$(ROBOT_RELEASE_IMPORT_MODE) \{% if project.release_use_reasoner %}
 		reason --reasoner {{ project.reasoner }} --equivalent-classes-allowed {{ project.allow_equivalents }} --exclude-tautologies {{ project.exclude_tautologies }} --annotate-inferred-axioms {{ project.release_annotate_inferred_axioms|default('false')|lower }} \{% endif %}
-		relax \
+		relax $(RELAX_OPTIONS) \
 		remove --axioms equivalent \
 		relax \
```
gouttegd, line 1136: Unrelated to this issue: Not sure why this second `relax` command is there? What could it relax that has not already been relaxed by the first command?

matentzn: I made this comment in a separate context; I cant imagine it doing anything at all! `reduce` makes sense, a second relax does not.

matentzn (approved):
This is fine, if people dont want the relax option they can just reset the variable outside make, so there is a workaround.

PR #1116, opened 2024-10-30
Fix publishing of SSSOM mapping sets.

### LINKED ISSUE #1036: Publishing SSSOM mapping sets is broken
Trying to get the ODK to publish the SSSOM sets that it manages, by adding the `release_mappings: True` option to the `sssom_mappingset_group` section in the config file, leads to the following error when running `make prepare_release`:

```sh
rsync -R   sssomo.owl   sssomo-base.owl   sssomo-full.owl    ../mappings/modmini-onta.sssom.tsv  ../mappings/modmini-ontb.sssom.tsv  ../mappings/modmini-ontc.sssom.tsv  ../mappings/modmini-ontd.sssom.tsv ../.. &&\
rm -f   sssomo.owl   sssomo-base.owl   sssomo-full.owl  tmp/merged-sssomo-edit.owl tmp/sssomo-preprocess.owl &&\
echo "Release files are now in ../.. - now you should commit, push and make a release \
        on your git hosting site such as GitHub or GitLab"
found ".." dir in relative path: ../mappings/modmini-onta.sssom.tsv
rsync error: syntax or usage error (code 1) at flist.c(2358) [sender=3.2.7]
rsync error: received SIGUSR1 (code 19) at main.c(1613) [Receiver=3.2.7]
make: *** [Makefile:300: prepare_release] Error 1
```

This is because the mappings are in `src/mappings`, while the `rsync` command that does the bulk of copying the release assets to the top-level directory is run from within the `src/ontology` directory. As a result, we are ultimately trying to do `rsync -R [other release files] ../mappings/<mappingset>.sssom.tsv ../..`, which `rsync` does not like at all – the `-R` option is incompatible with some source files being in any directory _above_ the current directory.

Possible solutions:

1) Moving the mappings directory from `src/mappings` to `src/ontology/mappings`. Easiest solution, but a breaking change. And whether it is desirable to have the mappings in `src/ontology` is debatable.

2) Update the `prepare_release` rule to have a separate command that copies the mappings to the top-level directory (similar to what we already do for the patterns). This involves making sure the mappings to be released are _not_ listed in the general `RELEASE_ASSETS` variable, but are instead listed in a separate variable (again, similar to what we do for the patterns).

gouttegd:
Publishing the SSSOM mapping sets (which is not enabled by default, but can be requested with the `release_mappings` option in the SSSOM option group) is broken because the rsync command that should take care of copying the mappings to the top-level directory cannot handle the case where the input files are in a directory located above the current directory (because of rsync's `-R` option).

We fix that by

(1) not including the mapping files in the `RELEASE_ASSETS` variables, so that they are excluded from the rsync command;

(2) adding a new step in the `prepare_release` rule to explicitly copy the mappings from `src/mappings` to the top-level `mappings` directory (similarly to what we already do for the pattern files);

(3) amend the definition of the `RELEASE_ASSETS_AFTER_RELEASE` variable so that it includes the mapping files in their final, top-level location.

closes #1036

### DIFF
--- a/template/src/ontology/Makefile.jinja2
+++ b/template/src/ontology/Makefile.jinja2
@@ -340,8 +340,7 @@ ASSETS = \
 RELEASE_ASSETS = \
   $(MAIN_FILES) {% if project.import_group is defined %}{% if project.import_group.release_imports %}$(IMPORT_FILES) {% endif %}{% endif %}\
   $(SUBSET_FILES){% if project.robot_report.release_reports %} \
-  $(REPORT_FILES){% endif %}{% if project.sssom_mappingset_group is defined %}{% if project.sssom_mappingset_group.release_mappings %} \
-  $(MAPPING_FILES){% endif %}{% endif %}
+  $(REPORT_FILES){% endif %}
 
 .PHONY: all_assets
 all_assets: $(ASSETS) {% if project.ensure_valid_rdfxml %}check_rdfxml_assets{% endif %}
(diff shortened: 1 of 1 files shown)


template/src/ontology/Makefile.jinja2
matentzn, line 1: > .. the rsync command that should take care of copying the mappings to the top-level directory cannot handle the case where the input files are in a directory located above the current directory..

Isn't it located in `../mappings` ? That's not "above the current dir" (at least not insofer that `-R` would play a role?). Or cant rsync handle the `..`?

gouttegd: > Isn't it located in `../mappings` ?

Yes, and that’s “above“ the current directory (from a certain point of view). What matters here is that we are going one directory upwards relative to the current directory -- that’s what is incompatible with `rsync -R`. The fact that after that we are going downwards again (so that we end up “besides“ the current directory rather than “above“ it) is irrelevant.

matentzn: Such a pity that that place a role!

matentzn (approved):

PR #1119, opened 2024-10-30
Set proxy settings for Java applications.

### LINKED ISSUE #1063: robot needs java http proxy settings when behind a proxy
When building our [application ontology](https://github.com/airr-knowledge/ak-ontology) behind a proxy, robot gets an error when processing DOID.

```
curl -L http://purl.obolibrary.org/obo/DOID.owl --create-dirs -o tmp/DOID-download.owl --retry 4 --max-time 200 && \
robot --catalog catalog-v001.xml convert -i tmp/DOID-download.owl -o tmp/mirror-DOID.owl
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   366    0   366    0     0   3763      0 --:--:-- --:--:-- --:--:--  3734
100 23.8M  100 23.8M    0     0  38.9M      0 --:--:-- --:--:-- --:--:-- 38.9M
org.semanticweb.owlapi.model.UnloadableImportException: Could not load imported ontology: <http://purl.obolibrary.org/obo/doid/obo/ext.owl> Cause: Connection reset
Use the -vvv option to show the stack trace.
Use the --help option to see usage information.
make: *** [Makefile:427: mirror-DOID] Error 1
```

The issue is that Java needs to know about the proxy. As I hack, I manually modified `run.sh`

```
ROBOT_JAVA_ARGS="$ODK_JAVA_OPTS -Dhttp.proxyHost=proxy.swmed.edu -Dhttp.proxyPort=3128" 
```

I suppose the better solution is to check if `HTTP_PROXY` is set in the environment and use it...

gouttegd:
If we detect in the environment some variables indicating that a HTTP(S) proxy should be used, we translate them into the appropriate Java properties so that Java applications (mostly ROBOT) inside the container know about the proxy.

Specifically, we detect a variable like

```
http_proxy=https://example.net:1234
```

(or `HTTP_PROXY`, as this variable is widely used in both uppercase and lowercase forms) and turn it into

```
-Dhttp.proxyHost=example.net
-Dhttp.proxyPort=1234
```

and similarly for `https_proxy` (or `HTTPS_PROXY`), which is turned into `https.proxyHost` and `https.proxyPort` properties.

We also detect a variable like

```
no_proxy=example.com,example.org
```

(or, again, `NO_PROXY`) and turn it into

```
-Dhttp.noProxyHosts=example.com|example.org
```

as [expected by the Java virtual machine](https://docs.oracle.com/javase/8/docs/technotes/guides/net/proxies.html).

closes #1063

### DIFF
--- a/template/src/ontology/run.sh.jinja2
+++ b/template/src/ontology/run.sh.jinja2
@@ -86,6 +86,27 @@ for owlapi_var in $(env | sed -n s/^OWLAPI_//p) ; do
     ODK_JAVA_OPTS="$ODK_JAVA_OPTS -D$OWLAPI_OPTIONS_NAMESPACE.${owlapi_var%=*}=${owlapi_var#*=}"
 done
 
+# Proxy settings for Java applications
+[ -z "$HTTP_PROXY" ] && [ -n "$http_proxy" ] && HTTP_PROXY=$http_proxy
+if [ -n "$HTTP_PROXY" ]; then
+    proxy_host=$(echo $HTTP_PROXY | sed 's,^https\?://,,; s,:[0-9]\+$,,')
+    proxy_port=$(echo $HTTP_PROXY | sed 's,^.*:\([0-9]\+\)$,\1,')
+    ODK_JAVA_OPTS="$ODK_JAVA_OPTS -Dhttp.proxyHost=$proxy_host"
+    [ "$proxy_port" != "$proxy_host" ] && ODK_JAVA_OPTS="$ODK_JAVA_OPTS -Dhttp.proxyPort=$proxy_port"
+fi
+[ -z "$HTTPS_PROXY" ] && [ -n "$https_proxy" ] && HTTPS_PROXY=$https_proxy
+if [ -n "$HTTPS_PROXY" ]; then
+    proxy_host=$(echo $HTTPS_PROXY | sed 's,^https\?://,,; s,:[0-9]\+$,,')
+    proxy_port=$(echo $HTTPS_PROXY | sed 's,^.*:\([0-9]\+\)$,\1,')
+    ODK_JAVA_OPTS="$ODK_JAVA_OPTS -Dhttps.proxyHost=$proxy_host"
+    [ "$proxy_port" != "$proxy_host" ] && ODK_JAVA_OPTS="$ODK_JAVA_OPTS -Dhttps.proxyPort=$proxy_port"
+fi
+[ -z "$NO_PROXY" ] && [ -n "$no_proxy" ] && NO_PROXY=$no_proxy
+if [ -n "$NO_PROXY" ] ; then
+    no_proxy_hosts=$(echo $NO_PROXY | tr ',' '|')
+    ODK_JAVA_OPTS="$ODK_JAVA_OPTS -Dhttp.nonProxyHosts=$no_proxy_hosts"
+fi
+
 TIMECMD=
 if [ x$ODK_DEBUG = xyes ]; then
     # If you wish to change the format string, take care of using


template/src/ontology/run.sh.jinja2
```diff
@@ -86,6 +86,27 @@ for owlapi_var in $(env | sed -n s/^OWLAPI_//p) ; do
     ODK_JAVA_OPTS="$ODK_JAVA_OPTS -D$OWLAPI_OPTIONS_NAMESPACE.${owlapi_var%=*}=${owlapi_var#*=}"
 done
 
+# Proxy settings for Java applications
+[ -z "$HTTP_PROXY" ] && [ -n "$http_proxy" ] && HTTP_PROXY=$http_proxy
+if [ -n "$HTTP_PROXY" ]; then
+    proxy_host=$(echo $HTTP_PROXY | sed 's,^https\?://,,; s,:[0-9]\+$,,')
```
matentzn, line 92: 1. Is it certain that `$HTTP_PROXY` never contains spaces? Else I think you could consider to quote?
2. When I see an sed command like this I am concerned about macOS vs Linux compatibility; I do not know how to test this, but I remember at least that on mac I frequently had to put `sed '' '.....'` to get things to work; also, I remember struggling with `\?` and `\+` which do not work on all systems.. I think this had to do with lack of "extended regex" (`-E`?). My memory is very hazy here, I may be totally wrong.. Probably worth finding a volunteer like the issuer to try on a Mac?

matentzn (commented):
Nothing I am particularly sure of, feel free to dismiss my comments!

template/src/ontology/run.sh.jinja2
gouttegd: > Is it certain that $HTTP_PROXY never contains spaces?

It is never _supposed_ to contain spaces (it’s basically akin to a URL).

We _could_ quote it, but the `run.sh` script already contains several instances of variables that are not supposed to contain spaces, and for which we don’t bother protecting against the possibility that they do.

For example, we already assume that `ODK_IMAGE` is a single word. Try calling the script with

```sh
ODK_IMAGE="my image name" sh run.sh
```

and watch as Docker is trying to find an image called `obolibrary/my:latest`... :D

Overall, there’s only so much we can do to protect users from their own mistakes. If they define a space-containing HTTP_PROXY in their environment, they are asking for troubles, and not only with the ODK.

gouttegd: Re `sed` portability:

The problem is simply that GNU `sed` and BSD `sed` (macOS uses the BSD one) have different ideas of what “Basic Regular Expressions” (BRE). In particular, GNU `sed` considers `\+` and `\?` in BRE to be equivalent to `+` and `?` in “Extended Regular Expressions” (ERE), which BSD `sed` does not -- for BSD `sed`, in BRE mode `\+` has no special meaning at all, it’s just a regex attempting to find a literal `\` followed by a literal `+`.

But both the GNU and the BSD `sed` do have the same idea of what “Extended Regular Expressions” are (they’re POSIX-defined), so the only thing to remember really is to always use the `-E` option whenever portability is required.

(Which is, yes, exactly what I forgot to do here. ^^' )

gouttegd: > the only thing to remember really is to always use the -E option whenever portability is required.

To further clarify: the `-E` option may not be needed if the regex is a “simple” one. For example, elsewhere in `run.sh` we have this:

```sh
sed -n s/^OWLAPI_//p
```

This does not require the `-E` option as the regular expression does not contain any special character whose interpretation could vary across `sed` implementations (the only “special” character is `^`, which is universally recognised as the “start-of-match” anchor in virtually every regex engines ever written, long before POSIX was even a thing).

But overall, when in doubt about a `sed` command: use `-E`.

matentzn (approved):
I give this a LGTM review, because this is a bit beyond my expertise limits. Looks sane though, and very useful!

PR #1121, opened 2024-11-01
Allow generating the documentation for DOSDP patterns

gouttegd:
This PR adds new targets to the standard Makefile (`dosdp-docs-X`, where _X_ is the name of a DOSDP pipeline) to automatically generate the documentations for DOSDP patterns.

This is inspired by CL’s custom [`pattern_docs` target](https://github.com/obophenotype/cell-ontology/blob/1109cf0f40ad53666eb5b4ebda2635ac6bba45db/src/ontology/cl.Makefile#L204), except there here are using DOSDPTools and its (seemingly undocumented) `docs` subcommand rather than DOSDP, because DOSDP seems to crash on way too many patterns.

### DIFF
--- a/template/src/ontology/Makefile.jinja2
+++ b/template/src/ontology/Makefile.jinja2
@@ -39,6 +39,7 @@ ROBOT=                      robot --catalog $(CATALOG)
 {% if project.owltools_memory|length %}OWLTOOLS_MEMORY=            {{ project.owltools_memory }}{% endif %}
 OWLTOOLS=                   {% if project.owltools_memory|length %}OWLTOOLS_MEMORY=$(OWLTOOLS_MEMORY) {% endif %}owltools --use-catalog
 RELEASEDIR=                 ../..
+DOCSDIR=                    ../../docs
 REPORTDIR=                  reports
 TEMPLATEDIR=                ../templates
 TMPDIR=                     tmp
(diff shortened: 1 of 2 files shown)


template/src/ontology/Makefile.jinja2
```diff
@@ -829,28 +830,52 @@ $(PATTERNDIR)/dospd-patterns/%.yml: download_patterns
...
+.PHONY: dosdp-docs-default
+dosdp-docs-default: $(EDIT_PREPROCESSED) $(DOSDP_TSV_FILES_DEFAULT) $(DOSDP_YAML_FILES_DEFAULT)
+	mkdir -p $(DOCSDIR)/patterns/default
+	$(DOSDPT) docs {{ project.dosdp_tools_options }} --catalog=$(CATALOG) \
+		       --ontology=$< \
+		       --infile=$(PATTERNDIR)/data/default \
+		       --template=$(PATTERNDIR)/dosdp-patterns \
+		       --batch-patterns="$(DOSDP_PATTERN_NAMES_DEFAULT)" \
+		       --outfile=$(DOCSDIR)/patterns/default{% if project.github_org and project.repo %} \
+		       --data-location-prefix=https://github.com/{{ project.github_org }}/{{ project.repo }}/tree/{{ project.git_main_branch }}/src/patterns/data/default{% endif %}
```
matentzn, line 852: Just noting this is web location not "raw" which is probably correct for docs, and it is a bit inconvenient for our gitlab users like @ehartley to have GitHub assumptions hard coded into this - maybe just factor this into a make variable to make it  at least overwritable at runtime?

template/src/ontology/Makefile.jinja2
```diff
@@ -829,28 +830,52 @@ $(PATTERNDIR)/dospd-patterns/%.yml: download_patterns
...
+.PHONY: dosdp-docs-{{ pipeline.id }}
+dosdp-docs-{{ pipeline.id }}: $(EDIT_PREPROCESSED) $(DOSDP_TSV_FILES_{{ pipeline.id.upper() }}) $(DOSDP_YAML_FILES_{{ pipeline.id.upper() }})
+	mkdir -p $(DOCSDIR)/patterns/{{ pipeline.id }}
+	$(DOSDPT) docs {{ pipeline.dosdp_tools_options }} --catalog=$(CATALOG) \
+		       --ontology=$< \
+		       --infile=$(PATTERNDIR)/data/{{ pipeline.id }} \
+		       --template=$(PATTERNDIR)/dosdp-patterns \
+		       --batch-patterns="$(DOSDP_PATTERN_NAMES_{{ pipeline.id.upper() }})" \
+		       --outfile=$(DOCSDIR)/patterns/{{ pipeline.id }}{% if project.github_org and project.repo %} \
+		       --data-location-prefix=https://github.com/{{ project.github_org }}/{{ project.repo }}/tree/{{ project.git_main_branch }}/src/patterns/data/{{ pipeline.id }}{% endif %}
```
matentzn, line 878: Same as above (hard baked GitHub assumption)

template/src/ontology/Makefile.jinja2
gouttegd: > Just noting this is web location not "raw" which is probably correct for docs

Yes. For context, what this `--data-location-prefix` option does is providing the link that `dosdp-tools docs` will insert in the documentation, along with a message saying something like “to see the full TSV table for this pattern, please go there”.

> maybe just factor this into a make variable to make it at least overwritable at runtime?

Note that this is already conditional to the ODK configuration defining the `github_org` variable. For a project hosted elsewhere than on GitHub, the `--data-location-prefix` option will be skipped entirely.

Since we already have `github_org` and `repo` in the ODK configuration file, I think that if we want to enable this feature for non-GitHub-hosted repos, it should also happen in the ODK configuration rather than in the Makefile.

And since we cannot obviously handle all possible repository hosting services, I would suggest adding a single `repo_url` (or similar) configuration option, to be used whenever `github_org` is not set.

So, a GitHub-hosted project would define:

```
repo: my-repo-name
github_org: my-organisation
```

which would yield the `--data-location-prefix` URL shown above, while a non-GitHub-hosted project would define:

```
repo: my-repo-name
repo_url: https://my.custom.hosting.provider.example.net/root/of/my/repository
```

which would yield `--data-location-prefix=https://my.custom.hosting.provider.example.net/root/of/my/repository/src/patterns/data/default/`.

matentzn: I checked the rest of the ODK setup and `github_org` is connected to soo many assumptions on GitHub that I have to agree with you (I was going to say that gitlab has indeed quite a similar structure and this specific feature _could_ work for gitlab at least, but as you say, it would not support any other hosting infrastructure).

Cool lets proceed then. The github_org refactor is massive so I don't think its worth trying to do this here as part of this specific PR unless you want to just start for this feature alone!

matentzn (dismissed):

template/src/ontology/Makefile.jinja2
gouttegd: > I don't think its worth trying to do this here as part of this specific PR unless you want to just start for this feature alone!

Well, we have to start somewhere :) , so yes, if you agree with the idea of a generic `repo_url` config key I’ll amend this PR to add it.

matentzn: Totally agreed!

matentzn (approved):
Nicccce!!

PR #1124, opened 2024-11-04
Avoid adding setuptools to constraints.txt

matentzn:
This is a bit of a hack, but I cant imagine it being a problem, as during ODK install https://github.com/INCATools/ontology-development-kit/commit/f1bdb983e2ae0128e7a8fbfb60b7c05755c4c5d0 kicks in.

However, it might create some non-deterministic behaviour in the process.. not sure.

@gouttegd what do you think?

You were indeed right that #1123 broke the build!

### DIFF
--- a/constraints.txt
+++ b/constraints.txt
@@ -11,11 +11,12 @@ asttokens==2.4.1
 async-lru==2.0.4
 attrs==24.2.0
 babel==2.16.0
-babelon==0.3.1
+babelon==0.2.9
+backports.tarfile==1.2.0
 bcp47==0.1.0
 beautifulsoup4==4.12.3
 bidict==0.23.1
-bleach==6.2.0
+bleach==6.1.0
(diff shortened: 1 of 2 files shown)


gouttegd:
> This is a bit of a hack, but I cant imagine it being a problem

I can. Even if you forcefully remove `setuptools` from the generated constraints list, the constraints in the list will still have been computed on the assumption that a more recent version of `setuptools` is available, which can and most likely will cause further issues -- if only because a more recent version of `setuptools` can bring a newer version of `wheels`, which would conflict with the version installed by Debian.

We need a better solution where we force PIP not to upgrade any package that is already installed by Debian (not only `setuptools`). On it.

matentzn:
Ooooook!! :) feel free to just revert the commit by merging #1125 then just to unblock master branch

gouttegd:
Slightly better solution in #1126.

PR #1127, opened 2024-11-04
Use a dedicated token to create the "update constraints" PR.

gouttegd:
When updating the Python constraints, create the PR using a dedicated token instead of the default GITHUB_TOKEN. Using the default token has the side-effect that the PR will not trigger any workflow, which in particular means that the test suite (`build-and-test.yml` workflow) will not be run.

A new token has been created on @gouttegd's account and added to the ODK repository. This commit amend the "update-constraints" workflow to specifically use that token.

### DIFF
--- a/.github/workflows/constraints.yml
+++ b/.github/workflows/constraints.yml
@@ -20,6 +20,7 @@ jobs:
         uses: peter-evans/create-pull-request@v3
         if: ${{ success() }}
         with:
+          token: ${{ secrets.WORKFLOW_TOKEN }}
           commit-message: Update constraints.txt
           title: 'Update constraints.txt'
           body: |


matentzn:
I was just going to suggest to simply add QC as a separate step to this action: 

image

But I guess this is more DRY :P Thanks

gouttegd:
> I was just going to suggest to simply add QC as a separate step to this action

That would work as well.

My only concern is that, with this approach, if the build fails because the new constraints are causing a problem, we won’t get to the stage where we create a PR, which means that we won’t know what the new constraints were. We will know that they caused a problem, but we will need to dig through the log of the workflow in the hope of finding a meaningful PIP error message (assuming PIP is capable of producing meaningful error messages, something I never saw it doing…).

By letting the action create a PR no matter what, and _then_ running the normal test suite, we can immediately see the new constraints; if the tests fail, we will already know which packages have been upgraded, so it will be easier to find the faulty package(s).

matentzn:
> By letting the action create a PR no matter what, and then running the normal test suite, we can immediately see the new constraints; if the tests fail, we will already know which packages have been upgraded, so it will be easier to find the faulty package(s).

This can be achieved by doing the PR first, then running QC! Given that, which solution is preferable? You choose :P

gouttegd:
> This can be achieved by doing the PR first, then running QC!

Yes, this is exactly the goal of the present PR: making sure the QC workflow is run once the `update-constraints` workflow is over.

matentzn (approved):
Lets try this!

PR #1128, opened 2024-11-04
Update constraints.txt

gouttegd:
This pull request updates constraints.txt. Make sure you run make tests manually.

### DIFF
--- a/constraints.txt
+++ b/constraints.txt
@@ -177,7 +177,7 @@ ontobio==2.9.10
 ontodev-cogs==0.3.3
 ontodev-gizmos==0.3.2
 ontoportal-client==0.0.4
-openai==1.53.0
+openai==1.53.1
 openpyxl==3.1.5
 ordered-set==4.1.0
 overrides==7.7.0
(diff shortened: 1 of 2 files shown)


gouttegd:
@matentzn Not sure why the action was triggered now (I didn’t change the cron setting, the workflow is still supposed to be run at 1pm every Monday)… But at least we can see that it seems to be working.

The fact that the PR seems to be coming from me is the expected consequence of using a dedicated token instead of `GITHUB_TOKEN`: I created the dedicated token, so everything it does appear to come from me.

What puzzles me is the fact that the _commit_ seems to be coming from you. I don’t know why that is the case.

matentzn:
i triggered it!

gouttegd:
Ah, OK! This also explains why you are the committer, instead of the `github-actions` bot. All fine then.

gouttegd:
Was it really necessary to merge that PR just to get the latest version of `openai`? I thought you just wanted to check that everything was working.

Now I have to update #1129 because `pip-constraints` has changed while the test suite was running.

matentzn:
Sorry man, no it was not necessary; I thought of your PR too late. My apologies, keeping my hands off it now.

PR #1129, opened 2024-11-04
Improve the “update constraints” mechanism

gouttegd:
This PR builds upon #1126 to make the updating of Python constraints more robust.

The key change is that, when computing the Python constraints, we install all the required Python packages in a **minimal** image that is just `ubuntu:24.04` (same image as the base used to build the ODK) in which we install nothing else than `python3-dev` and `python3-pip`.

Previously, we were installing the Python packages in an image that was the ODK itself (`obolibrary/odkfull:latest`), which was problematic for at least two reasons:

* The final ODK image already contains (obviously) all the Python packages we want. In theory this should not cause any problem because when computing the Python constraints we are working in a virtual environment, but for some packages (at least `setuptools` and `wheels`, and possibly others) this is not enough. Using a minimal image ensures that we are always computing the Python constraints from a clean state.
* This cannot work when we need to reset the Python constraints after updating the base image itself (e.g., when we will upgrade to Ubuntu 26.04), because of a chicken-and-egg problem: we need the new `obolibrary/odkfull:latest` (based on the updated Ubuntu image) to compute the new Python constraints, but we need the new Python constraints to build the new `obolibrary/odkfull:latest`. Previously, we were working around that issue by having a separate, undocumented process specifically to update the Python constraints after a base image upgrade. Now we will be able to use exactly the same process (same `update-constraints.sh` script) both for the regular updates of the Python constraints, and for the “big updates” after a migration to a new base image.

In addition, when building the ODK itself (more precisely, the `odkbuilder` image), we check that the list of system-managed Python packages matches what we are expecting (what was determined by the `update-constraints.sh` script, and stored in `pip-constraints.txt`), and we log a warning if there is a discrepancy (which may happen if a Ubuntu-provided Python package has been updated, e.g. for security reasons, since the last refresh of the constraints -- something that should be unlikely to happen often, but just in case).

### DIFF
--- a/.github/workflows/constraints.yml
+++ b/.github/workflows/constraints.yml
@@ -6,7 +6,8 @@ on:
 jobs:
   constraints:
     runs-on: ubuntu-latest
-    container: obolibrary/odkfull:latest
+    # This must be the same image as the one the ODK is built on.
+    container: ubuntu:24.04
     strategy:
       max-parallel: 1
     steps:
@@ -15,7 +16,7 @@ jobs:
       - name: work around permission issue
         run: git config --global --add safe.directory "$GITHUB_WORKSPACE"
       - name: Build constraints.txt
-        run: sh update-constraints.sh
+        run: sh update-constraints.sh --in-docker
       - name: Create Pull Request
         uses: peter-evans/create-pull-request@v3
         if: ${{ success() }}
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -380,12 +380,15 @@ If the component needs to be built from source, do so in [the Dockerfile for odk
 and install the compiled file(s) in either the `/staging/full` tree or the `/staging/lite` tree, for
 inclusion in `odkfull` or `odklite` respectively.
 
-If the component is a Python package, adds it to the `requirements.txt`
+If the component is a Python package, add it to the `requirements.txt.full`
 file, and *also* in the `requirements.txt.lite` file if it is to be part
 of `odklite`. Please try to avoid version constraints unless you can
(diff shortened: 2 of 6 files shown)


gouttegd:
And another run of the test suite, yeah!

matentzn (approved):
I know you have not asked yet, but I have spent 20 min or so reviewing this (I admit, update-constraints.sh with the help of chatgpt to explain some things), and it all looks good to me!

gouttegd:
> I know you have not asked yet

As a general rule, I wait for the test suite to have passed before asking for a review. :)

(Officially to avoid wasting people’s time, by not asking them to review something that may need to be modified because the tests failed; in reality, to preserve the chance that, if I make a stupid mistake – the kind of mistake that the test suite will catch immediately –, I can fix it before anyone else can see it. :D )

gouttegd:
> update-constraints.sh with the help of chatgpt to explain some things

I am mildly curious of what in the `update-constraints.sh` script requires ChatGPT to be understood. O_o

matentzn:
> I am mildly curious of what in the update-constraints.sh script requires ChatGPT to be understood. O_o

1. I have never heard of PIP_CONSTRAINT as an environment variable
2. I was too lazy to read https://github.com/INCATools/ontology-development-kit/pull/1129/files#diff-151c918f71ba2bcb208cea5edafa9abecc62d8b466ad400717ede5a0ee7e112dR18 until I realised you had already explained it to me.

gouttegd:
> I have never heard of PIP_CONSTRAINT as an environment variable

To be fair I think that Pip’s documentation is quite lacking here.

PR #1131, opened 2024-11-12
cicd: update action refs

jsstevenson:
This PR updates version references for a few CI actions, both in the repo's workflows and in the ODK template itself. It's probably _fine_ the way it is but you'll start to see warnings from GH on a few of them (e.g. I believe download-artifact v2 is deprecated and will include a complaint every time you run it).

I don't personally use ODK, so apologies if this is off-base, but I randomly saw a diff on the OBO call and went digging.


gouttegd (approved):
It’s something we should have done quite a while ago already. Thanks :)

PR #1135, opened 2024-11-22
Install the SPARQLProg package system-wide.

### LINKED ISSUE #1134: SPARQLProg tools are unusable in ODK 1.5 under a non-root user
Chris Mungall’s [SPARQLProg tools](https://github.com/cmungall/sparqlprog) are unusable in ODK 1.5 when running under a non-privileged user (which is the default setting).

This is because SWIPL installs the SPARQLProg package, by default, under the the running user’s home directory (`~/.local/share/swi-prolog/pack/sparqlprog`), and of course during the ODK building process the “running user” is _root_. Once the image is built, the non-privileged user has no access to anything under `/root`, so no access to `/root/.local/share/swi-prolog/pack/sparqlprog`.

gouttegd:
When installing the SPARQLProg package using SWIPL, we must install make sure that the package is installed in a system-wide directory (under /usr/local) rather than the home directory of the running user (root). Otherwise only root would be able to use the package.

This PR does that by setting the `global` option to `true` when calling SWIPL’s `pack_install` function. Then we update the running PATH to include the `/usr/local/share/swi-prolog/pack/sparqlprog/bin` directory. And we add SPARQLProg’s `pl2sparql` program to the list of programs to check after a build.

closes #1134

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@ ARG ODKLITE_TAG=latest
 FROM obolibrary/odklite:${ODKLITE_TAG}
 LABEL maintainer="obo-tools@googlegroups.com"
 
-ENV PATH="/tools/apache-jena/bin:/tools/sparqlprog/bin:$PATH"
+ENV PATH "/tools/apache-jena/bin:/usr/local/share/swi-prolog/pack/sparqlprog/bin:$PATH"
 
 ARG ODK_VERSION 0.0.0
 ENV ODK_VERSION=$ODK_VERSION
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -4,7 +4,7 @@ ARG ODKLITE_TAG=latest
 FROM obolibrary/odklite:${ODKLITE_TAG}
 LABEL maintainer="obo-tools@googlegroups.com"
 
-ENV PATH="/tools/apache-jena/bin:/tools/sparqlprog/bin:$PATH"
+ENV PATH "/tools/apache-jena/bin:/usr/local/share/swi-prolog/pack/sparqlprog/bin:$PATH"
```
matentzn, line 7: https://docs.docker.com/reference/build-checks/legacy-key-value-format/

😂

gouttegd: Doh!

So what probably happened is that I started fixing this issue while I was on the 1.5 maintenance branch (where your fix for the `ENV` declaration is not present), and then when I realized that I switched over to the master branch -- and in the process, I ported back to the master branch the old `ENV` declaration.

My bad.

PR #1138, opened 2024-11-26
Remove sudo from constraints.yml

matentzn:
Since sudo is not installed in your ubuntu:24.04 container, removing it avoids the [error](https://github.com/INCATools/ontology-development-kit/actions/runs/12027519909/job/33528700033).

In Docker containers, commands are typically executed as root. Therefore, we don't seem to need sudo.

### DIFF
--- a/.github/workflows/constraints.yml
+++ b/.github/workflows/constraints.yml
@@ -13,8 +13,8 @@ jobs:
     steps:
       - name: Install Git
         run: |
-          sudo apt-get update
-          sudo apt-get install -y --no-install-recommends git
+          apt-get update
+          apt-get install -y --no-install-recommends git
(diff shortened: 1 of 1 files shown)


gouttegd (approved):
I had been wondering yesterday whether going through `sudo` was needed, but I thought that, even if _not_ needed, it couldn’t do any harm.

I am quite surprised that `sudo` is _not_ present in Ubuntu, though -- I thought it was always part of the base install, since Ubuntu was (as I recall) one of the first GNU/Linux distros to disallow logging in as root and forcing the use of `sudo` for all administrative tasks.

I guess whoever created the Ubuntu Docker image decided it was not needed since Docker’s default behaviour is to run everything inside the container as root anyway.

PR #1141, opened 2024-11-27
Adding ca-certificates install to constraints.yml

matentzn:
Just fumbling in the dark now.. But the error suggests that is needed, and perhaps --no-install-recommends precludes this being installed?

https://github.com/INCATools/ontology-development-kit/actions/runs/12045881518

### DIFF
--- a/.github/workflows/constraints.yml
+++ b/.github/workflows/constraints.yml
@@ -11,10 +11,10 @@ jobs:
     strategy:
       max-parallel: 1
     steps:
-      - name: Install Git
+      - name: Install Required Packages
         run: |
           apt-get update
-          apt-get install -y --no-install-recommends git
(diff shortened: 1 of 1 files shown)


gouttegd:
The year is 2035. While the world is slowly dealing with the aftermath of World War III, Nico & Damien are trying their 764th patch to make the ODK’s `update-constraints` workflow to work…

Joking aside, I agree about `ca-certificates` being almost certainly needed here (and again, I thought this would have come with the base image).

gouttegd:
Note that if this still doesn’t work, I’ll spin up a `ubuntu:24.04` image locally, get a shell into it, and try to find out once and for all what exactly is needed for `git` to be able to `pull` and `push` from such an image.

gouttegd (approved):

PR #1143, opened 2024-11-28
Earth in 2150: fixing constraints.yml part 5

matentzn:
At my wits end, I now consulted with various AI tools and they say:

```
persist-credentials: false:

Ensures the credentials are correctly stored for subsequent Git operations, preventing authentication issues.
```

This is super annoying.. Now your suggestion (of testing locally) wont even work because the problem is _specifically_ about the last github action step, so you cant really test it anywhere other than in the repo (or in a fork, which is a lot of churn).

I wonder, alternatively, if it would work to run it all in Ubuntu base image, and then

```
run: |
          docker run --rm -v ${{ github.workspace }}:/workspace -w /workspace ubuntu:24.04 \
          sh update-constraints.sh --in-docker
```

just in the Build constraints.txt step?

### DIFF
--- a/.github/workflows/constraints.yml
+++ b/.github/workflows/constraints.yml
@@ -6,26 +6,23 @@ on:
 jobs:
   constraints:
     runs-on: ubuntu-latest
-    # This must be the same image as the one the ODK is built on.
-    container: ubuntu:24.04
     strategy:
       max-parallel: 1
     steps:
-      - name: Install Required Packages
-        run: |
-          apt-get update
-          apt-get install -y --no-install-recommends git ca-certificates
+      - name: Check Docker version
+        # This is to fail early in case docker is not available on the base image..
+        run: docker --version
       - name: Checkout main branch
         uses: actions/checkout@v4
-        with:
-          persist-credentials: false  # Ensure credentials are stored properly
       - name: Build constraints.txt
-        run: sh update-constraints.sh --in-docker
+        run: |
+          docker run --rm -v ${{ github.workspace }}:/workspace -w /workspace ubuntu:24.04 \
+          sh update-constraints.sh --in-docker
       - name: Create Pull Request
         uses: peter-evans/create-pull-request@v3
         if: ${{ success() }}
         with:
-          token: ${{ secrets.GITHUB_TOKEN }}
+          token: ${{ secrets.WORKFLOW_TOKEN }}
           commit-message: Update constraints.txt
           title: 'Update constraints.txt'
           body: |
--- a/.github/workflows/constraints.yml
+++ b/.github/workflows/constraints.yml
@@ -17,13 +17,15 @@ jobs:
           apt-get install -y --no-install-recommends git ca-certificates
       - name: Checkout main branch
         uses: actions/checkout@v4
+        with:
+          persist-credentials: false  # Ensure credentials are stored properly
       - name: Build constraints.txt
         run: sh update-constraints.sh --in-docker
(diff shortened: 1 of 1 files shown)


gouttegd:
> I now consulted with various AI tools and they

did the typical AI crap of not having the slightest understanding of the entire issue, seeing a few keywords and jumping at a conclusion. “PhD-level intelligence“ my arse!

Looking at the [last failed workflow](https://github.com/INCATools/ontology-development-kit/actions/runs/12065006941/job/33642914714), at the “create pull request” step:

```
Save persisted git credentials
  /usr/bin/git config --local --name-only --get-regexp http.https://github.com/.extraheader ^AUTHORIZATION:
  fatal: --local can only be used inside a git repository
Determining the base and head repositories
  /usr/bin/git config --local --get remote.origin.url
  fatal: --local can only be used inside a git repository
  Error: The format of '' is not a valid GitHub repository URL
Restore persisted git credentials
  /usr/bin/git config --local --name-only --get-regexp http.https://github.com/.extraheader ^AUTHORIZATION:
  fatal: --local can only be used inside a git repository
```

Look how _all_ Git commands fail with a `fatal: --local can only be used inside a Git repository` -- _including_ the second command (“determining the base and head repositories“), which has nothing to do with the “saving/restoring of persisted git credentials“. Disabling the persistence of Git credentials would remove the first and third commands, but those commands are clearly not the root cause of the problem -- despite what an over-hyped, power-hungry, time-wasting stochastic parrot may say.

Compare with a [previous non-failed workflow](https://github.com/INCATools/ontology-development-kit/actions/runs/11671387329/job/32497650341):

```
Save persisted git credentials
  /usr/bin/git config --local --name-only --get-regexp http.https://github.com/.extraheader ^AUTHORIZATION:
  http.https://github.com/.extraheader
  /usr/bin/git config --local --get-regexp http.https://github.com/.extraheader ^AUTHORIZATION:
  http.https://github.com/.extraheader AUTHORIZATION: basic ***
  /usr/bin/git config --local --unset http.https://github.com/.extraheader ^AUTHORIZATION:
  Unset config key 'http.https://github.com/.extraheader'
Determining the base and head repositories
  /usr/bin/git config --local --get remote.origin.url
  https://github.com/INCATools/ontology-development-kit
Pull request branch target repository set to INCATools/ontology-development-kit
Configuring credential for HTTPS authentication
  /usr/bin/git config --local http.https://github.com/.extraheader AUTHORIZATION: basic ***
```

See how the “determining the base and head repositories“ step did _not_ fail here?

So, the real problem here, which has nothing to do with credentials that are persisted or not, in that in our new workflow, for some reason `git` does not seem to think it is running within a repository. Now _why_ does that happen, I have no clue, but playing with credentials persistence is not going to change anything.

> ```
> run: |
>          docker run --rm -v ${{ github.workspace }}:/workspace -w /workspace ubuntu:24.04 \
>          sh update-constraints.sh --in-docker
> ```

This looks like a _much better_ solution. One that actually makes perfect sense, contrary to that AI crap.

matentzn:
Next shot in the oven..

gouttegd (approved):
I am no longer sure of anything with this workflow 😅 , but this _does_ look good to me.

matentzn:
It worked!!!! 🎉

gouttegd:
Can’t help but noticing that in the end, it was the idea tentatively advanced by a human with a modest “I wonder if…” that was correct over the pseudo-solution confidently advanced by a LLM. Isn’t that interesting?

PR #1145, opened 2024-11-28

### LINKED ISSUE #1142: Add SQLite/SemSQL as an export format

gouttegd:
This PR adds `db` as a new export format for release artefacts.

A `.db` file is a SQLite3 file containing a SemSQL representation of the release product.

I would have preferred to use a different name (like `semsql`) for the export name, but currently, the code in the ODK assumes that the _name_ of a format is necessarily the same thing as its _extension_, so it is not possible to declare a format named `semsql` that is supposed to produce files with an extension of `.db`. This is something that we may want to change in the future, but arguably being able to produce SQLite/SemSQL files is more important than being able to name the format the way we would like.

All tools needed to build SQLite3/SemSQL files (`semsql`, `rdftab`, `sqlite3`, and `relation-graph`) are moved from ODKFull to ODKLite, so that the standard, ODK-generated workflows still only require ODKLite. This increases the size of ODKLite from ~1.34GB to ~1.5GB (compared to ~3.06GB for ODKFull).

closes #1142

### DIFF
--- a/template/src/ontology/Makefile.jinja2
+++ b/template/src/ontology/Makefile.jinja2
@@ -757,6 +757,10 @@ $(SUBSETDIR)/%.json: $(SUBSETDIR)/%.owl
 	$(ROBOT) convert --input $< --check false -f json -o $@.tmp.json &&\
 	mv $@.tmp.json $@
 {% endif -%}
+{% if 'db' in project.export_formats -%}
+$(SUBSETDIR)/%.db: $(SUBSETDIR)/%.owl
+	semsql make $(SUBSETDIR)/$*.db && rm -f $(SUBSETDIR)/$*-relation-graph.tsv.gz
+{% endif -%}
 
 {% endif %}
 
@@ -1056,6 +1060,10 @@ $(TRANSLATIONSDIR)/%.babelon.json: $(TRANSLATIONSDIR)/%.babelon.tsv
 		convert --check false -f json -o $@.tmp.json &&\
 		mv $@.tmp.json $@
 {% endif -%}
+{% if 'db' in project.export_formats -%}
+{{ release }}.db: {{ release }}.owl
+	semsql make {{ release }}.db && rm -f {{ release }}-relation-graph.tsv.gz
+{% endif -%}
 {% endfor -%}
 
 # ----------------------------------------
@@ -1080,6 +1088,10 @@ $(ONT).json: $(ONT).owl
 	$(ROBOT) annotate --input $< --ontology-iri $(URIBASE)/$@ $(ANNOTATE_ONTOLOGY_VERSION) \
 		convert --check false -f json -o $@.tmp.json &&\
 		mv $@.tmp.json $@
+{% endif -%}
+{% if 'db' in project.export_formats -%}
+$(ONT).db: $(ONT).owl
+	semsql make $(ONT).db && rm -f $(ONT)-relation-graph.tsv.gz
 {% endif -%}
 
 # -----------------------------------------------------
--- a/Dockerfile
+++ b/Dockerfile
@@ -33,7 +33,6 @@ RUN apt-get update && DEBIAN_FRONTEND="noninteractive" apt-get install -y --no-i
     automake \
     aha \
     dos2unix \
-    sqlite3 \
     libjson-perl \
     libbusiness-isbn-perl \
     pkg-config \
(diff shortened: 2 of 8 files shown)


template/src/ontology/Makefile.jinja2
```diff
@@ -1056,6 +1060,10 @@ $(TRANSLATIONSDIR)/%.babelon.json: $(TRANSLATIONSDIR)/%.babelon.tsv
 		convert --check false -f json -o $@.tmp.json &&\
 		mv $@.tmp.json $@
 {% endif -%}
+{% if 'db' in project.export_formats -%}
+{{ release }}.db: {{ release }}.owl
+	semsql make {{ release }}.db && rm -f {{ release }}-relation-graph.tsv.gz
```
matentzn, line 1065: Here is an example of a fully battle hardened semsql pipeline with years of use:

https://github.com/monarch-initiative/mondo-ingest/blob/52541e9e0e4737168c36a17a921963675e821adc/src/ontology/mondo-ingest.Makefile#L31

```
%.db: %.owl
	@rm -f $*.db
	@rm -f .template.db
	@rm -f .template.db.tmp
	@rm -f $*-relation-graph.tsv.gz
	RUST_BACKTRACE=full semsql make $*.db -P config/prefixes.csv
	@rm -f .template.db
	@rm -f .template.db.tmp
	@rm -f $*-relation-graph.tsv.gz
	@test -f $*.db || (echo "Error: File not found!" && exit 1)
```

1. all the `rm` we added were from experience when the db was _partially created_. As the pipeline in its inner workings uses `make`, failures could have quite weird effects on caching.
2. `-P config/prefixes.csv` without this, the semsql was virtually unusable, as there were always missing prefixes. However, getting `config/prefixes.csv` right is not easy.
3. `RUST_BACKTRACE=full` I don't remember but somehow this was usfull for debugging?

Pick and choose, I am fine with what you did, just wanted to let you know what I usually do in case any of it is of any relevance. (For really due diligence you could add all the temporary files to .gitignore as well)

matentzn (commented):
AWESOME!

template/src/ontology/Makefile.jinja2
gouttegd: > all the rm we added were from experience when the db was partially created.

😱 Makes me wonder if `semsql` is robust enough to be enabled by default… Maybe it should only be enabled by projects that can rely on “ontology pipeline engineers” being available to take care of it.

gouttegd: If any case I like the idea of having a single pattern rule to handle _all_ productions of `.db` files, instead of having a similar rule at two or three different places.

gouttegd: > -P config/prefixes.csv without this, the semsql was virtually unusable, as there were always missing prefixes. However, getting `config/prefixes.csv` right is not easy.

Sounds like another reason for maybe leaving the decision to enable SemSQL export to each project. Not sure how the ODK could automatically make sure the prefix file is always “right“.

gouttegd: I mean, what good would it do to have all ontologies producing their own SQLite files, if said files end up being unusable and people conclude that they would be better off continuing to use the files from the “SemSQL collection”?

gouttegd: I’ll leave it by default _for now_, as this has the nice side effect that SemSQL export is tested with _all_ the test ontologies from the test suite, instead of only the one in which I explicitly added the `db` export format.

But I will remove it as a default export format once the PR is finalised.

gouttegd:
Oh, apache.org, why are you doing that to me? 😭 

```
#14 [stage-0  8/13] RUN wget -nv http://archive.apache.org/dist/jena/binaries/apache-jena-4.9.0.tar.gz -O- | tar xzC /tools &&     mv /tools/apache-jena-4.9.0 /tools/apache-jena
#14 133.5 failed: Connection timed out.
#14 133.5 failed: Network is unreachable.
#14 133.5 
#14 133.5 gzip: stdin: unexpected end of file
#14 133.5 tar: Child returned status 1
#14 133.5 tar: Error is not recoverable: exiting now
#14 ERROR: process "/bin/sh -c wget -nv http://archive.apache.org/dist/jena/binaries/apache-jena-$JENA_VERSION.tar.gz -O- | tar xzC /tools &&     mv /tools/apache-jena-$JENA_VERSION /tools/apache-jena" did not complete successfully: exit code: 2
```

gouttegd:
Superseded by #1147

PR #1147, opened 2024-11-28

### LINKED ISSUE #1142: Add SQLite/SemSQL as an export format

gouttegd:
This PR adds db as a new export format for release artefacts.

A .db file is a SQLite3 file containing a SemSQL representation of the release product.

All tools needed to build SQLite3/SemSQL files (semsql, rdftab, sqlite3, and relation-graph) are moved from ODKFull to ODKLite, so that the standard, ODK-generated workflows still only require ODKLite. This increases the size of ODKLite from ~1.34GB to ~1.5GB (compared to ~3.06GB for ODKFull).

If the project is configured to use a global prefix context (`use_context=true`), the same context (converted from JSON to CSV) is used when building the SemSQL files. Otherwise, we let `semsql` use its built-in list of prefixes.

closes #1142

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -33,7 +33,6 @@ RUN apt-get update && DEBIAN_FRONTEND="noninteractive" apt-get install -y --no-i
     automake \
     aha \
     dos2unix \
-    sqlite3 \
     libjson-perl \
     libbusiness-isbn-perl \
     pkg-config \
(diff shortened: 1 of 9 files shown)


gouttegd:
So enabling SemSQL export by default means that running the test suite now takes so long, it cannot be completed before running into the 360-minutes time limit…

gouttegd:
Superseded by #1148, which does not enable the new export format by default.

PR #1148, opened 2024-11-29

### LINKED ISSUE #1142: Add SQLite/SemSQL as an export format

gouttegd:
(This is exactly the same PR as #1147, just without the last commit which enabled SQLite/SemSQL export by default on all repos.)

This PR adds `db` as a new export format for release artefacts.

A `.db` file is a SQLite3 file containing a SemSQL representation of the release product.

All tools needed to build SQLite3/SemSQL files (`semsql`, `rdftab`, `sqlite3`, and `relation-graph`) are moved from ODKFull to ODKLite, so that the standard, ODK-generated workflows still only require ODKLite. This increases the size of ODKLite from ~1.34GB to ~1.5GB (compared to ~3.06GB for ODKFull).

If the project is configured to use a global prefix context (`use_context=true`), the same context (converted from JSON to CSV) is used when building the SemSQL files. Otherwise, we let `semsql` use its built-in list of prefixes.

closes #1142

### DIFF
--- a/scripts/context2csv.py
+++ b/scripts/context2csv.py
@@ -1,0 +1,16 @@
+#!/usr/bin/env python3
+
+import json
+import sys
+
+try:
+    context = json.load(sys.stdin)
+except json.JSONDecodeError:
+    sys.exit("Cannot read context file")
+
+if not '@context' in context:
+    sys.exit("No @context in supposed context file")
+
+print("prefix,base")
+for prefix_name, url_prefix in context['@context'].items():
+    print(f"{prefix_name},{url_prefix}")
--- a/template/.gitignore.jinja2
+++ b/template/.gitignore.jinja2
@@ -6,6 +6,8 @@ bin/
 *.tmp.obo
 *.tmp.owl
 *.tmp.json
+*-relation-graph.tsv.gz
+.template.db
 
 .github/token.txt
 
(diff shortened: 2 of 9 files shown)


scripts/context2csv.py
matentzn, line 1: Very smart idea.

template/.gitignore.jinja2
```diff
@@ -16,11 +18,13 @@ src/ontology/reports/*
 src/ontology/{{ project.id }}.owl
 src/ontology/{{ project.id }}.obo
 src/ontology/{{ project.id }}.json
+src/ontology/{{ project.id }}.db
 src/ontology/{{ project.id }}-base.*
 src/ontology/{{ project.id }}-basic.*
 src/ontology/{{ project.id }}-full.*
 src/ontology/{{ project.id }}-simple.*
 src/ontology/{{ project.id }}-simple-non-classified.*
+src/ontology/config/context.csv
```
matentzn, line 27: I saw your own preferences - I personally like this path here, but tbh I am fine with src/config too - whatever makes more sense. I have used src/ontology/config widely but this will continue to work even if ODK related config appears in `src/config`.

template/src/ontology/Makefile.jinja2
```diff
@@ -1058,6 +1061,19 @@ $(TRANSLATIONSDIR)/%.babelon.json: $(TRANSLATIONSDIR)/%.babelon.tsv
 {% endif -%}
 {% endfor -%}
 
+{% if 'db' in project.export_formats -%}
+{% if project.use_context -%}
```
matentzn, line 1065: What a great idea to piggy bag on `use_context`!!

matentzn (approved):
Awesome!

template/.gitignore.jinja2
gouttegd: > I have used src/ontology/config widely

OK, let’s keep it there then. I’ll fix the `_dynamic_files.jinja2` template accordingly.

template/src/ontology/Makefile.jinja2
gouttegd: Better to have only one source of truth for the prefixes. ;)

PR #1150, opened 2024-11-29
Fix location of generated context.json file.

gouttegd:
This PR ensures that, when seeding, if the project is configured to use a context, the initial context file is created in `src/ontology/config`, where the Makefile expects to find it.

While we are at it, and related, we fix two small issues with the #1148 PR:

* The `context.csv` file is a file that is _derived_ from the `context.json` file; as such, it should be written into the `tmp` directory, rather than into the `config` directory. Having such a file in `config` could lead users to think they can (or are even expected to) edit that file, while they are only supposed to edit the `context.json` file.

* The target to create a SQLite file should depend on the CSV context file _only_ when the project is configured to use a context.


matentzn (approved):
looks great!

PR #1155, opened 2024-12-12
Make sure SPARQL checks cover the entire ontology.

### LINKED ISSUE #1154: `sparql_test` does not cover components
The standard Makefile includes a `sparql_test` target (executed as part of the `test` target) which runs a bunch of SPARQL-based checks using `robot verify`:

```Makefile
sparql_test: $(EDIT_PREPROCESSED) | $(REPORTDIR)
        $(ROBOT) verify -i $(EDIT_PREPROCESSED) --queries $(SPARQL_VALIDATION_QUERIES) -O $(REPORTDIR)
```

Those checks only operate on the contents of the (preprocessed) `-edit` file; they do _not_ cover any imported ontology (actual ODK imports and components). This may not be explicitly documented in ROBOT’s manual, but a [brief look at the code](https://github.com/ontodev/robot/blob/35b830fac23056203c89b3b1984902b61cd07e58/robot-command/src/main/java/org/obolibrary/robot/VerifyCommand.java#L116) shows this is the normal behaviour of the `verify` command, and that this behaviour is not configurable:

```java
// Load into dataset without imports
Dataset dataset = QueryOperation.loadOntologyAsDataset(state.getOntology(), false);
```

I don’t think it is correct for the `sparql_test` to cover only the contents of the `-edit` file. Of course it is fine to exclude _imports_ (in the ODK sense) from the SPARQL tests (we may not want those tests to fail because of something coming from an external ontology), but _components_ (and I include under that umbrella the pattern-derived `definitions.owl` file) are merely parts of the “local” ontology that just happen to be in different files -- I see no reason for excluding them from the SPARQL tests.

Therefore, and since `verify` will not follow OWL imports declarations, I believe the `sparql_test` should be executed over the `$(SRCMERGED)` file (in which components, that is, all files listed as `$(OTHER_SRC)`, have been explicitly merged in) rather than `$(EDIT_PREPROCESSED)`.

gouttegd:
ROBOT's `verify` command does not follow OWL Import declarations, so the checks it performs only applies strictly to the file it is given as input.

Components and pattern-derived definitions are part of the "local" ontology and should be covered by the SPARQL checks, and we ensure they are by running the `verify` command on the `$(SRCMERGED)` intermediate file, in which components and pattern-derived definitions (basically all files listed in $(OTHER_SRC)`) are explicitly merged (but _not_ ODK imports).

closes #1154

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,7 +29,7 @@ For more detailed changes see:
 ## Bugfixes
 
 - Added back `class-count-by-prefix.sparql` ([#1030](https://github.com/INCATools/ontology-development-kit/issues/1030)).
-- Disabled `table-reader` mkdocs plygin ([#1028](https://github.com/INCATools/ontology-development-kit/issues/1028)).
+- Disabled `table-reader` mkdocs plugin ([#1028](https://github.com/INCATools/ontology-development-kit/issues/1028)).
 - SPARQL-based tests on the `-edit` file now also cover any component included in the ontology, as well as any pattern-derived contents ([#1154](https://github.com/INCATools/ontology-development-kit/issues/1154)).
 
 # v1.5
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,8 +28,9 @@ For more detailed changes see:
 
 ## Bugfixes
 
-- Added back `class-count-by-prefix.sparql` ([https://github.com/INCATools/ontology-development-kit/issues/1030](#1030))
-- Disabled `table-reader` mkdocs plygin ([https://github.com/INCATools/ontology-development-kit/issues/1028](#1028))
+- Added back `class-count-by-prefix.sparql` ([#1030](https://github.com/INCATools/ontology-development-kit/issues/1030)).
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -28,8 +28,9 @@ For more detailed changes see:
 
 ## Bugfixes
 
-- Added back `class-count-by-prefix.sparql` ([https://github.com/INCATools/ontology-development-kit/issues/1030](#1030))
-- Disabled `table-reader` mkdocs plygin ([https://github.com/INCATools/ontology-development-kit/issues/1028](#1028))
+- Added back `class-count-by-prefix.sparql` ([#1030](https://github.com/INCATools/ontology-development-kit/issues/1030)).
+- Disabled `table-reader` mkdocs plygin ([#1028](https://github.com/INCATools/ontology-development-kit/issues/1028)).
```
matentzn, line 32:
```suggestion
- Disabled `table-reader` mkdocs plugin ([#1028](https://github.com/INCATools/ontology-development-kit/issues/1028)).
```

PR #1160, opened 2024-12-28
Make Ammonite cache its files in TMPDIR.

### LINKED ISSUE #1159: `validate_idranges` target is broken
Trying to validate the ID range file yields an error from Ammonite:

```sh
$ sh run.sh make validate_idranges
amm ..scripts/validate_id_ranges.sc uberon-idranges.owl
Failed to resolve ivy dependencies:/tools/.coursier-cache
make: *** [Makefile:986: validate_idranges] Error 1
```

Ammonite is seemingly trying to download and store the dependencies of the `validate_idranges` script into `/tools/.coursier-cache`, which it cannot do without root privileges.

gouttegd:
When running a Scala script, Ammonite attempts to download and store the script's dependencies in `/tools/.coursier-cache`, which requires root privileges.

We force Ammonite to write its cache in the current ontology's `TMPDIR` instead.

closes #1159

### DIFF
--- a/template/src/ontology/Makefile.jinja2
+++ b/template/src/ontology/Makefile.jinja2
@@ -1333,8 +1333,8 @@ normalize_src: $(SRC)
 	$(ROBOT) convert -i $< -f {% if 'obo' == project.edit_format %}obo --check false{% else %}ofn{% endif %} -o $(TMPDIR)/normalise && mv $(TMPDIR)/normalise $<
 
 .PHONY: validate_idranges
-validate_idranges:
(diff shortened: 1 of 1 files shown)


matentzn (approved):
Looks great, have not tested it!

gouttegd:
I have. :)

But this fix is only a stop-gap measure. Ultimately, I want to replace the `validate_id_ranges.sc` script by a better tool to manage the ID range file. I’ve started working on that [here](https://github.com/gouttegd/dicer).

PR #1163, opened 2025-01-04
Replace the ID range validation script

gouttegd:
This PR replaces the `validate_id_ranges.sc` Scala script by a program dedicated to the management of ID range files, [dicer-cli](https://github.com/gouttegd/dicer).

Rationale for this replacement:

* Better validation: `dicer-cli` detects more errors than `validate_id_ranges.sc`, which only detects _overlapping ranges_. For example, [this error in CL](https://github.com/obophenotype/cell-ontology/pull/2871) was only detected by `dicer-cli`.
* No need for downloads at runtime. Executing the `validate_id_ranges.sc` script requires the Scala interpreter to download the script’s dependencies (such as the OWLAPI library) at runtime. The `dicer-cli` tool is self-sufficient.
* Additional features. In addition to validating the ID range file, `dicer-cli` can also be used to easily add new ranges to the file.

With the `validate_id_ranges.sc` script removed, there is no longer any standard, ODK-generated workflow that requires a Scala interpreter, so Ammonite is moved from ODKLite to ODKFull.


matentzn (approved):
Big yes!

I wonder if it would make sense in cases like this to include a routine in the repo update process to clean up files that are no longer needed, like

template/src/scripts/validate_id_ranges.sc

Just a thought, not a blocker for this PR.

gouttegd:
> I wonder if it would make sense in cases like this to include a routine in the repo update process to clean up files that are no longer needed

Sounds like a good idea indeed.

Issue #1162, opened 2025-01-03
Consider replacing Ammonite by Scala-CLI


gouttegd:
[Scala-CLI](https://scala-cli.virtuslab.org/) is apparently intended to become the “official” Scala runner, and could therefore replace Ammonite to run Scala scripts.

We should consider adding it to _ODKFull_, possibly **in parallel** to Ammonite to begin with, and then in a later version as the sole Scala runner.

PR #1164, opened 2025-01-05
Update some ODK programs.

gouttegd:
Update some of the programs bundled with the ODK:

* Soufflé to version 2.4.1;
* KGCL-Java to version 0.5.1;
* SSSOM-Java to version 1.1.1.

Also add a note to remind us that we *cannot* update Jena to version 5, as that version requires Java >= 17 and the ODK base image still provides Java 11 for now.

There is a 3.0.0 version of Ammonite since september 2024, but we do not upgrade it for now because (1) I don't know or use Ammonite enough (at all, actually) to know whether the new major version is likely to break any existing custom workflow that requires Ammonite, and (2) Jim Balhoff is keen to migrate from Ammonite to Scala-CLI anyway (#1162).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,9 +10,10 @@ ARG ODK_VERSION 0.0.0
 ENV ODK_VERSION=$ODK_VERSION
 
 # Software versions
+# Jena 5.x requires Java 17, so for now we are stuck with Jena 4.x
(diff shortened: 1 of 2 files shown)


matentzn (approved):
Thanks

PR #1165, opened 2025-01-06
Replace Ammonite with Scala-CLI

### LINKED ISSUE #1161: Revise need for global `COURSIER_CACHE` environment variable
The main Dockerfile (for the _ODKFull_ image) defines the following environment variable:

```
# Avoid repeated downloads of script dependencies by mounting the local coursier cache:
ENV COURSIER_CACHE="/tools/.coursier-cache"
```

This variable instructs Ammonite (Scala interpreter) to download its dependencies in `/tools/.coursier-cache`, rather than in the default location of `$HOME/.cache/coursier`. As the comment indicates, this is done so that a user can choose to reuse their own local cache (from their own home directory on the host system) in order to share the cache across ODK invocations.

This is both broken and no longer necessary. Broken because the `odkuser` account has no write access to the `/tools` directory, and no longer necessary because the ODK image now comes with a `/home/odkuser` directory that is owned by the `odkuser`, so whoever wishes to reuse their own local Coursier cache with the ODK container can simply bind their `$HOME/.cache/coursier` directory to the `/home/odkuser/.cache/coursier` directory within the container.

### LINKED ISSUE #1162: Consider replacing Ammonite by Scala-CLI

gouttegd:
This PR adds the [Scala-CLI](https://scala-cli.virtuslab.org) runner to the ODKFull image, as a replacement for Ammonite.

closes #1161
closes #1162

### DIFF
--- a/scripts/entrypoint.sh
+++ b/scripts/entrypoint.sh
@@ -13,6 +13,10 @@ mkdir -p /home/odkuser
 chown odkuser:odkuser /home/odkuser
 [ -d /home/odkuser/.data ] && chown odkuser:odkuser /home/odkuser/.data
 [ -d /home/odkuser/.data/oaklib ] && chown odkuser:odkuser /home/odkuser/.data/oaklib
+if [ ! -d /tools/.coursier-cache ] ; then
+    mkdir /tools/.coursier-cache
+    chown odkuser:odkuser /tools/.coursier-cache
+fi
 PATH=$PATH:/home/odkuser/.local/bin
 [ -S /run/host-services/ssh-auth.sock ] && chown odkuser /run/host-services/ssh-auth.sock
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,8 +14,7 @@ For more detailed changes see:
 - New [ROBOT Version 1.9.6](https://github.com/ontodev/robot/releases/tag/v1.9.6). This came with a great number of updates and upgrades, see release notes.
 - J2cli, a command-line tool to process Jinja2 templates, has been replaced by [Jinjanator](https://github.com/kpfleming/jinjanator). If your custom workflows invoke the `j2` tool, you will need to update them to use `jinjanate` instead.
 - New program `dicer-cli` to manage the ID range file.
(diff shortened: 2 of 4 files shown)


scripts/entrypoint.sh
```diff
@@ -13,6 +13,10 @@ mkdir -p /home/odkuser
 chown odkuser:odkuser /home/odkuser
 [ -d /home/odkuser/.data ] && chown odkuser:odkuser /home/odkuser/.data
 [ -d /home/odkuser/.data/oaklib ] && chown odkuser:odkuser /home/odkuser/.data/oaklib
+if [ ! -d /tools/.coursier-cache ] ; then
+    mkdir /tools/.coursier-cache
+    chown odkuser:odkuser /tools/.coursier-cache
+fi
```
matentzn, line 19: You removed Ammonite, but build logic to handle coursier cache? Is `/tools/.coursier-cache` used for anything other than `ammonite`?

matentzn (approved):
Apart from a little non-breaking comment LGTM

scripts/entrypoint.sh
gouttegd: Scala-CLI uses it as well. :)

gouttegd:
@balhoff I would appreciate your input on this PR before merging it. In particular, do you think we can remove Ammonite immediately? You said on Slack that switching to Scala-CLI would require Scala users to adapt their scripts – if the migration is thought to be a difficult one, I am open to the idea of having both Ammonite and Scala-CLI co-existing in ODK 1.6, and removing Ammonite only in ODK 1.7.

gouttegd:
> If you can have both in one release, then drop Ammonite, that would be amazing.

OK, then I will merge only the first two commits of that PR (adding Scala-CLI, but leaving Ammonite in place). We won’t remove Ammonite until ODK 1.7.

gouttegd:
Closing in favour if #1166, which does include the Ammonite removal.

PR #1166, opened 2025-01-07
Add Scala-CLI, but do not remove Ammonite yet

### LINKED ISSUE #1161: Revise need for global `COURSIER_CACHE` environment variable

gouttegd:
This PR supersedes #1165. It also adds Scala-CLI to ODKFull, but leaves Ammonite in place, with a deprecation notice warning users of its future complete removal (probably in ODK 1.7).

fixes #1161
partially addresses #1162

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ For more detailed changes see:
 - New [ROBOT Version 1.9.6](https://github.com/ontodev/robot/releases/tag/v1.9.6). This came with a great number of updates and upgrades, see release notes.
 - J2cli, a command-line tool to process Jinja2 templates, has been replaced by [Jinjanator](https://github.com/kpfleming/jinjanator). If your custom workflows invoke the `j2` tool, you will need to update them to use `jinjanate` instead.
 - New program `dicer-cli` to manage the ID range file.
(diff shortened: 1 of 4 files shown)


gouttegd:
Merging without review (YOLO!) since the original PR (#1165) has already been reviewed and approved.

PR #1168, opened 2025-01-13
Use Make conditionals

### LINKED ISSUE #912: Schnapsidee: Favour Make conditionals over shell conditionals
The standard ODK-generated Makefile contains several instances where the entire recipe of a rule is enclosed in a shell conditional construct (`if... then... fi`) based on a Make variable, such as in this example:

```Makefile
(TMPDIR)/pattern_owl_seed.txt: $(PATTERNDIR)/pattern.owl
        if [ $(PAT) = true ]; then $(ROBOT) query --use-graphs true -f csv -i $< --query ../sparql/terms.sparql $@; fi
```

Since what is tested in the condition is a Make variable, not a shell variable, this kind of construct could be replaced by a _Make conditional_ instead:

```Makefile
ifeq ($(PAT),true)
(TMPDIR)/pattern_owl_seed.txt: $(PATTERNDIR)/pattern.owl
        $(ROBOT) query --use-graphs true -f csv -i $< --query ../sparql/terms.sparql $@
endif
```

I think this would be less surprising to readers, and easier to debug when things go wrong.

We’d need to be careful about prerequisites, though, because the two above constructs are _not_ strictly equivalent: in the first case (shell conditional), even if the _recipe_ does nothing when `PAT` is false, the _rule_ is still there and its prerequisites are evaluated (so here, Make will find that it needs to build `$(PATTERNDIR)/pattern.owl`), whereas in the second case (Make conditional), the entire rule is skipped when `PAT` is false, including its prerequisites. But we should always be careful about prerequisites anyway, whether we use shell or Make conditionals.

This idea would also apply to the various `custom.Makefile` that exist in the wild, where constructs similar to those in use in the ODK Makefile are also widely used.

gouttegd:
This PR updates the standard Makefile template to replace, wherever possible, shell conditionals by Make conditionals.

That is, rules like the following:

```Make
my_target: my_dependencies
    if [ $(MY_VAR) = true ]; then <code to generate my_target> ; fi

my_other_target: my_other_dependencies
    if [ $(MY_VAR) = true ]; then <code to generate my_other_target> ; fi
```

is replaced by

```Make
ifeq ($(MY_VAR),true)
my_target: my_dependencies
    <code to generate my_target>

my_other_target: my_other_dependencies
    <code to generate my_other_target>
endif
```

closes #912

### DIFF
--- a/template/src/ontology/Makefile.jinja2
+++ b/template/src/ontology/Makefile.jinja2
@@ -464,23 +464,30 @@ ANNOTATION_PROPERTIES={% for p in project.import_group.annotation_properties %}{
 # These live in the imports/ folder
 # This pattern uses ROBOT to generate an import module
 
+# Set to false to bypass all code to generate/refresh import modules
+IMP = true
+
+# Likewise, but for "large" modules only
+IMP_LARGE = true
+
+ifeq ($(IMP),true)
 # Should be able to drop this if robot can just take a big messy list of terms as input.
 $(IMPORTDIR)/%_terms_combined.txt: $(IMPORTSEED) $(IMPORTDIR)/%_terms.txt
-	if [ $(IMP) = true ]; then cat $^ | grep -v ^# | sort | uniq >  $@; fi
+	cat $^ | grep -v ^# | sort | uniq >  $@
 
 {% if project.import_group.use_base_merging %}
 ALL_TERMS_COMBINED = $(patsubst %, $(IMPORTDIR)/%_terms_combined.txt, $(IMPORTS))
 $(IMPORTDIR)/merged_terms_combined.txt: $(ALL_TERMS_COMBINED)
-	if [ $(IMP) = true ]; then cat $^ | grep -v ^# | sort | uniq >  $@; fi
+	cat $^ | grep -v ^# | sort | uniq >  $@
 
 {% if 'slme' == project.import_group.module_type %}
 $(IMPORTDIR)/merged_import.owl: $(MIRRORDIR)/merged.owl $(IMPORTDIR)/merged_terms_combined.txt
-	if [ $(IMP) = true ]; then $(ROBOT) merge -i $< \{% if project.import_group.exclude_iri_patterns is not none %}
+	$(ROBOT) merge -i $< \{% if project.import_group.exclude_iri_patterns is not none %}
 		{% for pattern in project.import_group.exclude_iri_patterns %}remove  --select "{{ pattern }}" {% endfor %} \{% endif %}
 		extract -T $(IMPORTDIR)/merged_terms_combined.txt --force true --copy-ontology-annotations true --individuals {{ project.import_group.slme_individuals }} --method {{ project.import_group.module_type_slme }} \
 		remove $(patsubst %, --term %, $(ANNOTATION_PROPERTIES)) -T $(IMPORTDIR)/merged_terms_combined.txt --select complement --select annotation-properties \
 		query --update ../sparql/inject-subset-declaration.ru --update ../sparql/inject-synonymtype-declaration.ru --update ../sparql/postprocess-module.ru \
-		$(ANNOTATE_CONVERT_FILE); fi
+		$(ANNOTATE_CONVERT_FILE)
 {% else %}
 $(IMPORTDIR)/merged_import.owl: $(MIRRORDIR)/merged.owl $(IMPORTDIR)/merged_terms_combined.txt
 	echo "ERROR: You have configured your default module type to be {{project.import_group.module_type}}; this behavior needs to be overwritten in {{ project.id }}.Makefile!" && false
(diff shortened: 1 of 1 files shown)


template/src/ontology/Makefile.jinja2
```diff
@@ -621,47 +637,49 @@ $(COMPONENTSDIR)/%.owl: | $(COMPONENTSDIR)
...
 {% if project.import_group is defined -%}
 # ----------------------------------------
 # Mirroring upstream ontologies
 # ----------------------------------------
 
-IMP=true # Global parameter to bypass import generation
-MIR=true # Global parameter to bypass mirror generation
-IMP_LARGE=true # Global parameter to bypass handling of large imports
+# Set to false to bypass all code to create/refresh mirrors
+MIR=true
```
matentzn, line 680: Is it a problem that the MIR variable is declared after it is being used by the COMP section?

template/src/ontology/Makefile.jinja2
```diff
@@ -707,17 +725,14 @@ mirror-{{ ont.id }}: | $(TMPDIR)
 {%- endif %}
 {%- endif %}
 {%- if ont.is_large %}
-else
-mirror-{{ ont.id }}:
-	@echo "Not refreshing {{ ont.id }} because refreshing large imports is disabled (IMP_LARGE=$(IMP_LARGE))."
```
matentzn, line 712: Some message at run.sh level indicating the IMP MIR COMP parameters might be useful when the job is finished but no sweat. This can help for example when custom workflows with parameter configs are defined (like prepare_release_fast with IMP = false), and you just want to tell the use: just so you know, you didn't refresh imports.

template/src/ontology/Makefile.jinja2
```diff
@@ -781,7 +793,11 @@ endif
 # DOSDP Templates/Patterns
 # ----------------------------------------
 
-PAT=true # Global parameter to bypass pattern generation
+# Set to false to bypass all code to generate/refresh patterns
+PAT = true
+
+ifeq ($(PAT),true)
```
matentzn, line 799: this will be the most useful for all the parameter refactorings you did.. :D

template/src/ontology/Makefile.jinja2
```diff
@@ -889,40 +905,44 @@ dosdp-docs-{{ pipeline.id }}: $(EDIT_PREPROCESSED) $(DOSDP_TSV_FILES_{{ pipeline
...
-	if [ $(PAT) = true ] && [ "${DOSDP_PATTERN_NAMES_DEFAULT}" ] {% if project.pattern_pipelines_group is defined %} {% for pipeline in project.pattern_pipelines_group.products %} || [ "${DOSDP_PATTERN_NAMES_{{ pipeline.id.upper() }}}" ]{% endfor %}{% endif %} && [ $(PAT) = true ]; then $(ROBOT) merge $(addprefix -i , $^) \
+	if [ "${DOSDP_PATTERN_NAMES_DEFAULT}" ] {% if project.pattern_pipelines_group is defined %} {% for pipeline in project.pattern_pipelines_group.products %} || [ "${DOSDP_PATTERN_NAMES_{{ pipeline.id.upper() }}}" ]{% endfor %}{% endif %} && [ $(PAT) = true ]; then $(ROBOT) merge $(addprefix -i , $^) \
 		annotate --ontology-iri $(ONTBASE)/patterns/definitions.owl  --version-iri $(ONTBASE)/releases/$(TODAY)/patterns/definitions.owl \
       --annotation owl:versionInfo $(VERSION) -o definitions.ofn && mv definitions.ofn $@; fi
 
+else # PAT=false
+# Even if pattern generation is disabled, we still extract a seed from definitions.owl
```
matentzn, line 942: Is this necessary under IMP=false?

matentzn (commented):
Ohhh thanks that will be soo much help for me.

Just a few minor comments, reassign me to approve.

template/src/ontology/Makefile.jinja2
gouttegd: Good catch. It is indeed – until it is declared (and unless of course `Make` is invoked with a declaration on the command line), `MIR` would not exist, so the test in the COMP section would always evaluate to false.

Better to declare all MIR, IMP, IMP_LARGE, COMP, and PAT variables all at the same place, somewhere near the top of the Makefile.

template/src/ontology/Makefile.jinja2
gouttegd: Why not, but then displaying the message at the moment a target is skipped would not be very useful – the message would undoubtedly be lost in the middle of the entire Make output.

If we want such a message, better to do that explicitly at the end of a workflow – something like:

```Make
prepare_release: all_odk
      [ ... normal commands of the prepare_release target ... ]
      @echo "Release files are not in $(RELEASEDIR) - now you should commit, push and make a release on your Git hosting site."
      @if [ "$(IMP)" != true ]; then echo "Note: Imports were not refreshed." ; fi
```

matentzn: Yes that is probably better. I was thinking of something at odk wrapper level to always print this information but that is silly since many pipelines do not care about these three variables.. Thanks!

template/src/ontology/Makefile.jinja2
gouttegd: 1. This was the original behaviour:

```sh
if [ $(PAT) = true ]; then cat $^ | sort | uniq > $@; else $(ROBOT) query --use-graphs true -f csv -i $(PATTERNDIR)/definitions.owl --query ../sparql/terms.sparql $@; fi
```

If `PAT` is _false_ then we run the `terms.sparql` query on the `definitions.owl` file. I didn’t change that behaviour, just rewrote the rule.

2. Under `IMP=false`, there is no target that ultimately depends on `$(TMPDIR)/all_pattern_terms.txt`, so the rule would never be invoked anyway. It would only be invoked under `IMP=true PAT=false`.

matentzn: Good enough, thanks for clarificatiosn!

matentzn (approved):
Looks great, didnt do testing, just code review. THANKS!

gouttegd:
Regarding the need for testing: I’d say that it can be done as part of the normal testing we do prior to a major release (as outlined [here](https://github.com/INCATools/ontology-development-kit/blob/master/CONTRIBUTING.md#sop-for-creating-a-major-release). During development, it’s OK to rely mostly on the CI-based testing for most PRs.

Issue #1169, opened 2025-01-13
Consider replacing SPARQL pre/post-processing scripts by ROBOT pluggable commands


gouttegd:
One of the steps that is performed when preparing an import module is the execution of the `src/sparql/preprocess-module.ru` SPARQL query, which simply sets a `dc:source` ontology annotation:

```sparql
INSERT { 
    ?ontology dc:source ?version_iri .
}

WHERE {
  ?ontology rdf:type owl:Ontology ;
        owl:versionIRI ?version_iri .
}
```

Running a SPARQL query with ROBOT requires (1) converting the entire ontology in memory from the OWLAPI model to the JENA model, (2) actually running the query on the JENA model, (3) serialising the JENA model to Turtle, and (4) parsing the serialised Turtle back into a OWLAPI model. Doing all of that for a task that could be done in 3 lines of Java code is a huge waste of resources.

In fact, the cost of that process is a big reason why some import modules need to be marked as “large”, so that we can selectively avoid that step for those modules:

```jinja2
$(ROBOT) \
    {% if ont.is_large %}extract -i $< -T $(IMPORTDIR)/{{ ont.id }}_terms_combined.txt --copy-ontology-annotations true --force true --method BOT \
    {% else %}query -i $< --update ../sparql/preprocess-module.ru extract -T $(IMPORTDIR)/{{ ont.id }}_terms_combined.txt --copy-ontology-annotations true --force true --method BOT {% endif %}
```

A simple ROBOT plugin could easily replace the SPARQL query. This would make the preprocessing much faster for all modules (since we avoid the whole roundtrip to JENA and back), and would remove the need for treating “large” imports differently, which would in turn make the Makefile template easier to read and maintain.

Likewise for the post-processing step, which simply strips all ontology annotations except the `dc:source` annotation.

matentzn:
I fully agree, this would be great. Any thoughts on avoiding proliferation of robot plugins? maybe an ODK - utils one for such cases?

gouttegd:
> maybe an ODK - utils one for such cases?

That’s what I was thinking of, yes. Since it would be used in standard ODK workflows, a `odk` plugin would definitely be appropriate.

And such a plugin, once it exists, could become the landing site for other pluggable commands that were initially created elsewhere but were found to be useful enough to be re-used across several repositories (such as FlyBase’s `rewrite-definition`, initially written for FlyBase but is now also used by CL).

gouttegd:
Done with #1174

Issue #622, opened 2022-06-20
Remove dependency to OWLTools in standard workflows?


gouttegd:
While most of the standard, ODK-generated workflows use ROBOT, there are still rules where OWLTOOLS is used:

```Makefile
$(SUBSETDIR)/%.owl: $(ONT).owl | $(SUBSETDIR)
        $(OWLTOOLS) $< --extract-ontology-subset --fill-gaps --subset $* -o $@.tmp.owl && mv $@.tmp.owl $@ &&\
        $(ROBOT) annotate --input $@ --ontology-iri $(ONTBASE)/$@ $(ANNOTATE_ONTOLOGY_VERSION) -o $@.tmp.owl && mv $@.tmp.owl $@
```

```Makefile
normalize_obo_src: $(SRC)
        $(OWLTOOLS) $< --merge-axiom-annotations -o -f obo $(TMPDIR)/NORM.obo && $(ROBOT) convert -i $(TMPDIR)/NORM.obo -o $(TMPDIR)/NORM.tmp.obo && mv $(TMPDIR)/NORM.tmp.obo $(SRC)
```

Since we want to make it clear that ROBOT is the modern tool and that OWLTOOLS are no longer maintained, we should probably investigate how to replace the use of OWLTOOLS in those rules.

gouttegd:
In the ODK’s standard workflow, `owltools` is still used in two places.

# Subset generation
Subset ontologies are produced with the `owltools --extract-ontology-subset` command:

```make
$(SUBSETDIR)/%.owl: $(ONT).owl | $(SUBSETDIR)
        $(OWLTOOLS) $< --extract-ontology-subset --fill-gaps --subset $* -o $@.tmp.owl && mv $@.tmp.owl $@ &&\
        $(ROBOT) annotate --input $@ --ontology-iri $(ONTBASE)/$@ $(ANNOTATE_ONTOLOGY_VERSION) -o $@.tmp.owl && mv $@.tmp.owl $@
```

What the `--extract-ontology-subset` command does mostly resides in the `makeMinimalSubsetOntology` method of owltools’ [owltools.mooncat.Mooncat class](https://github.com/owlcollab/owltools/blob/master/OWLTools-Core/src/main/java/owltools/mooncat/Mooncat.java), but basically it extracts a coherent, minimal subset ontology around the terms carrying a given `oboInOwl:inSubset` annotation.

It does not seem that this command can easily be replaced by other existing tools. I tried two approaches:

[...]

PR #1174, opened 2025-02-03
Replace SPARQL queries and OWLTools by a ROBOT plugin

gouttegd:
This PR is intended to experiment using a ROBOT plugin to replace both

(1) SPARQL queries (as suggested in #1169), and

(2) OWLTools (still used in standard workflows for two things: normalising a OBO source file, and creating subsets -- #622).

For now, this is using my own “[experimental ROBOT plugin](https://github.com/gouttegd/exp-robot-plugin)”. If we are happy with the experiment, we can then create a proper ODK plugin for ROBOT (and/or push some of the features in upstream ROBOT).


template/src/sparql/inject-synonymtype-declaration.ru.jinja2
matentzn, line 1: We have to be careful here of one thing: Not to _delete_ this file in repos during some future implementation `update_repo` cleanup. Currently I think this is ok.

The reason, I know for a fact that this file (and the subset one) is used in at least 8 or so `custom.Makefile` (basically everytime someone had to _customise_ a `imports/*_import.owl` goal).

template/src/ontology/Makefile.jinja2
```diff
@@ -290,7 +288,8 @@ all_mappings: $(MAPPING_FILES)
 # ----------------------------------------
 
 OBO_REPORT = {% for x in project.robot_report.report_on|default(["edit"]) %} {% if x=="edit" %}$(SRC){% else %}{{ x }}{% endif %}-obo-report{% endfor %}
-REPORTS = $(OBO_REPORT)
+ALIGNMENT_REPORT = {% for x in project.robot_report.report_on|default(["edit"]) %} {% if x =="edit" %}$(SRC){% else %}{{ x }}{% endif %}-align-report{% endfor %}
```
matentzn, line 291: 😱

template/src/ontology/Makefile.jinja2
```diff
@@ -342,6 +341,26 @@ $(REPORTDIR)/$(SRC)-obo-report.tsv: $(SRCMERGED) | $(REPORTDIR)
 
 $(REPORTDIR)/%-obo-report.tsv: % | $(REPORTDIR)
 	$(ROBOT) report -i $< $(REPORT_LABEL) $(REPORT_PROFILE_OPTS) --fail-on $(REPORT_FAIL_ON) {% if project.robot_report.use_base_iris %}{% if project.namespaces is not none %}{% for iri in project.namespaces %}--base-iri {{ iri }} {% endfor %}{% else %}--base-iri $(URIBASE)/{{ project.id.upper() }}_ --base-iri $(URIBASE)/{{ project.id }} {% endif %}{% endif -%} --print 5 -o $@
+{%- if project.robot_report.upper_ontology is defined and project.robot_report.upper_ontology %}
+
+$(REPORTDIR)/$(SRC)-align-report.tsv: $(SRCMERGED) | $(REPORTDIR) all_robot_plugins
+	$(ROBOT) odk:validate -i $< --reasoner $(REASONER) \
+		 --upper-ontology-iri {{ project.robot_report.upper_ontology }}
```
matentzn, line 348: This is a fantastic design decision as it allows us to use version IRIs.

template/src/ontology/Makefile.jinja2
```diff
@@ -342,6 +341,26 @@ $(REPORTDIR)/$(SRC)-obo-report.tsv: $(SRCMERGED) | $(REPORTDIR)
...
+		 {%- if project.robot_report.use_base_iris %} \
+		 {% if project.namespaces is not none %}{% for iri in project.namespaces %}--base-iri {{ iri }} \
+		 {% endfor %}{% else %}--base-iri $(URIBASE)/{{ project.id.upper() }}_ --base-iri $(URIBASE)/{{ project.id }} \
+		 {% endif %}{% else %} \
+		 {% endif %}--report-output $@
```
matentzn, line 353: I have some additional thoughts here (what would be useful in that report), not sure if you want to discuss this here?

template/src/ontology/Makefile.jinja2
```diff
@@ -512,30 +531,31 @@ $(IMPORTDIR)/merged_import.owl: $(MIRRORDIR)/merged.owl $(IMPORTDIR)/merged_term
...
-	$(ROBOT) query -i $< --update ../sparql/preprocess-module.ru \
-		extract -T $(IMPORTDIR)/$*_terms_combined.txt --force true --copy-ontology-annotations true --individuals {{ project.import_group.slme_individuals }} --method {{ project.import_group.module_type_slme }} \
-		query --update ../sparql/inject-subset-declaration.ru --update ../sparql/inject-synonymtype-declaration.ru --update ../sparql/postprocess-module.ru \
+$(IMPORTDIR)/%_import.owl: {% if project.import_group.use_base_merging %}$(MIRRORDIR)/merged.owl{% else %}$(MIRRORDIR)/%.owl{% endif %} $(IMPORTDIR)/%_terms_combined.txt | all_robot_plugins
+	$(ROBOT) odk:annotate -i $< --remove-all true --add-source true \
+		extract -T $(IMPORTDIR)/$*_terms_combined.txt --force true --copy-ontology-annotations false --individuals {{ project.import_group.slme_individuals }} --method {{ project.import_group.module_type_slme }} \
+		odk:normalize --base-iri {{ project.uribase }} --inject-subset-declarations true --inject-synonym-declarations true --merge-axioms true \
```
matentzn, line 537: This will be such a relief on memory consumption for larger ontologies! awesome!

template/src/ontology/Makefile.jinja2
```diff
@@ -498,12 +517,12 @@ $(IMPORTDIR)/merged_terms_combined.txt: $(ALL_TERMS_COMBINED)
...
 
 {% if 'slme' == project.import_group.module_type %}
-$(IMPORTDIR)/merged_import.owl: $(MIRRORDIR)/merged.owl $(IMPORTDIR)/merged_terms_combined.txt
+$(IMPORTDIR)/merged_import.owl: $(MIRRORDIR)/merged.owl $(IMPORTDIR)/merged_terms_combined.txt | all_robot_plugins
 	$(ROBOT) merge -i $< \{% if project.import_group.exclude_iri_patterns is not none %}
 		{% for pattern in project.import_group.exclude_iri_patterns %}remove  --select "{{ pattern }}" {% endfor %} \{% endif %}
 		extract -T $(IMPORTDIR)/merged_terms_combined.txt --force true --copy-ontology-annotations true --individuals {{ project.import_group.slme_individuals }} --method {{ project.import_group.module_type_slme }} \
 		remove $(patsubst %, --term %, $(ANNOTATION_PROPERTIES)) -T $(IMPORTDIR)/merged_terms_combined.txt --select complement --select annotation-properties \
-		query --update ../sparql/inject-subset-declaration.ru --update ../sparql/inject-synonymtype-declaration.ru --update ../sparql/postprocess-module.ru \
+		odk:normalize --base-iri {{ project.uribase }} --inject-subset-declarations true --inject-synonym-declarations true --merge-axioms true \
```
matentzn, line 525: I am mildly concerned of using `{{ project.uribase }}`; what exactly is this parameter used for in odk:normalize?

template/src/ontology/Makefile.jinja2
```diff
@@ -765,9 +786,9 @@ $(SUBSETDIR)/%.tsv: $(SUBSETDIR)/%.owl
 	$(ROBOT) query -f tsv -i $< -s ../sparql/labels.sparql $@
 .PRECIOUS: $(SUBSETDIR)/%.tsv
 
-$(SUBSETDIR)/%.owl: $(ONT).owl | $(SUBSETDIR)
-	$(OWLTOOLS) $< --extract-ontology-subset --fill-gaps --subset $* -o $@.tmp.owl && mv $@.tmp.owl $@ &&\
-	$(ROBOT) annotate --input $@ --ontology-iri $(ONTBASE)/$@ $(ANNOTATE_ONTOLOGY_VERSION) -o $@.tmp.owl && mv $@.tmp.owl $@
+$(SUBSETDIR)/%.owl: $(ONT).owl | $(SUBSETDIR) all_robot_plugins
+	$(ROBOT) odk:subset -i $< --subset $* --fill-gaps true --replace true \
```
matentzn, line 790: 😱 WOW! That _would_ be awesome, but... 

I was a bit surprised to see this owltools use anyways because I thought this might have been replaced by `robot extract -m subset`?

template/src/ontology/Makefile.jinja2
```diff
@@ -557,34 +577,35 @@ $(IMPORTDIR)/%_import.obo: $(IMPORTDIR)/%_import.owl
...
+		odk:normalize --base-iri {{ project.uribase }} --inject-subset-declarations true --inject-synonym-declarations true --merge-axioms true \
+		$(ANNOTATE_CONVERT_FILE)
 {% elif ('minimal' == ont.module_type) or (ont.module_type is none and 'minimal' == project.import_group.module_type) %}
-$(IMPORTDIR)/{{ ont.id }}_import.owl: {% if project.import_group.use_base_merging %}$(MIRRORDIR)/merged.owl{% else %}$(MIRRORDIR)/{{ ont.id }}.owl{% endif %} $(IMPORTDIR)/{{ ont.id }}_terms_combined.txt
-	$(ROBOT) {% if ont.is_large %}extract -i $< -T $(IMPORTDIR)/{{ ont.id }}_terms_combined.txt --copy-ontology-annotations true --force true --method BOT \{% else %}query -i $< --update ../sparql/preprocess-module.ru \
-    extract -T $(IMPORTDIR)/{{ ont.id }}_terms_combined.txt --copy-ontology-annotations true --force true --method BOT \{% endif %}
+$(IMPORTDIR)/{{ ont.id }}_import.owl: {% if project.import_group.use_base_merging %}$(MIRRORDIR)/merged.owl{% else %}$(MIRRORDIR)/{{ ont.id }}.owl{% endif %} $(IMPORTDIR)/{{ ont.id }}_terms_combined.txt | all_robot_plugins
+	$(ROBOT) odk:annotate -i $< --remove-all true --add-source true \
```
matentzn, line 605: what does `--remove-all true` do?

matentzn (commented):
this is sooooo awesome. I have a few questions (in code) and 1 general concern.

Is there any way you would agree to move the ODK robot plugin into the INCATools org? I would feel a bit better if that component that is shaping up to be a core component of the ODK build system would live a bit more visibly here..

gouttegd:
> Is there any way you would agree to move the ODK robot plugin into the INCATools org?

There is currently _no_ ODK ROBOT plugin. This PR is using my [experimental plugin](https://github.com/gouttegd/exp-robot-plugin) (emphasis on **experimental**), which nobody should ever think about using seriously in a production pipeline. That plugin is basically my playground, where I ~~have fun~~ test various ideas that may or may not turn out to be good ideas.

When there _will_ be a ROBOT plugin (using consolidated ideas from my experimental one), it could of course be hosted somewhere in the INCATools organisation. Though its Java namespaces would likely still be somewhere under `org.incenp.obofoundry`, since the “INCATools organisation” is merely a GitHub thing that does not have its own domain name.

template/src/ontology/Makefile.jinja2
gouttegd: ?

template/src/ontology/Makefile.jinja2
gouttegd: It also means that the validation step will download the upper ontology whenever it is needed, which in turns mean that the validation becomes susceptible to network glitches.

I’d like to recommend that projects that want to enable validation against an upper ontology take the supplementary steps of:

* downloading the upper ontology (in the version that they want) once and for all, and commit it somewhere in the repo;
* add an entry in their XML catalog so that the upper ontology IRI redirects to the local file.

template/src/ontology/Makefile.jinja2
gouttegd: Why not in the dedicated issue (#1175)?

template/src/ontology/Makefile.jinja2
gouttegd: Of note, the `odk:annotate` could hopefully be replaced by

```
annotate --remove-annotations --interpolate --link-annotation dc:source %{version_iri}
```

if the [interpolation feature](https://github.com/ontodev/robot/pull/1242) makes it way into upstream ROBOT.

template/src/ontology/Makefile.jinja2
gouttegd: Same thing as in the original `inject-subset-declaration.ru.jinja2` SPARQL query:

```sparql
INSERT { ?y rdfs:subPropertyOf oboInOwl:SubsetProperty . }

WHERE {
  ?x oboInOwl:inSubset  ?y .
  FILTER(isIRI(?y))
  FILTER(regex(str(?y),"^(http://purl.obolibrary.org/obo/)")
         || regex(str(?y),"^(http://www.ebi.ac.uk/efo/)")
         || regex(str(?y),"^(https://w3id.org/biolink/)")
         || regex(str(?y),"^({{ project.uribase }})"))
}
```

That is, it restricts the injection of the `SubAnnotationPropertyOf` axiom to subset properties that are in the namespace of the project (in addition to the built-namespaces `http://purl.obolibrary.org/obo/`, `http://www.ebi.ac.uk/efo/`, `https://w3id.org/biolink/`).

So a `oboInOwl:inSubset` annotation with a value of `http://example.org/external/namespace`, where `http://example.org/external/namespace` is unrelated to `{{ project.uribase }}`, would _not_ cause `http://example.org/external/namespace` to be classified as a subproperty of `oboInOwl:SubsetProperty`.

This is the original behaviour, which is unchanged by the use of the plugin.

Likewise for synonym type declarations.

template/src/ontology/Makefile.jinja2
gouttegd: It removes all existing ontology annotations (which is what the `postprocess-module.ru` query currently does).

I wrote that before realizing that the original `annotate` command in upstream ROBOT already has the same feature (`robot annotate --remove-annotations`). :facepalm:

template/src/ontology/Makefile.jinja2
gouttegd: You wish!

The `subset` method of `robot extract` only implements¹ the “gap spanning” mode of `owltools --extract-ontology-subset` (the default mode, used when the `--fill-gaps` option is _not_ specified). There is no way with ROBOT to reproduce the behaviour of the “gap filling” mode – I have already raised that issue [here](https://github.com/INCATools/ontology-development-kit/issues/622).

More generally, OWLTools has 3 different ways of creating an “ontology subset” (that I know of – there might even be more that I have never encoutered). And of course they all behave slightly differently, because where would be the fun otherwise?

* `--extract-ontology-subset --subset <SUBSET_NAME>`

Used to extract a subset defined by `oboInOwl:inSubset` annotations. This uses “gap spanning”, meaning basically that the subset only contains the classes defined in the subset, but if two classes are _indirectly_ related through several relationships that involve intermediate classes that are not in the subset, that relation will be preserved (a direct relation between the two classes will be automatically fabricated).

This is what the `-m subset` method of `robot extract` does.

* `--extract-ontology-subset --subset <SUBSET_NAME> --fill-gaps`

Also used to extract a subset defined by `oboInOwl:inSubset` annotations, but using “gap filling”, meaning basically that the initial list of classes in the subset is expanded to include any classes referred from within the subset. This creates a self-sufficient “closure”.

Not doable at all with ROBOT.

* `--reasoner-query <DL_QUERY> --make-ontology-from-results`

Used to create a subset defined by a DL query, such as `'part of' some 'nervous system'`. It doesn’t do any “gap filling” or “gap spanning”, but it does take care of including any object property used in the subset. It can optionally do a limited form of “gap filling”, but only for is-a relationships (including all parent classes all the way to `owl:Thing`), by asking the reasoner to include the ancestors that match the query (the default is to only include the equivalent and the subclasses).

Not doable at all with ROBOT.

My proposed `odk:subset` command implements the last two methods.

(I do wonder why the first method is the one that got implemented in ROBOT, as it seems to be the _least used_ method. It is not used at all in any ODK standard workflow, and even Uberon, which creates literally dozens of subsets, only uses it for _one_ subset.)

---
¹ Kind of. `robot extract -m subset` does not take care of retrieving the classes flagged as belonging to the subset, so that must be done by another step beforehand. Also, it requires an explicit list of properties to include in the subset, instead of automatically using all properties that are used by the classes of the subset.

gouttegd:
> There is currently no ODK ROBOT plugin.

Well, this is no longer true. :) The “[ODK ROBOT plugin](https://github.com/gouttegd/odk-robot-plugin)” exists now. Derived from my experimental plugin, but containing only the commands that are deemed useful for the ODK, not the more dubious experimental ones.

It is in my own account for now, but happy to transfer its ownership to the INCATools organisation if you prefer.

template/src/ontology/Makefile.jinja2
gouttegd: Could you elaborate your concerns here?

template/src/ontology/Makefile.jinja2
gouttegd: FYI, the [documentation of the `odk:subset` command](https://incatools.github.io/odk-robot-plugin/subset.html) has [more details](https://incatools.github.io/odk-robot-plugin/subset.html#Internals_and_comparison_with_OWLTools.2FROBOT_extract) on how the command works and why it is needed to fully replace OWLTools (i.e. why `robot extract -m subset` does not fulfil all needs).

template/src/ontology/Makefile.jinja2
matentzn: No concerns at all, just amaaaaazement!! That was a _fast turnaround_ even for you!

template/src/ontology/Makefile.jinja2
matentzn: Does `--upper-ontology-iri` respect the catalogue? That would really be all that is needed to address your concern..

template/src/ontology/Makefile.jinja2
```diff
@@ -512,30 +531,34 @@ $(IMPORTDIR)/merged_import.owl: $(MIRRORDIR)/merged.owl $(IMPORTDIR)/merged_term
...
+		odk:normalize --base-iri {{ project.uribase }} --subset-decls true --synonym-decls true --merge-axioms true \
 		remove $(patsubst %, --term %, $(ANNOTATION_PROPERTIES)) -T $(IMPORTDIR)/$*_terms_combined.txt --select complement --select "classes individuals annotation-properties" \
 		$(ANNOTATE_CONVERT_FILE)
 {% elif 'mirror' == project.import_group.module_type %}
-$(IMPORTDIR)/%_import.owl: {% if project.import_group.use_base_merging %}$(MIRRORDIR)/merged.owl{% else %}$(MIRRORDIR)/%.owl{% endif %} $(IMPORTDIR)/%_terms_combined.txt
-	$(ROBOT) merge -i $< query --update ../sparql/preprocess-module.ru --update ../sparql/inject-subset-declaration.ru --update ../sparql/inject-synonymtype-declaration.ru --update ../sparql/postprocess-module.ru \
+$(IMPORTDIR)/%_import.owl: {% if project.import_group.use_base_merging %}$(MIRRORDIR)/merged.owl{% else %}$(MIRRORDIR)/%.owl{% endif %} $(IMPORTDIR)/%_terms_combined.txt | all_robot_plugins
+	$(ROBOT) annotate -i $< --remove-annotations \
+		odk:normalize --base-iri {{ project.uribase }} --subset-decls true --synonym-decls true --merge-axioms true --add-source true \
```
matentzn, line 552: for some reason i have come to like long variable names like `--subset-declarations` but I guess your suggestion is well positioned somewhere between Chris `-s` and my extreme stance.

template/src/ontology/Makefile.jinja2
```diff
@@ -512,30 +531,34 @@ $(IMPORTDIR)/merged_import.owl: $(MIRRORDIR)/merged.owl $(IMPORTDIR)/merged_term
...
 		$(ANNOTATE_CONVERT_FILE)
 {% elif 'filter' == project.import_group.module_type %}
-$(IMPORTDIR)/%_import.owl: {% if project.import_group.use_base_merging %}$(MIRRORDIR)/merged.owl{% else %}$(MIRRORDIR)/%.owl{% endif %} $(IMPORTDIR)/%_terms_combined.txt
+$(IMPORTDIR)/%_import.owl: {% if project.import_group.use_base_merging %}$(MIRRORDIR)/merged.owl{% else %}$(MIRRORDIR)/%.owl{% endif %} $(IMPORTDIR)/%_terms_combined.txt | all_robot_plugins
 	$(ROBOT) merge -i $< \
-		query --update ../sparql/preprocess-module.ru \
+		annotate --remove-annotations \
```
matentzn, line 557: never knew that existed

matentzn (approved):
Fantastic! Looks great. 

I am approving this but under a the tiny reservation that the _exact_ nature of the odk:validate command is still a bit in flux (i.e. we would label this feature as "experimental" if we were to make a release tomorrow)

template/src/ontology/Makefile.jinja2
gouttegd: Yes it does. I wouldn’t have suggested adding an entry for the upper ontology IRI to the catalog if the `odk:validate` command didn’t use the catalog. ;)

PR #1177, opened 2025-02-06
Move the pattern_clean rule out of the PAT=true block.

### LINKED ISSUE #1176: `make clean PAT=false` yields an error
Trying to run `make clean PAT=false` yields the following error:

```sh
$ odkrun make clean PAT=false
make[1]: *** No rule to make target 'pattern_clean'.  Stop.
make[1]: Leaving directory '/work/src/ontology'
make: *** [Makefile:1064: clean] Error 2
```

This is a mistake introduced when we switched to using Make conditionals (#1168): The `pattern_clean` target is defined _within_ the `ifeq ($(PAT),true)` block:

```Make
ifeq ($(PAT),true)

[...]
pattern_clean:
    rm -f $(PATTERN_CLEAN_FILES)

[...]

endif
```

so the entire rule does not exist under `PAT=false`, even though it is a prerequisite of the general `clean` rule.

This is not critical but very annoying, because while there is no reason to call `make clean PAT=false` in itself, it is common to explicitly call the `clean` target as part of a larger command, as in:

```sh
$ odkrun make clean uberon.owl PAT=false
```

to rebuild the ontology out of a clean state.

The fix is simply to move the `pattern_clean` rule out of the conditional block.

gouttegd:
This PR makes sure that the `pattern_clean` rule is always defined, even when PAT is set to false. This is because that rule is explicitly called by the general `clean` rule.

Of note, under PAT=false, `pattern_clean` can be called but it will not clean _all_ the pattern-derived files, because the variables defining the list of files to clean are themselves defined within the PAT block, and moving them out of the block would be too much of a hassle for little benefits.

While we are at it, this commit also removes an obsolete comment about the prepending of `./` to prevent deleting files outside of the repository, and clears some superfluous whitespaces.

closes #1176

### DIFF
--- a/template/src/ontology/Makefile.jinja2
+++ b/template/src/ontology/Makefile.jinja2
@@ -804,21 +804,19 @@ endif
 # DOSDP Templates/Patterns
 # ----------------------------------------
 
-ifeq ($(PAT),true)
-
 ALL_PATTERN_FILES=$(wildcard $(PATTERNDIR)/dosdp-patterns/*.yaml)
 ALL_PATTERN_NAMES=$(strip $(patsubst %.yaml,%, $(notdir $(wildcard $(PATTERNDIR)/dosdp-patterns/*.yaml))))
 
 PATTERN_CLEAN_FILES=../patterns/all_pattern_terms.txt \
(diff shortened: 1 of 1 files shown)


matentzn (approved):
awesome, thanks!

PR #1178, opened 2025-02-15
Check the ID range file when running the tests.

### LINKED ISSUE #211: travis checks should ensure that the idranges file is valid
Currently idranges.owl is never parsed by machine, it is informational for humans to adjust their Protege settings. We should have more support for this, and it is likely Protege will include support in the future (thanks @matthewhorridge!)

Basic validation: Should be valid OWL, and id ranges should not overlap. 

Note that GO's idranges is invalid right now as the ranges include the prefix - it has never been parsed!

gouttegd:
This PR makes `validate_idranges` a dependency of the `test` target, so the ID range file is systematically checked for errors as part of the standard test suite.

closes #211

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,7 +23,7 @@ For more detailed changes see:
 
 ## Makefile workflows
 
-- TBD
+- The `$(ont)-idranges.owl` file is now checked for validity as part of the normal test suite [#211](https://github.com/INCATools/ontology-development-kit/issues/211).
 
 ## Runner and Infrastructure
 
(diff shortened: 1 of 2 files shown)


matentzn (approved):
I am a bit worried many ontologies have malformed id ranges.. will you help me a bit when the crying users are trickling in?

gouttegd:
> I am a bit worried many ontologies have malformed id ranges

I am hoping that in the past two years, the support for ID range files in Protégé 5.6 has given enough incentive to ontology maintainers to fix their files (since Protégé cannot use an invalid ID range file).

But yeah, we know that some ontologies (e.g. GO) have kept an invalid ID range file for years, and it would not be surprising that some of them never got to fix it.

> will you help me a bit when the crying users are trickling in?

Sure.

PR #1180, opened 2025-02-19
Fix the update-repo system

### LINKED ISSUE #989: `update_repo`: Run 2x to alleviate errors
## Overview
I'm not sure why, but `update_repo` currently produces errors the first time you run it (after any changes): https://github.com/monarch-initiative/mondo-ingest/issues/308

Until that gets fixed, the current practice is that users must run `update_repo` a second time. I propose that rather than requiring users to know that they need to do this and remember to do it, we automatically run `update_repo` 2x in a row. See: [ignoring errors in make](https://www.gnu.org/software/make/manual/html_node/Errors.html#:~:text=To%20ignore%20errors%20in%20a,to%20the%20shell%20for%20execution.)

gouttegd:
This PR implements the idea outlined in [this comment](https://github.com/INCATools/ontology-development-kit/issues/989#issuecomment-2661573527) to avoid having to run the `update_repo` script twice.

Basically, instead of having the update code in a local script installed in the repo under `src/scripts/update_repo.sh`, we put all the update logic directly in the `odk.py` script. That script accepts a new command `update`. If executed from within the `src/ontology` directory of an ODK repository, that command will read the `*-odk.yaml` configuration file and re-instantiate the templates directly into the repository, without overwriting most files if they already exist – only a handful of files (most importantly the `src/ontology/Makefile` file) are forcibly overwritten.

The command can be invoked as `sh run.sh odk.py update`, or (preferably) as `sh run.sh update_repo`. For backwards compatibility, `sh run.sh make update_repo` is still supported, but is better avoided.

As said in the aforementioned comment, this approach has (at least) two benefits, which both stems from the fact that the updating code is now entirely within the Docker image rather than in a local file in the repository:

* it is always the latest version of the code that is executed;
* the updating code is under no risk of being overwritten while it is executed.

Another benefit is the logic to update the code is now at the same place as the logic that seeds an initial repository (in the `odk.py` script), instead of being in the separate `update_repo.sh.jinja2` template – this should facilitate the maintenance.

closes #989


odk/odk.py
```diff
@@ -980,6 +1073,63 @@ def dump_schema(class_name):
...
+    # because they are new files that didn't exist in the
+    # templates of the previous version of the ODK). But a
+    # handful of files are not reinstalled even if they are
+    # missing (e.g. DOSDP example files) or on the contrary
+    # always reinstalled to overwrite any local changes (e.g.
+    # the main Makefile). We declare the corresponding policies.
+    policies = [
+            ('CODE_OF_CONDUCT.md', NEVER),
+            ('CONTRIBUTING.md', NEVER),
+            ('issue_template.md', NEVER),
```
matentzn, line 1105: Should really be a structured template in the `.github` folder. I am digressing.

odk/odk.py
```diff
@@ -980,6 +1073,63 @@ def dump_schema(class_name):
...
+    # When updating, for most files, we only install them if
+    # they do not already exist in the repository (typically
+    # because they are new files that didn't exist in the
+    # templates of the previous version of the ODK). But a
+    # handful of files are not reinstalled even if they are
+    # missing (e.g. DOSDP example files) or on the contrary
+    # always reinstalled to overwrite any local changes (e.g.
+    # the main Makefile). We declare the corresponding policies.
+    policies = [
+            ('CODE_OF_CONDUCT.md', NEVER),
```
matentzn, line 1103: For the ones that are set to NEVER, will the code be robust to the possibility of the user deleting the file altogether?

matentzn (approved):
This _looooooks_ good - from the general approach perspective :P Haha but my confidence with the python code is rather low. I would say, if you tested it, lets merge and once we get into ODK 1.6 release mode, this funcition will anyways undergo massive testing.

odk/odk.py
gouttegd: Yes, with that PR we have the infrastructure that could allow us to give users a way to precisely control what they want to overwrite or not during an update.

But I think this would warrant more discussion and can be postponed to a future feature.

odk/odk.py
gouttegd: Yes. That's the point. If the policy is NEVER, well, we _never_ install the file, without even taking the time to look whether it already exists or not.

PR #388, opened 2021-02-02
ODK 1.2.26

matentzn:
- New versions:
  - fastobo validator to new version ([pull](https://github.com/INCATools/ontology-development-kit/pull/379)
  - ROBOT 1.8.1 (lots of new changes, see: [Changelog](https://github.com/ontodev/robot/blob/master/CHANGELOG.md))
  - DOSDPTOOLS 0.16 (lots of speed-up for bulk pattern generation)
- New features:
  - new python dependencies ([cogs](https://github.com/ontodev/cogs), a tool to directly manage tsv files in your repo on Google sheets)
  - stable serialisation order for JSON files using jq's walk function. -> this decreases the size of the diff for git!
  - Some improvements to logging when seeding a new repo, to make it easier to find errors
  - A new method to validate the id-ranges file can be invoked using `sh run.sh make validate_idranges` (after update to latest ODK repo)
  - modules are now annotated with a dc:source annotation to their original ontology (version) - FIX THIS
- New configuration options:
  - module_type ([example slme](tests/test-module-star.yaml), [example minimal](tests/test-module-minimal.yaml), [example mireot](tests/test-module-star.yaml)). Direct support for MIREOT and a new module type, minimal.
  - To encourage stable versions and releases, ODK, by default, merges imports into the release files. Previously, we continued to release the imports as well - which we do not recommend anymore. If you still wish to release your imports as usual, you can set a flag `release_imports` in the `import_group` section of your makefile (see [example](tests/test-release.yaml)).
  - the same as the above applies for reports (see [example](tests/test-robot-report.yaml))
  - The custom sparql checks, and the custom sparql exports, are now directly configurable
    - `custom_sparql_checks` : Chose which additional sparql checks you want to run. The related sparql query must be named CHECKNAME-violation.sparql, and be placed in the src/sparql directory (see [example](tests/test-robot-report.yaml))
    - `custom_sparql_exports` : Chose which additional sparql checks you want to run. The related sparql query must be named CHECKNAME-violation.sparql, and be placed in the src/sparql directory (see [example](tests/test-robot-report.yaml))
    - `custom_sparql_exports` : Chose which custom reports to generate. The related sparql query must be named CHECKNAME.sparql, and be placed in the src/sparql directory (see [example](tests/test-robot-report.yaml))
  - `git_main_branch` : The `main` branch for your repo, default `main`, or (now discouraged, previously) `master`.
  - `ci`: continuous integration defaults; currently available: `travis`, `github_actions`
[...]

PR #1182, opened 2025-02-19
Remove unused mireot test.

### LINKED ISSUE #970: Remove unused mireot test
I was wondering how to extract a module using MIREOT with ODK. I see a test file https://github.com/INCATools/ontology-development-kit/blob/master/tests/test-module-mireot.yaml with module_type: mireot but the Makefile.jinja2 template doesn't seem to support it: https://github.com/INCATools/ontology-development-kit/blob/master/template/src/ontology/Makefile.jinja2#L418. I suggest removing the test.

gouttegd:
It is confusing to have a test case for a `mireot` module_type while `mireot` is NOT actually a supported module type.

Remove the test for now. We can always add it back later if/when we actually implement a `mireot` module type.

closes #970


matentzn (approved):
Haha look at this: https://github.com/INCATools/ontology-development-kit/pull/388

There was an annoucement that we implemented mireot support, but somehow.. we didnt?? I bet we lost a branch somewhere in these crazy times (before you brought order here).

PR #1183, opened 2025-02-19
Do not commit release artefacts by default

### LINKED ISSUE #1144: Seeding process should not default to commit release artefacts
… or there should at least be an option to disable this.

When a new ODK-managed repository is seeded, all release artefacts from the “initial release“ in the top-level directory are forcefully committed to the repository.

So, even if a project takes the (sensible) decision that they do no want to commit their release artefacts, they will still end up with a "first release" commit in their history in which all the release files are in fact committed.

This behaviour should be configurable, and I’d argue that the default should be _not_ to commit the release artefacts.

gouttegd:
This PR updates the seeding process so that, by default, release artefacts are _not_ committed to the repository.

It adds a project-level option named `commit_release_artefacts` that can be set to `True` to restore the previous behaviour of committing the artefacts.

closes #1144


template/src/ontology/Makefile.jinja2
```diff
@@ -426,14 +426,12 @@ prepare_release: all_odk
...
 
 .PHONY: prepare_initial_release
-prepare_initial_release: all_assets
-	rsync -R $(RELEASE_ASSETS) $(RELEASEDIR) &&\{% if project.sssom_mappingset_group is defined %}{% if project.sssom_mappingset_group.released_products is defined %}
-	mkdir -p $(RELEASEDIR)/mappings && cp -rf $(RELEASED_MAPPING_FILES) $(RELEASEDIR)/mappings &&\{% endif %}{% endif %}{% if project.use_dosdps %}
-	mkdir -p $(RELEASEDIR)/patterns && cp -rf $(PATTERN_RELEASE_FILES) $(RELEASEDIR)/patterns &&\{% endif %}
-	rm -f $(patsubst %, ./%, $(CLEANFILES)) &&\
-	cd $(RELEASEDIR) && git add $(RELEASE_ASSETS)
+prepare_initial_release: prepare_release
+	git add $(foreach asset,$(RELEASE_ASSETS),$(RELEASEDIR)/$(asset))
```
matentzn, line 433: Second thought, why not drop the `prepare_initial_release` entirely in favour of a line the odk.py? This is also a noisy goal, once used, never used again, so belongs conceptually in odk.py?

matentzn (commented):
Some initial thoughts on this. In a nutshell:

1. obliterate `prepare_initial_release` entirely
2. make `commit_release_artefacts` a parameter to odk.py rather than a config option

See comments for details

template/.gitignore.jinja2
```diff
@@ -10,6 +10,18 @@ bin/
 .template.db
 
 .github/token.txt
+{%- if not project.commit_release_artefacts %}
```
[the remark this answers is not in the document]
gouttegd: I need to check something about Git's behaviour regarding ignored files, but I _think_ I have an ideal solution here: ignore those files unconditionally, but if the user opts for committing the release artefacts, add them anyway (`git add -f`).

This would both prevent _inadvertently_ committing those files, while allowing to add them once at seeding time.

Key trick is that I _think_ that once a file has been `git add`ed once, Git will track their changes and allow them to be committed even if they are in `.gitignore` (that's what I need to check, but can't do that right now).

odk/odk.py
```diff
@@ -687,6 +687,9 @@ class OntologyProject(JsonSchemaMixin):
     release_artefacts : List[str] = field(default_factory=lambda: ['full', 'base'])
     """A list of release artefacts you wish to be exported. Supported: base, full, baselite, simple, non-classified, 
     simple-non-classified, basic."""
+
+    commit_release_artefacts : bool = False
```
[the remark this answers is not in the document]
gouttegd: It makes more sense indeed.

template/src/ontology/Makefile.jinja2
gouttegd: My concern with that is that it requires duplicating a logic that currently only exists in the Makefile, but maybe it's not so much of a big deal, I'll have another look.

odk/odk.py
```diff
@@ -1213,10 +1214,18 @@ def seed(config, clean, outdir, templatedir, dependencies, title, user, source,
...
+        runcmd("cd {dir} && git init -b {branch} && git add {files} && git commit -m 'initial commit'".
                format(dir=outdir,
+                      branch=project.git_main_branch,
                       files=" ".join([t.replace(outdir, ".", 1) for t in tgts])))
-        runcmd("cd {dir}/src/ontology && make && git commit -m 'initial commit' -a && git branch -M {branch} && make prepare_initial_release && git commit -m 'first release'".format(dir=outdir, branch=project.git_main_branch))
+        runcmd("cd {dir}/src/ontology && make all_assets && cp $(make show_release_assets) ../../"
+               .format(dir=outdir))
+        if commit_artefacts:
+            runcmd("cd {dir}/src/ontology "
+                    "&& for asset in $(make show_release_assets) ; do git add -f ../../$asset ; done"
```
matentzn, line 1225: smart using git add -f and saving a condititional in .gitignore!

matentzn (approved):
looks great!

PR #1184, opened 2025-02-20
Fix the odk-info script.

gouttegd:
Now that Ammonite is no longer bundled in the ODKLite image, the `odk-info.sh` script needs to check whether Ammonite is present before attempting to display its version, otherwise it would fail on ODKLite.

While we are at it, we update the script to also display the version of Scala-CLI and SSSOM-CLI.

### DIFF
--- a/scripts/odk-info.sh
+++ b/scripts/odk-info.sh
@@ -23,7 +23,7 @@ case $1 in
     ;;
 
 *)
-    echo "Usage: odkinfo [--tools|--all-python|--all]"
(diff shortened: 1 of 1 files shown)


matentzn (commented):
Thanks, I am a bit stretched this week hence I am a bit delayed in reviewing stuff, much appreciated!

gouttegd:
No problem. I didn’t see the point of asking you to review a trivial change like this one (I’d much rather have you review decidedly not trivial changes like #1180 or #1183 ;) ).

PR #1186, opened 2025-02-20
Use `robot export` to generate subset TSV files.

### LINKED ISSUE #798: Subset TSV makes a lot of noise in diff?
See https://github.com/oborel/obo-relations/pull/492/files

gouttegd:
For every subset, we produce a TSV file that is supposed to list the classes in the subset. This is done by a SPARQL query.

However, this can be done much more efficiently with a `robot export` command, which is what we do here. On some ontologies that I have tested (Uberon, CL), this is about 2 to 3 times faster.

This also automatically excludes blank nodes from the listing, contrary to the original SPARQL query.

closes #798


matentzn (approved):
Looks good!

PR #1187, opened 2025-02-20
Put the current ROBOT profile in the templates directory.

### LINKED ISSUE #999: Default profile.txt (ROBOT report config) not synced with ROBOT repo
See #998 

profile.txt is currently hardcoded in `/template/_dynamic_files.jinja2`, so we need to manually update this with every major release

gouttegd:
For now, when building the image, we download the ROBOT report profile for the current version of ROBOT (the version used in the ODK) and put it in `/tools/robot_report_profile.txt`. We do that so that we can check, as part of the standard workflow, whether the custom profile that the project may be using is not lacking any new tests that are available in the current version of ROBOT (see #998).

Here, we simply move the location of the profile file within the image, from `/tools/robot_report_profile.txt` to `/tools/templates/src/ontology/profile.txt`.

This does not change how the check mentioned above works (we just update the path involved in that check), but this will have the benefit of automatically installing the ROBOT profile when seeding a repository, since the file is now within the template directory used by the seeding system.

Therefore, we can remove the `src/ontology/profile.txt` block from the dynamic template pack (which needed to be manually updated by us whenever we updated ROBOT – something I don't think we have really been doing).

But we do not want to _always_ install this `profile.txt` file in a seeded repository – we only need to do that when the use of a custom profile is enabled in the configuration (that was the reason why the profile was part of the dynamic template pack). To avoid installing the file when a custom profile is _not_ enabled, we use the recently introduced per-template policies system: we set the policy for the `src/ontology/profile.txt` template to NEVER if a custom profile is not enabled.

closes #999

### DIFF
--- a/docker/odklite/Dockerfile
+++ b/docker/odklite/Dockerfile
@@ -46,8 +46,9 @@ RUN wget -nv $ROBOT_JAR \
     wget -nv https://raw.githubusercontent.com/ontodev/robot/v$ROBOT_VERSION/bin/robot \
         -O /tools/robot && \
     chmod 755 /tools/robot &&\
+    mkdir -p /tools/templates/src/ontology &&\
     wget -nv https://raw.githubusercontent.com/ontodev/robot/v$ROBOT_VERSION/robot-core/src/main/resources/report_profile.txt \
-        -O /tools/robot_report_profile.txt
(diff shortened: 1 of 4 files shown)


matentzn (approved):
Very nice, thanks!

PR #1190, opened 2025-02-23
Error out cleanly if the config file cannot be parsed.

### LINKED ISSUE #1167: Add polite odk.py error response for non-parseable yaml.
I was trying to get ODK to run the seed-via-docker.sh command to create an initial configuration (via "> sh seed-via-docker.sh -c -C foodon-odk.yaml") and didn't notice one line of my foodon-odk.yaml file was syntactically invalid.  ODK errored out with the following cryptic message:

```
File "/tools/odk.py", line 829, in load_config
    project = from_dict(data_class=OntologyProject, data=obj)
                                                         ^^^
UnboundLocalError: cannot access local variable 'obj' where it is not associated with a value
```
I fixed the syntax error, and the above went away.  Can ODK be revised to indicate politely when the input yaml file can't be parsed?

gouttegd:
Do not catch any exception thrown by the YAML parser in the `load_config` method. Let the exception be propagated up to the level of one of the Click commands and process it there.

"Processing" the exception includes:

(1) making a clean error message (including details about what the problem is, if we got them from the YAML parser);
(2) raising a Click exception, which will in turn ensure that (a) the error message is printed out, (b) the process is terminated, (c) the process returns a non-zero exit code.

closes #1167

### DIFF
--- a/odk/odk.py
+++ b/odk/odk.py
@@ -845,10 +845,7 @@ class Generator(object):
                 h.update(stream.read().encode())
                 config_hash = h.hexdigest()
                 stream.seek(0)
-                try:
-                    obj = yaml.load(stream, Loader=yaml.FullLoader)
-                except yaml.YAMLError as exc:
-                    print(exc)
+                obj = yaml.load(stream, Loader=yaml.FullLoader)
             project = from_dict(data_class=OntologyProject, data=obj)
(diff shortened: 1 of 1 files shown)


matentzn (approved):
Looks good to me!

Issue #650, opened 2022-08-09
Document development principle


matentzn:
1. It should be very easy to do the standard stuff, and make it possible for knowledgeable people to be able to customise.

Issue #631, opened 2022-06-30
Document all tools we should review for updates _every time_ there is a new release


matentzn:
- oaklib
- sssom-py
- rdftab
- robot (ok no need to check)
- dosdp
- dosdp-tools

PR #1193, opened 2025-03-02
Cleaning up obsolete docs, more CONTRIBUTING docs

gouttegd:
This PR:

* removes a bunch of old documentation pages that now (1) merely point to the OBOOK (2) are no longer published anywhere, and are consequently useless;
* reorganises a bit the CONTRIBUTING.md document and adds to it a section about development principles (#650) and about tools that need to be checked for updates (#631).


CONTRIBUTING.md
```diff
@@ -3,37 +3,140 @@
...
+Jinja2 templates should be fairly easy to grok for anyone familiar with
+templating systems. We feed the template engine with a project object
+that is passed in by the user (see below).
+
+Logic in the templates should be kept to a minimum (though the
+aforementioned `Makefile.jinja2` template is a great offender of this
+principle).
 
-Logic in the templates should be non-existent.
+### Some notes on templating and logic
```
gouttegd, line 114: @matentzn This entire section needs to be reconsidered. It is self-contradictory: we cannot simultaneously say that the logic in the Python odk code must be kept to the absolute minimum, _and_ that the templates must be as simple as possible. The logic must be _somewhere_, if it is not in the Python code then it must be in the templates, or the other way around.

I personally think we should in fact put _more_ logic into the Python code, in order to _really_ keep the templates as simple as possible.

CONTRIBUTING.md
```diff
@@ -3,37 +3,140 @@
...
+project does not have the luxury of a full-time ontology pipeline
+engineer always available to fix snafus and keep things running.
 
-See [Dockerfile](Dockerfile) for details on how to obtain these
+From this aim, we derive the following guidelines that developers should
+keep in mind whenever working on the ODK (especially when adding new
+features):
 
-## How it works
+(1) No feature should require from the user to install anything on their
```
matentzn, line 20: technically (just being a bit extreme now) you will need `bash` or similar as well :P

CONTRIBUTING.md
matentzn: This is some old text, I agree with your changes and suggestions!

matentzn (approved):
This is amazing! A huuge improvement!

PR #1198, opened 2025-03-13
spelling in project-schema.md

turbomam:


gouttegd (approved):

matentzn:
This has to be changed in

https://github.com/INCATools/ontology-development-kit/blob/master/odk/odk.py

as 

https://github.com/INCATools/ontology-development-kit/blob/master/docs/project-schema.md

is generated via

https://github.com/turbomam/ontology-development-kit/blob/master/schema/project-schema.json

which in turn is generated via

https://github.com/INCATools/ontology-development-kit/blob/f3e0745a960157120a236bd1911ef698a6c3eecc/Makefile#L70

Sorry this is.. Horrendous. And not in our SOP either..

matentzn:
The best choice!

PR #1206, opened 2025-03-30
Make the stripping of annotation properties optional

### LINKED ISSUE #1093: Revise stripping of annotation properties in imports
The ODK 1.5 forcefully introduced a new feature in which almost all annotation properties, except `rdfs:label` and `IAO:0000115`, are forcibly removed from the import modules.

I say “forcefully” because, unless I missed something, that feature is not deactivatable. The only thing a user can do is amend the list of annotation properties that are _not_ removed, to preserve more properties than just `rdfs:label` and `IAO:0000115`.

This was intended to reduce the size of the import modules, and based on the premise that annotation properties are never “significant” (especially from a reasoner’s point of view) anyway.

However that premise is blatantly false, as annotation properties loaded with a “logical meaning“ are certainly not unheard of. (One can argue about whether this is a good idea or not, but that’d be beside the point.)

In particular, it is a relatively common practice to use annotation properties as “shortcuts” to be replaced by more complex logical axioms. That idea has been around for example with OWLTools’ macros for more than a decade ([Mungall, Ruttenberg, and Osumi-Sutherland, 2010](https://www.nature.com/articles/npre.2010.5292.1.pdf)), and while the use of OWLTools is being phased out, ROBOT’s [expand command](https://robot.obolibrary.org/expand) merely continues on the same principle, the only difference with OWLTools being the language used to express the expanded form of the “macro” (OWL Manchester for `owltools --expand-macros`, SPARQL for `robot expand`).

Examples of annotation properties with a “logical meaning” include:

* [mutually spatially disjoint with](http://purl.obolibrary.org/obo/RO_0002171)
* [present in taxon](http://purl.obolibrary.org/obo/RO_0002175)
* [never in taxon](http://purl.obolibrary.org/obo/RO_0002161)
* [has no connections with](http://purl.obolibrary.org/obo/RO_0002475)

All that to say that removing almost all annotation properties from imports is _not_ something that should be done lightly because it can definitely have drastic consequences for the ontologies that use the ODK to manage their imports. As an example, the entire check for violation of taxon constraints in Uberon has been completely neutralised by this “feature” (https://github.com/obophenotype/uberon/pull/3336).

So what should the ODK do now?

**A.** Nothing. It’s up to the users to know that they depend on some specific annotations in the ontologies that they import, and to declare said annotations in the `import_group.annotation_properties` option of their ODK configuration file so that they are not removed.

Why not, but then this needs to be thoroughly documented somewhere. As far as I know, the only “documentation” for now is this line in the description of the [ODK Project Configuration Schema](https://incatools.github.io/ontology-development-kit/project-schema/):

> **annotation_properties** *(list)*(: Define which annotation properties to pull in. Default: [`rdfs:label`, `IAO:0000115`]

I believe this is woefully insufficient. At no point do we actually explain that the ODK does not “pull” all annotation properties from imports, only those that are listed here. Combined with the fact that the feature was introduced without any announcement in the release notes, users have had no chance at all to know that from now on they should declare the annotation properties they need.

**B.** Revert the feature entirely.

That’d probably be too much. There is no doubt the feature _is_ useful.

**C.** Leave the feature as it is, but make it optional.

Simply add an option to enable it (if disabled by default) or disable it (if enabled by default).

**D.** Leave the feature as it is, but preserve more annotation properties by default than just `rdfs:label` and `IAO:0000115`.

At the very least, we should preserve:

* [expand assertion to](http://purl.obolibrary.org/obo/IAO_0000425) (used by OWLTools’ `--expand-macros` command);
* [expand expression to](http://purl.obolibrary.org/obo/IAO_0000424) (likewise);
* [defined by construct](http://purl.obolibrary.org/obo/OMO_0002000) (used by ROBOT’s `expand` command);
* all annotation properties under RO’s [logical macro assertion](http://purl.obolibrary.org/obo/RO_0002416).

Also, I believe that preserving a given annotation property should automatically preserve all its _subproperties_. That is, users should not have to declare that they want to preserve both [present in taxon](http://purl.obolibrary.org/obo/RO_0002175) and [always present in taxon](http://purl.obolibrary.org/obo/RO_0002504) – just declaring the former should automatically preserve the latter.

gouttegd:
This PR adds a new option to the ImportGroup section, `strip_annotation_properties`. It allows to disable the stripping of annotation properties from the import modules (introduced in ODK 1.5). The option is enabled by default, preserving the ODK 1.5 behaviour.

It also makes sure that, when annotation stripping is enabled, it is performed on individual SLME modules as well, instead of only on the merged module and individual “filter” and ”“minimal” modules.

Lastly, it adds OMO:0002000 (the property used to define macros expanded by ROBOT’s `expand` command) to the list of properties preserved by default.

closes #1093

### DIFF
--- a/odk/odk.py
+++ b/odk/odk.py
@@ -359,6 +359,9 @@ class ImportGroup(ProductGroup):
     
     annotation_properties : List[str] = field(default_factory=lambda: ['rdfs:label', 'IAO:0000115'])
     """Define which annotation properties to pull in."""
+
+    strip_annotation_properties: bool = True
+    """If set to true, strip away annotation properties from imports, apart from explicitly imported properties and properties listed in annotation_properties."""
     
     directory : Directory = "imports/"
     """directory where imports are extracted into to"""
--- a/odk/odk.py
+++ b/odk/odk.py
@@ -147,7 +147,7 @@ class ImportProduct(Product):
     module_type_slme : str = "BOT"
     """SLME module type. Supported: BOT, TOP, STAR"""
     
-    annotation_properties : List[str] = field(default_factory=lambda: ['rdfs:label', 'IAO:0000115'])
+    annotation_properties : List[str] = field(default_factory=lambda: ['rdfs:label', 'IAO:0000115', 'OMO:0002000'])
     """Define which annotation properties to pull in."""
     
     slme_individuals : str = "include"
@@ -357,7 +357,7 @@ class ImportGroup(ProductGroup):
     export_obo: bool = False
     """If set to true, modules will not only be created in OWL, but also OBO format"""
     
-    annotation_properties : List[str] = field(default_factory=lambda: ['rdfs:label', 'IAO:0000115'])
+    annotation_properties : List[str] = field(default_factory=lambda: ['rdfs:label', 'IAO:0000115', 'OMO:0002000'])
     """Define which annotation properties to pull in."""
 
     strip_annotation_properties: bool = True
--- a/template/src/ontology/Makefile.jinja2
+++ b/template/src/ontology/Makefile.jinja2
@@ -506,8 +506,10 @@ T_IMPORTSEED = --term-file $(IMPORTSEED)
 {% endif -%}
 
 {% if project.import_group is defined -%}
+{% if project.import_group.strip_annotation_properties -%}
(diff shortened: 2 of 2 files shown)


matentzn (approved):
Looks great, thank you!

PR #1208, opened 2025-04-12
Improvements to the SSSOM workflows

### LINKED ISSUE #106: Add a standard pipeline step to export xrefs
Standard release layout (proposed)

 * ont
    * xrefs
       * ont-xref-ext1.{obo,owl,tsv}
       * ont-xref-ext2.{obo,owl,tsv}
       * ...

Note that sometimes we also release skos mappings (with semi-loose broad/narrow/exact semantics), and sometimes equivalence (owl)

See http://uberon.org downloads page for an example of how we do this for ontologies

@rctauber @jamesaoverton does it make sense to have this here or in robot?

gouttegd:
This PR primarily adds a new value for the `maintenance` field of SSSOM mapping sets declared in the `sssom_mappingset_group`: `merged`.

Given the following configuration:

```yaml
sssom_mappingset_group:
  products:
    - id: my-set-a
    - id: my-set-b
    - id: my-merged-set
      maintenance: merged
      source_mappings:
        - my-set-a
        - my-set-b
```

The `my-merged-set.sssom.tsv` set will be created by merging the `my-set-a.sssom.tsv` and `my-set-b.sssom.tsv` sets.

If `source_mappings` is not set for a `merged` set, the default behaviour is to use all the other sets as source, so the example above is actually equivalent to:

```yaml
sssom_mappingset_group:
  products:
    - id: my-set-a
    - id: my-set-b
    - id: my-merged-set
      maintenance: merged
```

(This is as discussed in #106.)

This PR also brings a few other changes to the SSSOM workflows:

A. Another `maintenance` value, `custom`, to explicitly declare that a set is to be built using a custom rule in the custom Makefile (similarly to the `module_type: custom` for import modules).

B. The possibility to use the `xref-extract` command of the SSSOM ROBOT plugin, rather than `sssom-py parse`, to extract mappings from an ontology. This is done with a new group-level option, `mapping_extractor`. The default value for that option is `sssom-py`, which preserves the existing behaviour of using `sssom-py parse`. If set to `robot`, mappings would be extracted using `robot sssom:xref-extract`.

C. New helper rules to force mapping sets to be re-serialized, similarly to the `normalize_src` rule (`normalize-sssom-X` to re-serialize mapping set _X_, `normalize_mappings` to re-serialize all sets).

closes #106

### DIFF
--- a/odk/odk.py
+++ b/odk/odk.py
@@ -518,6 +518,9 @@ class SSSOMMappingSetGroup(JsonSchemaMixin):
 
     release_mappings : bool = False
     """If set to True, mappings are copied to the release directory."""
+
+    mapping_extractor : str = "sssom-py"
+    """The tool to use to extract mappings from an ontology ('sssom-py' or 'robot')."""
     
     products : Optional[List[SSSOMMappingSetProduct]] = None
 
--- a/odk/odk.py
+++ b/odk/odk.py
@@ -531,15 +531,19 @@ class SSSOMMappingSetGroup(JsonSchemaMixin):
         if len(released_products) > 0:
             self.released_products = released_products
 
-        for product in [p for p in self.products if p.maintenance == "merged"]:
-            if product.source_mappings is None:
-                # Merge all other non-merge sets to make this one
-                product.source_mappings = [p.id for p in self.products if p.maintenance != "merged"]
-            else:
-                # Check that all listed source sets exist
-                for source in product.source_mappings:
-                    if source not in [p.id for p in self.products]:
-                        raise Exception(f"Unknown source mapping set '{source}'")
+        for product in self.products:
+            if product.maintenance == "merged":
+                if product.source_mappings is None:
+                    # Merge all other non-merge sets to make this one
+                    product.source_mappings = [p.id for p in self.products if p.maintenance != "merged"]
+                else:
+                    # Check that all listed source sets exist
+                    for source in product.source_mappings:
+                        if source not in [p.id for p in self.products]:
+                            raise Exception(f"Unknown source mapping set '{source}'")
+            elif product.maintenance == "extract":
+                if product.source_file is None:
+                    product.source_file = "$(EDIT_PREPROCESSED)"
 
 @dataclass_json
 @dataclass
--- a/odk/odk.py
+++ b/odk/odk.py
@@ -194,6 +194,9 @@ class SSSOMMappingSetProduct(Product):
     release_mappings: bool = False
     """If set to True, this mapping set is treated as an artifact to be released."""
 
+    source_mappings: Optional[List[str]] = None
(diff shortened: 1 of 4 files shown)


odk/odk.py
```diff
@@ -528,6 +534,20 @@ def fill_missing(self):
...
+                    # Merge all other non-merge sets to make this one
+                    product.source_mappings = [p.id for p in self.products if p.maintenance != "merged"]
+                else:
+                    # Check that all listed source sets exist
+                    for source in product.source_mappings:
+                        if source not in [p.id for p in self.products]:
+                            raise Exception(f"Unknown source mapping set '{source}'")
+            elif product.maintenance == "extract":
+                if product.source_file is None:
+                    product.source_file = "$(EDIT_PREPROCESSED)"
```
matentzn, line 549: I am assuming you thought this through - I recently struggled hard with a pipeline that provided a merged sssom mapping set which was partially extracted from upstream, but partially generated by the pipeline itself.. So, something like: build a bridge, then extract a mapping set from that, than merge that into a mapping set you publish. This caused some circularity. I don't think this is the case here, just wanted to share my experience.

odk/odk.py
```diff
@@ -515,6 +518,9 @@ class SSSOMMappingSetGroup(JsonSchemaMixin):
 
     release_mappings : bool = False
     """If set to True, mappings are copied to the release directory."""
+
+    mapping_extractor : str = "sssom-py"
+    """The tool to use to extract mappings from an ontology ('sssom-py' or 'robot')."""
```
matentzn, line 523: Maybe to be more future proof this could be sssom-py and sssom-java (in case someone implements a stellar mapping extraction method in robot itself :P). Ok fantasising now, but.. suggestion.

matentzn (approved):
Looks fantastic. Two optional comments!

odk/odk.py
gouttegd: I am not sure I understand your concern, and how it applies here.

In any case, the only change here is that the logic to handle the absence of a `source_file` key has been moved from within the Makefile template to the `odk.py` script. This is so the template can simply says

```Jinja2
$(MAPPINGDIR)/{{ mapping.id }}.sssom.tsv: {{ mapping.source_file }}
```

instead of

```Jinja2
$(MAPPINGDIR)/{{ mapping.id }}.sssom.tsv: {% if mapping.source_file is not none %}{{ mapping.source_file }}{% else %}$(EDIT_PREPROCESSED){% endif %}
```

This is functionally identical to what we had before, only with a (slightly) more readable template.

odk/odk.py
gouttegd: > in case someone implements a stellar mapping extraction method in robot itself

I think that’s _highly_ unlikely :D but OK, I have no objection with that.

PR #1209, opened 2025-04-13
Split the dynamic templates pack into several packs.

### LINKED ISSUE #1192: Break down the `_dynamic` templates in several packs
The existing `_dynamic_files.jinja2` “pack of templates” contains a lot of completely unrelated templates, ranging from documentation files to GitHub workflows to ID range file to import placeholders, etc.

There is in fact no reason for having all those files in a single big pack. The code in `odk.py` is perfectly able to deal with an arbitrary number of `_dynamic*.jinja2` files, so I think we should break down the single pack into a set of smaller, “thematic” packs (say, `_dynamic_documentation.jinja2`, `_dynamic_workflows.jinja2`, etc). The end result would be the same, but smaller packs would be easier to maintain.

gouttegd:
Split the monolithic `_dynamic_files.jinja2` templates pack into a few “thematic” templates:

* `_dynamic_documentation.jinja2` for all documentation files;
* `_dynamic_workflows.jinja2` for all GitHub/GitLab/Travis/etc. workflows;
* `_dynamic_sparql.jinja2` for all SPARQL files.
* and the original `_dynamic_files.jinja2` for all the other various dynamic templates.

This will make it easier to edit and maintain those templates.

This does not change anything to the way the ODK works, except for one thing: the `.github/workflows/docs.yml` workflow is now created, **as it should**, if only the following two conditions are met: (1) `project.ci` contains `github_actions` and (2) `project.workflows` contains `docs`. Previously, that file was only created if, in addition to those conditions, `project.use_dosdp` was set to `True` (because of a misplaced Jinja conditional statement – precisely the kind of errors that should be easier to avoid with smaller packs).

closes #1192

### DIFF
--- a/template/_dynamic_documentation.jinja2
+++ b/template/_dynamic_documentation.jinja2
@@ -1,0 +1,865 @@
+{#-
+   Documentation templates pack
+-#}
+{% if project.documentation is not none %}
+^^^ docs/index.md
+# {{ project.id.upper() }} Ontology Documentation
+
+[//]: # "This file is meant to be edited by the ontology maintainer."
+
+Welcome to the {{ project.id.upper() }} documentation!
+
+You can find descriptions of the standard ontology engineering workflows [here](odk-workflows/index.md).
+^^^ docs/cite.md
+# How to cite {{ project.id.upper() }}
+^^^ docs/contributing.md
+# How to contribute to {{ project.id.upper() }}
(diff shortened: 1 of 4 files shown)


matentzn (approved):
Amazing refactoring. I love love the jinja comments on the conditionals, this will make it much more easy to find the conditional block. intendation of conditional blocks also a great idea for readability!

PR #1211, opened 2025-04-14
Always use context file if requested.

### LINKED ISSUE #816: Add option to apply context file to release products to canonicalise prefixes
Similar to what we can already do for templates: https://gitlab.c-path.org/c-pathontology/critical-path-ontology/-/blob/main/src/ontology/cpont-odk.yaml#L77

This will also allow us to circumvent issues like this: https://github.com/ontodev/robot/issues/1101

gouttegd:
If the user requested the generation and use of a “context file”, then we ensure that said context file is systematically used in all calls to ROBOT.

closes #816

### DIFF
--- a/template/src/ontology/Makefile.jinja2
+++ b/template/src/ontology/Makefile.jinja2
@@ -35,7 +35,10 @@ EDIT_FORMAT=                {{ project.edit_format|default('owl') }}
 SRC =                       $(ONT)-edit.$(EDIT_FORMAT)
 MAKE_FAST=                  $(MAKE) IMP=false PAT=false COMP=false MIR=false
 CATALOG=                    {{ project.catalog_file }}
-ROBOT=                      robot --catalog $(CATALOG)
+{% if project.use_context -%}
+CONTEXT_FILE =              config/context.json
+{% endif -%}
+ROBOT=                      robot --catalog $(CATALOG){% if project.use_context %} --add-prefixes $(CONTEXT_FILE){% endif %}
 REASONER=                   {{ project.reasoner }}
 RELEASEDIR=                 ../..
 DOCSDIR=                    ../../docs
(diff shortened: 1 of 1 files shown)


gouttegd:
Tested on Uberon. Seems to be working fine, but users of that feature should be careful about what they put (and _don’t put_) in their context file.

The behaviour when the context file is missing some prefixes seems… weird, if not downright wrong.

Uberon is currently using both some `http://purl.org/dc/terms/` properties (notably `http://purl.org/dc/terms/contributor`) and some `http://purl.org/dc/elements/1.1/` properties (like `http://purl.org/dc/elements/1.1/creator` and `http://purl.org/dc/elements/1.1/source`).

Without a context file, the resulting OWL file is what I would expect: it contains namespace declarations for both `http://purl.org/dc/terms/` and http://purl.org/dc/elements/1.1/`:

```
xmlns:dc="http://purl.org/dc/elements/1.1/"
xmlns:terms="http://purl.org/dc/terms/"
```

and the properties use the correspondingly declared prefixes:

```
<dc:source rdf:resource="http://dbpedia.org"/>
<terms:contributor rdf:resource="https://orcid.org/0000-0001-5889-4463"/>
```

But if you inject (with `--add-prefixes` as in this PR) the default ODK context file, which contains a prefix declaration for `http://purl.org/dc/terms/` (associated with the prefix name `dc:`) but _not_ for `http://purl.org/dc/elements/1.1/`, the result is… weird. The file only contains a declaration for `http://purl.org/dc/elements/1.1/` and _not_ for `http://purl.org/dc/terms/` (basically the opposite of what I would expect given the context file).

```
xmlns:dc="http://purl.org/dc/elements/1.1/"
```

Even worse, the `http://purl.org/dc/terms/contributor` properties are transformed into `http://purl.org/dc/elements/1.1/contributor` properties!

```
<dc:contributor rdf:resource="https://orcid.org/0000-0001-5889-4463"/>
```

I had to add to the context file a declaration for the `http://purl.org/dc/elements/1.1/` namespace, as follows:

```
"dce": "http://purl.org/dc/elements/1.1/"
```

to obtain something I consider correct:

```
xmlns:dc="http://purl.org/dc/terms/"
xmlns:dce="http://purl.org/dc/elements/1.1/"

[…]

<dce:source rdf:resource="http://dbpedia.org"/>
<dc:contributor rdf:resource="https://orcid.org/0000-0001-5889-4463"/>
```

**TL;DR** `robot --add-prefixes` behaves weirdly when the ontology uses namespaces that are _not_ in the provided context file, so users should always make sure their context file is 100% complete.

PR #1212, opened 2025-04-15
Misc fixes

gouttegd:
This PR fixes a handful of issues with the current pipelines:

* It disables the systematic printing of Scala-CLI’s version number, which takes **several seconds** every time.
* It restores the `OWLTOOLS` variable in the standard Makefile; though it is no longer used there, custom Makefiles might still be dependent on it (e.g. Uberon).
* It allows running the test suite (or any other pipeline dependent on the `test` target) with the pattern pipeline disabled (PAT=false).


matentzn (approved):
Thanks looks good

Issue #1207, opened 2025-04-05
Remove all uses of `make -B` in standard Makefile and documentation


matentzn:
I am preparing some workshop atm so no time to verify but I think there is a big bug in the dosdp pipeline in ODK:

```
sh run.sh make ../patterns/definitions.owl -B --dry-run
```

If you run this in any ontology that uses dosdp, you will see that the -B will cause the DOSDP generate goal:

```
$(DOSDP_OWL_FILES_DEFAULT): $(EDIT_PREPROCESSED) $(DOSDP_TSV_FILES_DEFAULT) $(ALL_PATTERN_FILES)
	if [ $(PAT) = true ] && [ "${DOSDP_PATTERN_NAMES_DEFAULT}" ]; then $(DOSDPT) generate --catalog=$(CATALOG) \
    --infile=$(PATTERNDIR)/data/default/ --template=$(PATTERNDIR)/dosdp-patterns --batch-patterns="$(DOSDP_PATTERN_NAMES_DEFAULT)" \
    --ontology=$< --obo-prefixes=true --restrict-axioms-to=logical --outfile=$(PATTERNDIR)/data/default; fi
```

To be re-evaluated as often as the number of patterns that exist; **without -B it works fine**.

gouttegd:
> sh run.sh make ../patterns/definitions.owl -B --dry-run

First, you **cannot** rely on `--dry-run` to check how many times a goal would be executed.

Second, that’s the **completely expected** consequence of using the `-B` option: Make will unconditionally execute the rule for all targets (so, all the files listed in `$(DOSDP_OWL_FILES_DEFAULT)`) even if the pre-requisites for all the targets after the first one have not changed since the rule has been executed for the first target -- because that’s precisely what the `-B` option is for (executing a rule without regard for the status of the pre-requisites).

Consider the following minimal example:

```Make
pattern1 pattern2 pattern3: tsv1 tsv2 tsv3
        touch pattern1 pattern2 pattern3

definition: pattern1 pattern2 pattern3
        touch definition

tsv%:
        touch $@
```

Let’s call `make definition` **without -B**:

```sh
touch tsv1
touch tsv2
touch tsv3
touch pattern1 pattern2 pattern3
touch definition
```

[...]

PR #1214, opened 2025-04-15
Refactor the components pipeline

### LINKED ISSUE #1213: SSSOM-derived components do not work
Trying to create a component from a SSSOM mapping set like this:

```yaml
components:
  products:
    - filename: test-with-sssom.owl
      use_mappings: true
      mappings:
        - modmini-onta.sssom.tsv
        - modmini-ontd.sssom.tsv
```

does not work:

```sh
$ odkrun make recreate-components
make --assume-new=tmp/stamp-component-test-with-sssom.owl \
	COMP=true IMP=false MIR=true PAT=true all_components
make[1]: Entering directory '/work/src/ontology'
sssom merge ../mappings/modmini-onta.sssom.tsv ../mappings/modmini-ontd.sssom.tsv --output tmp/test-with-sssom.owl-merged.sssom.tsv &&\
	sssom convert tmp/test-with-sssom.owl-merged.sssom.tsv  --output tmp/test-with-sssom.owl-converted.sssom.owl &&\
	robot --catalog catalog-v001.xml merge -i tmp/test-with-sssom.owl-converted.sssom.owl \
	annotate --ontology-iri http://purl.obolibrary.org/obo/sssomo/components/test-with-sssom.owl annotate -V http://purl.obolibrary.org/obo/sssomo/releases/2025-04-15/components/test-with-sssom.owl --annotation owl:versionInfo 2025-04-15 convert -f ofn --output components/test-with-sssom.owl.tmp.owl && mv components/test-with-sssom.owl.tmp.owl components/test-with-sssom.owl
INVALID ONTOLOGY FILE ERROR Could not load a valid ontology from file: test-with-sssom.owl-converted.sssom.owl
For details see: http://robot.obolibrary.org/errors#invalid-ontology-file-error
Use the -vvv option to show the stack trace.
Use the --help option to see usage information.
make[1]: *** [Makefile:471: components/test-with-sssom.owl] Error 1
make[1]: Leaving directory '/work/src/ontology'
make: *** [Makefile:442: recreate-components] Error 2
```

The problem is here:

```
sssom convert tmp/test-with-sssom.owl-merged.sssom.tsv  --output tmp/test-with-sssom.owl-converted.sssom.owl
```

This is supposed to convert the SSSOM set into a OWL file, and obviously assumes that `sssom convert` behaves a bit like ROBOT in that it uses the extension of the output file to infer the file format. That is, however, not the case. Without an explicit `--output-format FMT` option, `sssom convert` writes a SSSOM/TSV file, which ROBOT obviously cannot read, hence the `INVALID ONTOLOGY FILE ERROR`.

gouttegd:
This PR slightly refactors the _Components_ section of the standard Makefile to:

* use the SSSOM ROBOT plugin to create SSSOM-derived components (which automatically fixes #1213);
* ensure that components are forcefully rebuilt when invoking one of the `recreate-*` targets without relying on Make’s `-B` option (#1207);
* move some logic out of the template and into the `odk.py` script to “lighten” the template;
* apply the new standard template formatting rules (e.g. indenting conditional blocks).


template/src/ontology/Makefile.jinja2
```diff
@@ -749,60 +750,76 @@ all_components: $(OTHER_SRC)
 
 .PHONY: recreate-components
 recreate-components:
-	$(MAKE) COMP=true IMP=false MIR=true PAT=true IMP_LARGE=false all_components -B
+	$(MAKE) {% for component in project.components.products %}--assume-new=$(TMPDIR)/stamp-component-{{ component.filename }} \
```
matentzn, line 753: We should really make a `make` reference with "useful make tricks for ontology pipelines" with all the could things you have taught me in the last years. `--assume-new`? WOW.

template/src/ontology/Makefile.jinja2
```diff
@@ -749,60 +750,76 @@ all_components: $(OTHER_SRC)
...
 .PRECIOUS: $(COMPONENTSDIR)/{{ component.filename }}
-{% elif component.use_mappings %}
-$(COMPONENTSDIR)/{{ component.filename }}:{% if component.mappings is not none %}{% for mapping in component.mappings %} $(MAPPINGDIR)/{{ mapping }}{% endfor %}{% else %} $(MAPPINGDIR)/{{ component.filename.split('.') | first }}sssom.tsv{% endif %}
-	$(SSSOMPY) merge $^ --output $(TMPDIR)/{{ component.filename }}-merged.sssom.tsv &&\
-		$(SSSOMPY) convert $(TMPDIR)/{{ component.filename }}-merged.sssom.tsv {{ component.sssom_tool_options }} --output $(TMPDIR)/{{ component.filename }}-converted.sssom.owl &&\
-		$(ROBOT) merge -i $(TMPDIR)/{{ component.filename }}-converted.sssom.owl \
-		$(ANNOTATE_CONVERT_FILE)
 
+{%     elif component.use_mappings -%}
+$(COMPONENTSDIR)/{{ component.filename }}:{% for mapping in component.mappings %} $(MAPPINGDIR)/{{ mapping }}{% endfor %} $(TMPDIR)/stamp-component-{{ component.filename }} | all_robot_plugins
```
matentzn, line 811: I see `{% for mapping in component.mappings %}` was removed (like in the respective templates section above, which I am assuming was redundant.

matentzn (approved):
Aweeeeesome! Looks great! Mini suggestion, and my usual chatter.

odk/odk.py
```diff
@@ -484,6 +484,15 @@ def _add_stub(self, filename : str):
             self.products = []
         self.products.append(ComponentProduct(filename=filename))
 
+    def _derive_fields(self, project):
+        for product in self.products:
+            if product.base_iris is None:
+                product.base_iris = [project.uribase + '/' + project.id.upper()]
```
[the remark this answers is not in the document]
gouttegd: This should be a distinct issue. The behaviour implement here preserves the _existing_ behaviour in the current Makefile template.

We are absolutely _not_ consistent in the way we build a default base IRI. Sometimes we do add a trailing `_` (as when we [generate ROBOT reports](https://github.com/INCATools/ontology-development-kit/blob/4cf23557deb6c41b4e096997744a85ad58aad8e2/template/src/ontology/Makefile.jinja2#L356)), and sometimes we don’t (as when we [download mirrors](https://github.com/INCATools/ontology-development-kit/blob/4cf23557deb6c41b4e096997744a85ad58aad8e2/template/src/ontology/Makefile.jinja2#L846), or here when we create components). Also, sometimes we use `$OBOBASE`, sometimes we use `$URIBASE`.

Having a simple, consistent approach would probably be a good idea, but it should be done in its own issue.

template/src/ontology/Makefile.jinja2
gouttegd: The reference exists, it’s called the GNU Make Manual. :stuck_out_tongue:  Available from `info make` on all decent operating systems, or [here](https://www.gnu.org/software/make/manual/html_node/index.html).

template/src/ontology/Makefile.jinja2
gouttegd: I don’t see what you mean. Which `{% for mapping in component.mappings %}` was removed? It’s right here.

odk/odk.py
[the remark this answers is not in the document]
matentzn: Ok, if that is so, defer to your judgement.

template/src/ontology/Makefile.jinja2
gouttegd: Ah, OK. Then yes, it is now redundant since the `odk.py` script now takes care of ensuring that `component.mappings` is always initialised to a default value if `component.use_mapping` is set.

That’s one the general rules I want to apply from now on throughout the templates: whenever possible, move most logic _out_ of the templates, because it’s easier to maintain that way.

Something like this:

```python
if product.use_template and product.templates is None:
    product.templates = [product.filename.split('.')[0] + '.tsv']
```

```Jinja2
{% for mapping in component.mappings %} $(MAPPINGDIR)/{{ mapping }}{% endfor %}
```

is much more readable/maintainable than

```Jinja2
{% if component.mappings is not none %}{% for mapping in component.mappings %} $(MAPPINGDIR)/{{ mapping }}{% endfor %}{% else %} $(MAPPINGDIR)/{{ component.filename.split('.') | first }}sssom.tsv{% endif %}
```

matentzn: Makes sense! Cool!

PR #1216, opened 2025-04-15
Remove remaining uses of `make -B`

gouttegd:
This PR:

* ensures that imports are forcefully refreshed when invoking one of the `refresh-*` targets without relying on Make’s `-B` option;
* removes the advice to use said option in the inline help.


matentzn (approved):
LGTM

PR #1218, opened 2025-04-17
Allow updating the .gitignore file.

### LINKED ISSUE #847: Updating .gitignore when updating a repository
Currently, when initialising a new repository the seeding process generates a `.gitignore` file, but when updating an existing repository the `.gitignore` file is _not_ updated – instead, migrating it is left as an exercise to the reader.

The rationale for not updating it automatically is that users might have added their own exclusion rules, which we do not want to erase. But the downside is then that any new exclusion rule added to the ODK template will not be added to existing repositories, _unless_ users take the time to manually update their `.gitignore` file – I strongly suspect nobody has ever bothered to do that.

There would be at least two ways of allowing automatic update of the ODK exclusion rules while preserving any user-specified rules.

# Using a distinct file for ODK rules and for user-specified rules

This is basically the same approach as the one used for `Makefile` rules: have one file that is entirely ODK-managed (`src/ontology/Makefile`) and therefore can be overwritten when updating, and a separate file that is entirely user-managed (`src/ontology/$ONT.Makefile`) and therefore is never overwritten.

This can be achieved by setting the Git configuration option `core.excludesFile` to, say, `.gitignore-odk` (this can be done when initialising the repository; when updating, we can check whether that configuration option has been set, and set it if needed). We can then put all ODK-managed rules in that file (which we can overwrite when updating). User-managed custom rules would still go into the standard `.gitignore` file, which we would never touch.

The downside is that users may already use a system-wide (or rather, “user session”-wide) `core.excludesFile` file (which by default is `$XDG_CONFIG_HOME/git/ignore`). That file would no longer be used in ODK repositories where `core.excludesFile` is set by the ODK, which may surprise those users.

# Carrying over custom exclusion rules when updating

The `update_repo.sh` script could do something as follows:

1. compare the existing `.gitignore` file with the one generated by the seeding step, and extract all rules that are not present in the ODK-generated file (those rules are assumed to be user-written rules);
2. catenate the ODK-generated file with the rules extracted in step 1.

The downside is that this would not allow to _remove_ rules in the ODK template (removed rules would be assumed to be user-written rules and therefore would always be carried over).

# Have commented section markers in the `.gitignore` file

A slightly more refined variant would be to use dedicated “section markers“ in the `.gitgnore` file, such as:

```
# ODK exclusion rules, do not modify
[all ODK-managed rules...]
# End of ODK exclusion rules
# You may add your own exclusion rules below that line
```

The update script could then use those markers to find which rules need to be carried over and which rules can be entirely replaced by the new ODK-generated file.

Compared to the previous solution, this fixes the issue of rules removal, but raises the question of what to do when we first update a repository where those markers do not exist yet.

gouttegd:
This PR implements the solution proposed [here](https://github.com/INCATools/ontology-development-kit/issues/847#issuecomment-2812670154) to make the `.gitignore` file partially “managed“ – that is, the file can be updated when upgrading to a newer ODK version.

This is done by

(1) adding “section markers” to the `.gitignore` template, to delineate the part of the file that is managed by the ODK;

(2) amending the update code in the `odk.py` script to explicitly update an existing `.gitignore file`; the update procedure completely overwrites the ODK-managed section, but preserves any ignore rule found outside of that section (unless it is a duplicate of a managed rule).

closes #847


odk/odk.py
```diff
@@ -1197,7 +1228,8 @@ def update(templatedir):
             ('src/ontology/Makefile', ALWAYS),
             ('src/ontology/run.sh', ALWAYS),
             ('src/sparql/*', ALWAYS),
-            ('docs/odk-workflows/*', ALWAYS)
+            ('docs/odk-workflows/*', ALWAYS),
+            ('.gitignore', NEVER)
```
matentzn, line 1232: Never?

matentzn (commented):
This looks great but what happens the very first time this is run?

gouttegd:
> what happens the very first time this is run?

You mean when the existing `.gitignore` file does not contain any marker lines?

Then _all_ lines from the existing file are assumed to be user-managed rules, so they will be added at the bottom of the new file below the ODK-managed rules — _unless_ they happen to be identical to ODK-managed rules, in which case they will be skipped so that they are not duplicated.

Practically, we can consider two scenarios:

(A) The user never added anything to her `.gitignore` file since she seeded her repository. The file therefore only contains ODK rules, albeit not delimited by any marker.

Then the update process will:

(1) keep all the rules in a temporary space, because it assumes they are user rules;
(2) write the new template-derived, ODK-managed rules (now between markers);
(3) attempt to write back the user rules saved in step (1), only to realise that they are identical to the rules already written in step (2), so in fact do nothing.

Therefore, a `.gitignore` file like this:

```
.DS_Store
semantic.cache
bin/
[... all the other rules that were installed when the repository was seeded ...]
```

will become

```
# ODK-managed rules, do not modify
# If you need to add your own ignore rules, you may do so below
# the "End ODK-managed rules" marker at the end of this file.
.DS_Store
semantic.cache
bin/
[... all the other rules that were installed when the repository was seeded ...]
# End of ODK-managed rules
```

(B) The user did add some of her own rules in the `.gitignore` file.

Then the update process will:

(1) keep all the rules in a temporary space, because it assumes they are user rules (and indeed some of them really are);
(2) write the new template-derived, ODK-managed rules (now between markers);
(3) attempt to write back the user rules saved in step (1), realise that most of them are identical to the rules already written in step (2), and only write the remaining rules.

Therefore, a `.gitignore` file like this:

```
.DS_Store
semantic.cache
bin/
[... all the other rules that were installed when the repository was seeded ...]

src/ontology/a_large_custom_file_user_does_not_to_commit
```

will become

```
# ODK-managed rules, do not modify
# If you need to add your own ignore rules, you may do so below
# the "End ODK-managed rules" marker at the end of this file.
.DS_Store
semantic.cache
bin/
[... all the other rules that were installed when the repository was seeded ...]
# End of ODK-managed rules

src/ontology/a_large_custom_file_user_does_not_to_commit
```

PR #1231, opened 2025-05-10
Update the 'build-multiarch' workflow.

gouttegd:
The `build-multiarch` workflow had been used for a brief time in 2021 to automatically build and publish the Docker images to the Docker Hub directly from GitHub.

We stopped using it after PR #475, because by then building the multiarch images took too long (more than 6 hours) and the process always ended up being killed by GitHub.

However, (1) this was nearly 4 years ago, so maybe the runners are now faster, and (2) more importantly, 4 years ago we were building Konclude on arm64 from source, which was the most time-consuming part of the build. We are no longer doing that, instead we use a pre-built version of Konclude for arm64.

So it is worth trying to reactivate that workflow, which is what this PR is about. Here, we update the Docker-provided actions used by the workflow, and add a parameter to allow choosing between building the "release" images (as the original workflow) or building a development snapshot.

### DIFF
--- a/.github/workflows/build-multiarch.yml
+++ b/.github/workflows/build-multiarch.yml
@@ -2,6 +2,15 @@ name: Build and publish the multi-arch ODK images
 
 on:
   workflow_dispatch:
+    inputs:
+      buildType:
+        description: 'Type of image to build'
+        required: true
+        default: 'release'
+        type: choice
+        options:
+          - release
+          - snapshot
 
 jobs:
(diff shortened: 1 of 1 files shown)


matentzn (approved):
Of course, try it!

gouttegd:
Doesn’t look too good… Two hours into the build and it has not even finish the `odkbuild` image…

Well, it was worth a try.

PR #1233, opened 2025-05-15
Update ROBOT

### LINKED ISSUE #1210: Update ROBOT and adapt workflows accordingly
When the next version of ROBOT will be released, it will include changes that should be reflected in the ODK workflows:

* Replace all uses of `odk:normalize --add-source true` by the new built-in “interpolation” feature of the `annotate` command (https://github.com/ontodev/robot/pull/1242).
* Replace all uses of `odk:normalize --merge-axioms` by the standard command `repair --merge-axiom-annotations` (which has been fixed in https://github.com/ontodev/robot/pull/1240).
* Add the new `--clean-obo` option (https://github.com/ontodev/robot/pull/1236 when exporting to OBO format.
* (If https://github.com/ontodev/robot/pull/1248 gets merged in time for that release) Add `--include-subproperties true` to `reduce` calls.

gouttegd:
This PR (1) updates the version of ROBOT bundled in the ODK to the latest 1.9.8 release (2)  updates all workflows to use features from upstream ROBOT instead of the ODK ROBOT plugin whenever possible, and (3) enables the `--include-subproperties` feature of `robot relax` by default (as announced [here](https://github.com/ontodev/robot/pull/1248#issuecomment-2711874042)).

closes #1210


template/src/ontology/Makefile.jinja2
```diff
@@ -68,6 +68,7 @@ SPARQL_VALIDATION_CHECKS =  {% for x in project.robot_report.custom_sparql_check
 SPARQL_EXPORTS =            {% for x in project.robot_report.custom_sparql_exports|default(['basic-report', 'class-count-by-prefix', 'edges', 'xrefs', 'obsoletes', 'synonyms']) %}{{ x }} {% endfor %}
 ODK_VERSION_MAKEFILE =      {% if env is defined %}{{env['ODK_VERSION'] or "Unknown" }}{% else %}"Unknown"{% endif %}
 RELAX_OPTIONS =             --include-subclass-of true
+REDUCE_OPTIONS =            --include-subproperties true
```
matentzn, line 71: This will need quite a lot of debate; just FYI. Not saying anything for or against it, just putting out a warning of changing the default behaviour of reduce in ODK.

matentzn (commented):
I was almost saying this was perfect but there is one half line that I am not convinced is worth the risk..

template/src/ontology/Makefile.jinja2
matentzn: (I know the discussion - but this is a big change; I would suggest to keeping the current default and making it configurable to include the new behaviour)

gouttegd: You realise that _nobody_ is looking at what options are available in the ODK, do you?

Keeping the old behaviour by default and hiding the new one behind an option is a sure way to guarantee that all ontologies will keep the default behaviour (well, except the FlyBase ones), because their owners won’t even know they can change it.

That ROBOT ignores SubObjectPropertyOf axioms when reducing is a _bug_; it has been [reported](https://github.com/ontodev/robot/issues/1014) [as such](https://github.com/ontodev/robot/issues/1208). I’m fixing it.

People who liked the buggy behaviour can opt in to keep it. All they have to do is to redefine `REDUCE_OPTIONS` in their custom Makefile. This will be explained in the release notes.

I am open to the idea of making it configurable at the level of the `-odk.yaml` file rather than the Makefile, but I maintain that the default behaviour should be the not-buggy one.

matentzn: As I said, I am not arguing against this, I am only not able to wave it through without deeply thinking through the ramifications.

If the option was at least configurable at config file level, I would probably think faster :P

gouttegd: Fine, I’ll add a config-level option. I’ll add one as well for the `RELAX_OPTION` then, since we are at it.

matentzn (approved):
This looks great, thank you @gouttegd !

PR #1234, opened 2025-05-15
Add yq tool to odkfull image

### LINKED ISSUE #838: Add yq (YAML version of jq) to ODK
https://github.com/mikefarah/yq

wget https://github.com/mikefarah/yq/releases/download/${VERSION}/${BINARY} -O /usr/bin/yq &&\
    chmod +x /usr/bin/yq

This will be useful when building applications on top of mapping commons which will make use of a lot of yaml.

matentzn:
The yq tool (https://github.com/mikefarah/yq) is a widely used command-line YAML, JSON and XML processor. Adding the tool, and a test.

Fixes #838

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -129,12 +129,13 @@ RUN chmod +x /tools/obodash && \
 # Install KGCL ROBOT plugin
 RUN wget -nv -O /tools/robot-plugins/kgcl.jar https://github.com/gouttegd/kgcl-java/releases/download/kgcl-java-$KGCL_JAVA_VERSION/kgcl-robot-plugin-$KGCL_JAVA_VERSION.jar
 
+# Install Mike Farah's (mf) YQ command-line YAML, JSON and XML processor
 RUN if [ "$TARGETARCH" = "amd64" ]; then \
-        wget -nv https://github.com/mikefarah/yq/releases/download/v$YQ_VERSION/yq_linux_amd64 -O /tools/yq && \
-        chmod 0755 /tools/yq ; \
+        wget -nv https://github.com/mikefarah/yq/releases/download/v$YQ_VERSION/yq_linux_amd64 -O /tools/yq-mf && \
+        chmod 0755 /tools/yq-mf ; \
     elif [ "$TARGETARCH" = "arm64" ]; then \
-        wget -nv https://github.com/mikefarah/yq/releases/download/v$YQ_VERSION/yq_linux_arm64 -O /tools/yq && \
-        chmod 0755 /tools/yq ; \
+        wget -nv https://github.com/mikefarah/yq/releases/download/v$YQ_VERSION/yq_linux_arm64 -O /tools/yq-mf && \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -127,3 +128,14 @@ RUN chmod +x /tools/obodash && \
...
 RUN wget -nv -O /tools/robot-plugins/kgcl.jar https://github.com/gouttegd/kgcl-java/releases/download/kgcl-java-$KGCL_JAVA_VERSION/kgcl-robot-plugin-$KGCL_JAVA_VERSION.jar
+
+RUN if [ "$TARGETARCH" = "amd64" ]; then \
+        wget -nv https://github.com/mikefarah/yq/releases/download/v$YQ_VERSION/yq_linux_amd64 -O /tools/yq && \
+        chmod 0755 /tools/yq ; \
+    elif [ "$TARGETARCH" = "arm64" ]; then \
+        wget -nv https://github.com/mikefarah/yq/releases/download/v$YQ_VERSION/yq_linux_arm64 -O /tools/yq && \
+        chmod 0755 /tools/yq ; \
+    else \
+        echo "Unsupported TARGETARCH: $TARGETARCH" && exit 1 ; \
```
gouttegd, line 139: There’s no plan (for now and the foreseeable future) to ever try building the ODK on anything else than x86-64 and arm64, but I like the carefulness. :)

Dockerfile
```diff
@@ -127,3 +128,14 @@ RUN chmod +x /tools/obodash && \
 
 # Install KGCL ROBOT plugin
 RUN wget -nv -O /tools/robot-plugins/kgcl.jar https://github.com/gouttegd/kgcl-java/releases/download/kgcl-java-$KGCL_JAVA_VERSION/kgcl-robot-plugin-$KGCL_JAVA_VERSION.jar
+
+RUN if [ "$TARGETARCH" = "amd64" ]; then \
+        wget -nv https://github.com/mikefarah/yq/releases/download/v$YQ_VERSION/yq_linux_amd64 -O /tools/yq && \
```
gouttegd, line 133: Annoyingly enough, there is also a [Python package also called yq](https://github.com/kislyuk/yq), which also installs an executable called `yq`. That package is _not_ provided in the ODK, but still I don’t like the risk of confusion (e.g. if someone installs the Python package in a custom workflow, they would probably expect that a call to `yq` invokes the Python script rather than this `yq` ).

Could we rename the binary to something like `mf-yq` (Mike Farrah’s `yq`), `yq-go` (`yq` in GO), or similar?

gouttegd (dismissed):
Looks fine (tested on arm64 only for now), but please consider renaming the binary (see comment below).

Dockerfile
matentzn: Makes sense, did!

gouttegd:
Merging despite the CI failure since (1) said failure has nothing to do with the PR, but is instead due to our test suite running against the rate limits now enforced by GitHub, and (2) I tested locally and it is fine.

PR #1235, opened 2025-05-15
Add OBO Extended Prefix Map to ODK

### LINKED ISSUE #969: Distribute OBO context / and extended prefix map in ODK
From @gouttegd 

https://github.com/obophenotype/uberon/pull/3141#issuecomment-1853891911

We should make a plan on how to best synchronise this extended prefix map (EPM) across resources. Right now we sync it in sssom-py. The reason for this is that I want to review every single change to the EPM myself, so I can go and intervene if I don't like something (say, the expansion of a new prefix being added).

The question is where this "protected" epm should live ideally. ODK is certainly an option!

matentzn:
Fixes #969 

I hope this captures our discussion @gouttegd.


.github/workflows/update-epm.yml
```diff
@@ -0,0 +1,28 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        
+      - name: Work around permission issue
+        run: git config --global --add safe.directory "$GITHUB_WORKSPACE"
+
+      - name: Update OBO context
+        run: |
+          wget "https://raw.githubusercontent.com/biopragmatics/bioregistry/main/exports/contexts/obo.epm.json" -O obo.epm.json
```
gouttegd, line 18: If you want to commit the prefix map to the repository (which is a good idea for build reproducibility, no argument here), please don’t write it in the top-level directory (which is cluttered enough already).

I’d suggest a `resources/` subdirectory or similar.

template/src/ontology/Makefile.jinja2
```diff
@@ -1518,6 +1518,12 @@ validate-all-tsv: $(ALL_TSV_FILES)
 # Editors Utilities
 # ----------------------------------------
 
+# This is an experimental target people that want to use ODK extended prefix map
+# can use to pull the (currently inofficial) OBO extended prefix map into the workspace
+# Only works with odkfull variant
+tmp/obo.epm.json: /tools/obo.epm.json
```
gouttegd, line 1524: I’d prefer if users avoided to refer to `tmp/obo.epm.json`, just in case we’d want to change the location later (this PR makes me think that we could later want to generalise the idea of “ODK-provided resources”, and thus maybe have a dedicated `resources` directory somewhere).

I’d suggest having a `EXTENDED_PREFIX_MAP` Makefile variable pointing to `tmp/obo.epm.json`, and users being instructed to only use that variable instead `tmp/obo.epm.json` directly.

matentzn: Alright, I hope I did that correctly!

.github/workflows/update-epm.yml
matentzn: Done!

docker/odklite/Dockerfile
```diff
@@ -110,3 +110,4 @@ COPY --chmod=755 scripts/update_repo.sh /tools/update_repo
 COPY --chmod=755 odk/make-release-assets.py odk/odk.py /tools
 COPY template/ /tools/templates/
 CMD python3 /tools/odk.py
+COPY obo.epm.json /tools
```
gouttegd, line 113:
```suggestion
COPY resources/obo.epm.json /tools
```

template/src/ontology/Makefile.jinja2
gouttegd: Looks good!

template/src/ontology/Makefile.jinja2
```diff
@@ -1518,6 +1519,13 @@ validate-all-tsv: $(ALL_TSV_FILES)
 # Editors Utilities
 # ----------------------------------------
 
+# This is an experimental target people that want to use ODK Extended Prefix Map (EPM)
+# can use to pull the (currently inofficial) OBO EPM into the workspace.
```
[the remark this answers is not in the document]
matentzn: Yes totally agreed, we could even make it configurable. For now, the command is just saying that the OBO EPM is pulled in, which is accurate; but we agree on the intention.

gouttegd (approved):

PR #1239, opened 2025-05-16
Extracts the version IRI of component mirrors

matentzn:
- [ ] Depends on #1233 

I am suggesting to inject the version IRI as dc:source in component mirrors as well, so the source version information is not lost downstream.

This patterns will be used soon for all imports, just want to make sure it is used for the component mirrors as well. I hope this wont crash when there is no version iri.


gouttegd:
No objection to the idea, but for imports, we are not merely _adding_ a `dc:source` annotation: we are also _removing_ all existing ontology annotations, so that the `dc:source` we add is the only one remaining in the final import module.

I note that you are not doing that here. This means for example that if a downloaded component already contains a `dc:source` annotation, it will now have two such annotations. Not sure if it will end up being helpful or confusing.

Also, not convinced this is in fact useful for components. It may be needed for imports because they are _derived_ from a foreign ontology (e.g. by SMLE extraction), and (as said above) we are removing their original annotations in the process, so it makes sense to add a `dc:source` annotation for traceability.

But components fetched from remote sources are used “as they are”. The only change made to them is the setting of their ontology IRI – which makes sense because components are considered to be part of the ontology they are in. That’s precisely the distinction that the ODK makes between a “import” (a _foreign_ ontology, not under the control of the importing one) and a “component” (an integral part of the present ontology, that just happens to be in a separate file).

So overall, while I am, again, not _opposed_ to the idea, I am not convinced it is in fact needed.

gouttegd:
> I hope this wont crash when there is no version iri.

This will not crash, but it will insert the original, unexpanded string, so you would end up with a version IRI that is _literally_ `%{version_iri}`.

Which, now that we are talking it, is worrying me a bit. My original `odk:normalize --add-source` command in the ODK plugin had a more sensible behaviour, in that it would _not_ inject a `dc:source` annotation at all if it did not find a version IRI in the ontology.

template/src/ontology/Makefile.jinja2
```diff
@@ -789,7 +789,8 @@ component-download-{{ component.filename }}: | $(TMPDIR)
 	$(ROBOT) merge -I {{ component.source }} \{% if component.make_base %}
 		 remove {% for iri in component.base_iris %}--base-iri {{ iri }} \
 		        {% endfor %}--axioms external --preserve-structure false --trim false \{% endif %}
-		 annotate --ontology-iri $(ONTBASE)/$@ \
+		 annotate --link-annotation dc:source "%{version_iri}" \
+		 		  --ontology-iri $(ONTBASE)/$@ \
```
gouttegd, line 793: You’re missing the `--interpolate true` flag.

gouttegd (changes requested):
To consider:

1) Do you think pre-existing ontology annotations (including, and maybe especially, a potential `dc:source`) should be removed, like we do for imports? If not, then I am even less sure there is a need for adding a `dc:source`, since original annotations would still be there (only the ontology IRI would be changed).

2) If there is a risk that a component may not have a version IRI, then we should rather use `odk:normalize --add-source`, which will avoid injecting a literal `%{version_iri}` version IRI. The interpolation feature in upstream ROBOT is nice because it is generic, but the price of that genericity is that it doesn’t check whether it actually has something to inject.

3) Even if prefer to keep using the upstream feature, in that case you need the `--interpolate true` flag.

gouttegd:
> But components fetched from remote sources are used “as they are”. The only change made to them is the setting of their ontology IRI – which makes sense because components are considered to be part of the ontology they are in. That’s precisely the distinction that the ODK makes between a “import” (a foreign ontology, not under the control of the importing one) and a “component” (an integral part of the present ontology, that just happens to be in a separate file).

In fact, for components that are fetched from remote sources I wonder if it would not be better, for traceability, to inject as `dc:source` the URL from which they are fetched.

Consider the case of CL’s `hra_subset` component. It is downloaded from here: https://raw.githubusercontent.com/hubmapconsortium/ccf-validation-tools/master/owl/CL_ASCTB_subset.owl

Note that it does not have a version IRI or even an ontology IRI. Shouldn’t we inject the URL above as `dc:source` then?

matentzn:
Hmmm... What I was _hoping_ was that I always knew which version of the upstream file the downstream (component, mirror, import) was created from. The raw.github file is symlink so you wont know from that which _exact_ version (aka commit) was used to get it from.

So I think what we do with the versionIRI interpolation is mostly correct; its just not ideal that it creates a literal when there is no version IRI present.

For this PR, I am radically changing the approach at least to just _not immediately overwriting the versionIRI - this way I can, at least if absolutely want to, look at the initial component mirror to figure out their version.

This is not addressing any of your comments though; my preferred option in most cases is still to use the version IRI as you currently are.

gouttegd:
> What I was hoping was that I always knew which version of the upstream file the downstream (component, mirror, import) was created from.

This is only possible if:

* whoever is maintaining the upstream file does bother with versioning it (and ”versioning” does not simply mean “put it in a version control system like Git” — it means attributing actual version numbers), which for example is not the case of the aforementioned `hra_subset.owl` file;
* whoever is fetching the upstream file into their own ontology downloads a precise version, instead of taking the file from a branch.

If the file does not provide itself any version information and you’re downloading it from a _branch_, then the “version” of the component you get is “the state in which the component was in that branch at the moment I run `refresh-components`” – you have no way to know anything better.

Alas, this is precisely what Uberon and CL are doing with their remote-fetched components. AFAIK they are _all_ fetched from master branches.

What I am trying to say is that if you want traceability of components (which would be a very good thing, no argument here!), what you actually need is to get people to stop using remote-fetched components in YOLO mode, “let’s just download the latest file from the top of the master branch, that will always be fine!”

In fact, I feel like people have “hijacked” the concept of components to treat them as version-less imports (something you can download at anytime to always get the most up-to-date “versions”, without having to wait for the upstream maintainers to make a formal release). This is terrible for traceability, and you won’t fix that with a `dc:source` injection.

gouttegd (approved):
Approving since the new behaviour (2nd commit) is sensible.

When merging, please do a squash rather than a standard merge (or a rebase). This is perfectly appropriate here and there is no need to keep a trace of the original commit which did something completely different from what the PR ended up to be.

PR #1240, opened 2025-05-16
Add merge_template workflow to Makefile

### LINKED ISSUE #579: Add ROBOT template merge pipeline as option to ODK
The purpose is to bulk add term data provided in the form of a table to an ontologies editors file, like in MONDO, RO and other ontologies.

matentzn:
fixes #579 

The `merge_template` workflow is in no way dependent on any other rule - isolated from the rest of the Makefile - so this addition is low risk.

But I am not 1000% set on this PR. I know in Mondo we use this workflow a lot; I have used it sometimes in HPO, MAXO and OMO.

Let me know your thoughts @gouttegd!

### DIFF
--- a/template/src/ontology/Makefile.jinja2
+++ b/template/src/ontology/Makefile.jinja2
@@ -1532,7 +1532,12 @@ TEMPLATE_URL=
 # The tmp/merge_template.tsv is hardcoded as placeholder for downloading template files
 # from a remote location
 tmp/merge_template.tsv:
-	wget "$(TEMPLATE_URL)" -O $@
+	@if [ -z "$(TEMPLATE_URL)" ]; then \
+		echo "ERROR: No TEMPLATE_URL defined. Please set TEMPLATE_URL to download the merge template."; \
+		exit 1; \
+	else \
+		wget "$(TEMPLATE_URL)" -O $@; \
+	fi
 
 merge_template: $(MERGE_TEMPLATE)
 	$(ROBOT) template $(MERGE_TEMPLATE_OPTIONS) --input $(SRC) \
--- a/template/src/ontology/Makefile.jinja2
+++ b/template/src/ontology/Makefile.jinja2
@@ -1522,10 +1522,17 @@ validate-all-tsv: $(ALL_TSV_FILES)
 # The merge template workflow is commonly used when ontology content is curated in a ROBOT template
 # with the intention to merge it directly into the editors file. The idea is to either create a template
 # manually or create a rule in {{ project.id }}.Makefile to download a template from, e.g., a Google Sheet, and then run
-# `./run.sh MERGE_TEMPLATE=tmp/merge_my_template.tsv merge_template` to merge the template into the editors file.
+# `sh run.sh make MERGE_TEMPLATE=tmp/merge_my_template.tsv merge_template` to merge the template into the editors file or
+# `sh run.sh make TEMPLATE_URL=http://example.org/merge_my_template.tsv merge_template` to download and then merge the template
 
 MERGE_TEMPLATE=tmp/merge_template.tsv
 MERGE_TEMPLATE_OPTIONS= --merge-before
+TEMPLATE_URL=
+
+# The tmp/merge_template.tsv is hardcoded as placeholder for downloading template files
+# from a remote location
+tmp/merge_template.tsv:
+	wget "$(TEMPLATE_URL)" -O $@
 
 merge_template: $(MERGE_TEMPLATE)
 	$(ROBOT) template $(MERGE_TEMPLATE_OPTIONS) --input $(SRC) \
--- a/template/src/ontology/Makefile.jinja2
+++ b/template/src/ontology/Makefile.jinja2
@@ -1519,6 +1519,18 @@ validate-all-tsv: $(ALL_TSV_FILES)
 # Editors Utilities
 # ----------------------------------------
 
+# The merge template workflow is commonly used when ontology content is curated in a ROBOT template
+# with the intention to merge it directly into the editors file. The idea is to either create a template
+# manually or create a rule in {{ project.id }}.Makefile to download a template from, e.g., a Google Sheet, and then run
+# `./run.sh MERGE_TEMPLATE=tmp/merge_my_template.tsv merge_template` to merge the template into the editors file.
+
+MERGE_TEMPLATE=tmp/merge_template.tsv
+MERGE_TEMPLATE_OPTIONS= --merge-before
+
+merge_template: $(MERGE_TEMPLATE)
+	$(ROBOT) template $(MERGE_TEMPLATE_OPTIONS) --input $(SRC) \
+			--template $(MERGE_TEMPLATE) convert -f $(EDIT_FORMAT) -o $(SRC)
+
 # This is an experimental target people that want to use ODK Extended Prefix Map (EPM)
 # can use to pull the (currently inofficial) OBO EPM into the workspace.
 # Users are instructed to refer to the EPM only through the variable $(EXTENDED_PREFIX_MAP) as 


template/src/ontology/Makefile.jinja2
```diff
@@ -1519,6 +1519,25 @@ validate-all-tsv: $(ALL_TSV_FILES)
 # Editors Utilities
 # ----------------------------------------
 
+# The merge template workflow is commonly used when ontology content is curated in a ROBOT template
+# with the intention to merge it directly into the editors file. The idea is to either create a template
+# manually or create a rule in {{ project.id }}.Makefile to download a template from, e.g., a Google Sheet, and then run
+# `sh run.sh make MERGE_TEMPLATE=tmp/merge_my_template.tsv merge_template` to merge the template into the editors file or
+# `sh run.sh make TEMPLATE_URL=http://example.org/merge_my_template.tsv merge_template` to download and then merge the template
+
+MERGE_TEMPLATE=tmp/merge_template.tsv
```
gouttegd, line 1528:
```suggestion
MERGE_TEMPLATE=$(TMPDIR)/merge_template.tsv
```

template/src/ontology/Makefile.jinja2
```diff
@@ -1519,6 +1519,25 @@ validate-all-tsv: $(ALL_TSV_FILES)
...
+MERGE_TEMPLATE_OPTIONS= --merge-before
```
gouttegd, line 1529: Is there really a need for that to be configurable in a variable? Why would anyone want to change that?

If they change it to nothing, then the `template` command below would default to not merging the instantiated template with the input ontology, which means that they would entirely replace their -edit file with the instantiated template! How can that be something that anyone would want to do?

If they change it to `--merge-after`, then it is the same effect as `--merge-before` as far as this command is concerned (`--merge-after` would only make a difference if `--output` was used, which is not the case here).

So, seems like a completely useless variable to me, which only makes things a bit more obscure that they need to be. Not all ROBOT options need to be in a Make variable.

template/src/ontology/Makefile.jinja2
```diff
@@ -1519,6 +1519,25 @@ validate-all-tsv: $(ALL_TSV_FILES)
...
+TEMPLATE_URL=
+
+# The tmp/merge_template.tsv is hardcoded as placeholder for downloading template files
+# from a remote location
+tmp/merge_template.tsv:
+	wget "$(TEMPLATE_URL)" -O $@
```
gouttegd, line 1535: I’d consider actively checking if `TEMPLATE_URL` is defined, and error out with an explicit error message in that case (something like “No URL defined for merging remote template, please define TEMPLATE_URL” or similar), rather than letting the rule fail with the error message from `wget` (when given an “empty” URL, `wget` spits a `http://: invalid hostname` message, which is correct from its point of view but not very helpful for an ODK user).

template/src/ontology/Makefile.jinja2
```diff
@@ -1519,6 +1519,25 @@ validate-all-tsv: $(ALL_TSV_FILES)
...
+
+merge_template: $(MERGE_TEMPLATE)
+	$(ROBOT) template $(MERGE_TEMPLATE_OPTIONS) --input $(SRC) \
+			--template $(MERGE_TEMPLATE) convert -f $(EDIT_FORMAT) -o $(SRC)
```
gouttegd, line 1539: Shouldn’t the `convert` part of this command includes a `--check false` option, in the case the edit format is OBO?

gouttegd (changes requested):
I don’t particularly like this, but if it used in the wild, I have no fundamental objection.

There are things that I’d rather change though, see inline comments.

template/src/ontology/Makefile.jinja2
matentzn: I did not have this before but I noticed that in all deployments of the workflow, I had to inject some prefixes into the command.. Thats why I factored it into a variable.. Otherwise I agree with you.

template/src/ontology/Makefile.jinja2
matentzn: I personally dont think so, no. It is horrible enough if obo release files are not really valid obo, but the editors file - no, I don't think there is an excuse. But I am willing to entertain counterarguments!

template/src/ontology/Makefile.jinja2
matentzn: Yes, of course: https://github.com/INCATools/ontology-development-kit/commit/afdb87143fb378d82d230b268459a2422e604264#diff-ab539569446338e97f29dddc47eccfb63d931e4efa0157d240796a7e9add1d68

template/src/ontology/Makefile.jinja2
gouttegd: So sometimes (all the times in fact, since you said “in all deployments”), you need to do this:

```
sh run.sh make MERGE_TEMPLATE_OPTIONS="--merge-before --add-prefix 'pfx: https://example.org/'" MERGE_TEMPLATE=tmp/merge_my_template.tsv merge_template
```

?

Err, then frankly, how is that workflow any simpler than invoking `robot template` directly???

gouttegd: Thinking back about this, I am now inclined to _reject_ this “feature“. Sorry, but I don’t see how a command like this:

```sh
sh run.sh make \
  MERGE_TEMPLATE_OPTIONS="--merge-before --add-prefix 'pfx: https://example.org/' \
  MERGE_TEMPLATE=tmp/merge_my_template.tsv \
  merge_template
```

would be any easier or more straightforward than:

```sh
sh run.sh robot template \
  --input myont.obo \
  --template tmp/merge_my_template.tsv \
  --add-prefix 'pfx: https://example.org/' \
  --merge-before -o myont.obo
```

Arguably it even makes things _more complex_, because now you need to know both the syntax of the original ROBOT command (so that you don’t forget the `--merge-before`, and can add the required prefixes) _and_ the variables expected by the `merge_template` rule.

Feel free to try and convince me otherwise, but as you don’t seem convinced yourself for now this is a no for me.

matentzn: No, I share your ambivalence. I think this would be much more useful in a conjunction with an _update_ workflow which is much much harder to realise. So essentially, someone wants to _change_ the ontology, but only supplying _additions_. This would be a real feature for 1.7:

1. Configure a template, potentially from URL (to support Gsheets workflows)
2. Scan the template and somehow determine what content needs to be deleted from editors file
3. Delete said content
4. Merge template into the edit 

Thats a much larger effort.

matentzn:
Closing this, as this feature does not really decrease complexity over building a custom solution.

template/src/ontology/Makefile.jinja2
gouttegd: > I think this would be much more useful in a conjunction with an update workflow which is much much harder to realise

I could imagine a more ”generic” update workflow that would work along the following lines:

* we scan the contents of a new, dedicated directory (that we could call `updates`);
* for each file in that directory, we apply a logic dependent on the file type:
  * if it is a OBO/OWL file, we assume the file contains pre-made axioms ready to be added as they are, and we merge them into the -edit file;
  * if it is a TSV file, we assume it is a template, we apply the template and merge its output to the -edit file;
  * if it is a KGCL file, we apply the changes it contains to the -edit file (that one in particular would allow _removing_ / _changing_ existing axioms, instead of only adding new ones like the previous two).

Such a workflow could even support “Gsheets workflows” (not that I think that such workflows are a good idea, personally) if we have a convention that a file with a name starting with `remote-` is supposed to contain a link pointing to the actual update file.

But in any case, definitely not something for 1.6. :D

PR #1241, opened 2025-05-17
[PoC] Allow changing the configuration schema transparently for users

gouttegd:
The schema for the ODK configuration file is a bit of a mess.

There are a lot of options directly in the top-level `project` dictionary, about wildly diverse stuff ranging from enabling validation of RDF/XML artefacts to setting the amount of memory allocated to ROBOT to setting the emails to which Travis should send failure reports, etc. That schema would benefit from some reorganisation to make things clearer, e.g. grouping all options related to GitHub in a `github` section and similar.

The problem, of course, is that any attempt at reorganising the schema by renaming or moving options around will unavoidably break existing ODK repositories the next time they try to run `update_repo`, which is a no-go: it must not be necessary for users to manually amend their `-odk.yaml` file when they update from one major version of the ODK to another (unless they _want_ to do that, say to benefit from a new feature that is disabled by default).

So, this PR is intended to experiment a system by which we could make the schema evolve as we see fit, while at all times preserving backwards compatibility with existing ODK configuration files.

The principle is simple. Currently, when we read the configuration file, we do that in two steps: (1) we load the YAML file into a Python dictionary, and (2) we instantiate a `OntologyProject` object from that dictionary.

```python
obj = yaml.load(stream, Loader=yaml.FullLoader)           # (1)
project = from_dict(data_class=OntologyProject, data=obj) # (2)
```

In this PR, we introduce an intermediate preprocessing step:

```python
obj = yaml.load(stream, Loader=yaml.FullLoader)           # (1)
preprocess_config_dict(obj)
project = from_dict(data_class=OntologyProject, data=obj) # (2)
```

The role of the preprocessing step is to transform the YAML dictionary as read from disk to make sure it is conform to what is needed to make a `OntologyProject`.

For example, let’s suppose we decide, in the next ODK version, to rename the `use_base_merging` option in the `import_group` to `merge_base`, because we decide that `merge_base` somehow sounds better (not saying we should do that, this is for the example only!). All we would need to do to preserve compatibility (and so, to make sure that existing files using `use_base_merging` are still accepted) is to declare the following tuple in the `preprocess_config_dict()` method:

```python
('import_group.use_base_merging', 'import_group.merge_base')
```

This will instruct the method to look for a `use_base_merging` key in the `import_group` subdictionary, and if such a key is found, to replace it by a key named `merge_base`.

A new command `update-config` will also give the users the possibility to actively migrate their old config files to the new syntax, should they want to do so (they would not _have_ to, since the compatibility layer ensures that files compliant with the old syntax can still be used).

As a complete example, this PR includes (2nd commit) a partial reorganisation of the schema in which all options related to ROBOT are grouped into a single `robot` section. That is, what currently looks like this in the current schema:

```yaml
reasoner: ELK
...
obo_format_options: --clean-obo 'strict drop-untranslatable-axioms'
...
robot_relax_options: --include-subclass-of true
robot_reduce_options: --include-subproperties true
...
robot_plugins:
  plugins:
    - name: my_plugin
...
robot_report:
  release_reports: True
  fail_on: ERROR
  ...
...
```

now becomes:

```yaml
...
robot:
  reasoner: ELK
  obo_format_options: --clean-obo 'strict drop-untranslatable-axioms'
  relax_options: --include-subclass-of true
  reduce_options: --include-subproperties true
  plugins:
    - name: my_plugin
  report:
    release_reports: True
    fail_on: ERROR
...
```

### DIFF
--- a/odk/odk.py
+++ b/odk/odk.py
@@ -615,13 +615,29 @@ class RobotPlugin(JsonSchemaMixin):
 
 @dataclass_json
 @dataclass
-class PluginsGroup(JsonSchemaMixin):
+class RobotOptionsGroup(JsonSchemaMixin):
     """
-    A configuration section to list extra ROBOT plugins not provided by the ODK
+    A configuration section for additional options specific to ROBOT.
     """
 
+    reasoner : str = "ELK"
+    """Reasoner to use in robot commands that need one."""
+
+    obo_format_options : str = ""
+    """Additional options to pass to robot convert when exporting to OBO. Default is '--clean-obo "strict drop-untranslatable-axioms"'."""
+
+    relax_options : str = "--include-subclass-of true"
+    """Additional options to pass to robot relax command."""
+
+    reduce_options : str = "--include-subproperties true"
+    """Additional options to pass to robot reduce command."""
+
     plugins : Optional[List[RobotPlugin]] = None
-    """The list of plugins to use"""
+    """List of ROBOT plugins used by this project."""
+
+    report : Dict[str, Any] = field(default_factory=lambda: ReportConfig().to_dict())
+    """Settings for ROBOT report, ROBOT verify and additional reports that are generated."""
+
 
     
 @dataclass_json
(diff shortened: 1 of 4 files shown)


gouttegd:
:arrow_up: @matentzn For your consideration.

I’d love to reorganise the schema to tidy things up, but we can’t do that without a way to preserve compatibility with existing configuration files. What do you think of that solution?

PR #1246, opened 2025-05-20
Ignore additional release files.

### LINKED ISSUE #1238: Subsets, mappings sets and pattern-derived artefacts should be .gitignored by default
The `prepare_release` target copies all the built release artefacts from the `src/ontology` directory to the top-level directory.

This step includes the copying of:

a) any subsets in the `src/ontology/subsets` directory to the top-level `subsets` directory;

b) any releasable mapping set (a set marked in the config with `release_mappings=True`, or all sets if `release_mappings=True` is set on the entire `sssom_mappingset_group`) from the `src/mappings` directory to the top-level `mappings` directory;

c) pattern-derived files (`definitions.owl` and `pattern.owl`) from the `src/patterns` directory to the top-level `patterns` directory.

Those three sets of files are currently _not_ ignored by Git (they are not in `.gitignore`), meaning that they are set to be committed to the repository by default.

We decided a little while ago that release artefacts should not, _by default_, be committed (#1144), but the associated PR (#1183) only took care of the “main” release artefacts, completely ignoring the subsets, the mappings, and the pattern-derived files.

It makes no sense to commit “secondary” release artefacts while the “main” artefacts are not committed, so we should add the top level `subsets/`, `mappings/`, and `patterns/` directory to the `.gitignore` file.

gouttegd:
Top-level release files are all listed in `.gitignore`, because the default behaviour is _not_ to commit them when the repository is initially seeded (unless the `--commit-artefacts` option is passed to the `seed` command).

For consistency, we should also ignore "additional" release files such as subsets, mapping sets, and reports.

closes #1238

### DIFF
--- a/template/.gitignore.jinja2
+++ b/template/.gitignore.jinja2
@@ -23,6 +23,10 @@ bin/
 /{{ project.id }}-full.*
 /{{ project.id }}-simple.*
 /{{ project.id }}-simple-non-classified.*
+/mappings/
+/patterns/
+/reports/
+/subsets/
 
 src/ontology/mirror
 src/ontology/mirror/*


matentzn (approved):
Yes. as long as they are still committed if I want to during the seeding process, this is right!

PR #1247, opened 2025-05-20
Fix how seeding process deals with release artefacts.

### LINKED ISSUE #1245: Seeding may fail when `--commit-artefacts` is used
Example with the `test-robot-report.yaml` sample configuration:

```sh
odkrun seed --commit-artefacts -C test-robot-report.yaml
[…]
INFO:root:RUNNING: cd target/rrc/src/ontology && for asset in $(make show_release_assets) ; do git add -f ../../$asset ; done
INFO:root:OUT: 
ERROR:root:fatal: pathspec '../../reports/rrc-edit.owl-obo-report.tsv' did not match any files

Traceback (most recent call last):
  File "/tools/odk.py", line 1522, in <module>
    cli()
  File "/usr/local/lib/python3.12/dist-packages/click/core.py", line 1161, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/click/core.py", line 1082, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/click/core.py", line 1697, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/click/core.py", line 1443, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/click/core.py", line 788, in invoke
    return __callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tools/odk.py", line 1483, in seed
    runcmd("cd {dir}/src/ontology "
  File "/tools/odk.py", line 1516, in runcmd
    raise Exception('Failed: {}'.format(cmd))
Exception: Failed: cd target/rrc/src/ontology && for asset in $(make show_release_assets) ; do git add -f ../../$asset ; done
```

This is because when attempting to commit the release artefact, the seeding script expects the `reports/rrc-edit.owl-obo-report.tsv` to be in the `reports/` subdirectory of the top-level (release) directory, but just before the script copied that file in the release directory instead (not within a `reports/` subdirectory).

gouttegd:
There are several problems with the way the seeding process (1) copies the release artefacts from the `src/ontology` directory to the top-level release directory and (2) optionally commits those files.

For (1), the problem is that the seeding script manually copies the files itself, and do so by assuming too many things about those files. Notably, it assumes they all go straight to the top-level directory, whereas some should actually go a subdirectory (for example, report files, when they are released, go into `$(RELEASEDIR)/reports`, not into `$(RELEASEDIR)` directly; so do subset files).

The problem with (2) is a consequence of (1): if not all release files have been correctly copied to their correct expected location in the top-level directory, then the seeding script will try to `git add` files that may not actually exist.

We fix (1) by adding a new helper target, `copy_release_files`, to take care of copying all release files to the release directory. It is called internally by the `prepare_release` target, and also by the seeding script. This ensures that the same logic to copy release files will always be used.

Wr fix (2) by updating the `show_release_assets` target (whose sole real purpose is to be used by the seeding script) to list the _actual_ release files: that is, the files that are in the release directory, rather than the files under `src/`.

closes #1245


template/src/ontology/Makefile.jinja2
```diff
@@ -426,22 +431,41 @@ KEEPRELATIONS=keeprelations.txt
...
 # It will ensure that all assets/files are fresh, and will copy to release folder
-
 .PHONY: prepare_release
-prepare_release: all_odk
-	rsync -R $(RELEASE_ASSETS) $(RELEASEDIR) &&\{% if project.sssom_mappingset_group is defined %}{% if project.sssom_mappingset_group.released_products is defined %}
-	mkdir -p $(RELEASEDIR)/mappings && cp -rf $(RELEASED_MAPPING_FILES) $(RELEASEDIR)/mappings &&\{% endif %}{% endif %}{% if project.use_dosdps %}
-	mkdir -p $(RELEASEDIR)/patterns && cp -rf $(PATTERN_RELEASE_FILES) $(RELEASEDIR)/patterns &&\{% endif %}
-	rm -f $(CLEANFILES) &&\
-	echo "Release files are now in $(RELEASEDIR) - now you should commit, push and make a release \
+prepare_release: all_odk copy_release_files
```
matentzn, line 438: Minor nitpick: since copy_release_files does not depend on anything, a make processor that parallelises goal execution would end up copying files that have not been updated yet. I am not too bothered, as I am sure that problem exists of 100 more things in the makefile, but I personally would have done:

```
prepare_release: all_odk
    $(MAKE) copy_release_files
```

No objections though to leave as is.

matentzn (dismissed):

template/src/ontology/Makefile.jinja2
gouttegd: > a make processor that parallelises goal execution would end up copying files that have not been updated yet

😱 This is a _very_ good point and definitely not a “minor nitpick”. I don’t think we officially support parallel builds (I know _I_ never tried one), and I agree with your assessment that there would probably be many problems with them). Still, that doesn’t mean that we should deliberately accept something that gets in the way of parallel builds.

> personally I would have done:
> ```
> prepare_release: all_odk
>    $(MAKE) copy_release_files 
> ```

Agreed!

PR #1250, opened 2025-05-21
Do not allow users to configure ODK directories.

### LINKED ISSUE #1244: Directories for imports, subsets, mappings and the likes should NOT be configurable
Theoretically, the ODK allows users to specify their own directories for several pieces of their repository:

* subsets (option `subset_group.directory`);
* imports (option `import_group.directory`);
* components (option `components.directory`);
* patterns (option `pattern_pipelines_group.directory`);
* mappings (option `sssom_mappingset_group.directory`)
* translations (option `babelon_translation_group.directory`).

Of those, only the options for components, mappings, and translations are actually respected (in that the corresponding `XXXDIR` variable in the generated Makefile will be set to the actual value of the option, if set in the configuration file). The directories for subsets, imports, and patterns are hardcoded.

I think those options should not exist at all. As we have explicitly written in the 2022 paper, the ODK favours a “convention-over-configuration” approach. For better or for worse, it is the ODK convention that subsets should be in `src/ontology/subsets`, imports should be in `src/ontology/imports`, components should be in `src/ontology/components`, patterns should be in `src/patterns`, mappings should be in `src/mappings`, and translations should be in `src/translations`. This should _not_ be user-configurable.

Allowing users to change those conventional directories is bound to create room for horrible snafus, like for example if an user decides that imports should go in `src/imports` (one directory _above_ `src/ontology`) or other funny things like that.

Sure, we could always say, “if you change the default directories, then we make no guarantee, your pipelines may break“, but then why allow changing the default directories at all?

gouttegd:
Directories for subsets, imports, components, patterns, mappings, and translations should _not_ be modified by users, and the ODK configuration file should therefore _not_ offer options to do precisely that.

We remove all `directory` options from the ODK configuration model.

closes #1244

### DIFF
--- a/odk/odk.py
+++ b/odk/odk.py
@@ -312,9 +312,6 @@ class SubsetGroup(ProductGroup):
     products : Optional[List[SubsetProduct]] = None
     """all subset products"""
     
-    directory : Directory = "subsets/"
-    """directory where subsets are placed after extraction from ontology"""
-    
     def _add_stub(self, id : OntologyHandle):
         if self.products is None:
             self.products = []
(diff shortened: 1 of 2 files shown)


matentzn (approved):
Absolutely!

PR #1251, opened 2025-05-21
Update generated documentation for components.

### LINKED ISSUE #1226: Documentation for components is outdated.
Hello - the documentation that comes in a newly seeded ontology has some outdated steps in `components.md`. In particular, it needs an new step to tell the user to add the template before doing `make components/your-component-name.owl`, and it also has a superfluous step - currently Step 6 which instructs the user to add a goal to the custom makefile. This appears to be unnecessary.

I have an updated version on my repo here: https://timalamenciak.github.io/elmo/odk-workflows/components/

Happy to make PR but I couldn't track down where in the ODK code this MD file lives because I think it is generated by the seed repo process.

gouttegd:
Update the generated `docs/odk-workflows/components.md` file to:

* remove the bits about having to manually update the -edit file and the catalog file, since this is now done automatically by the ODK;
* remove the bit about having to create a custom rule to generate a component derived from a template, and explain instead how the ODK can be configured to generate such a rule automatically (`use_template` option);
* also show how the ODK can be configured to derive components from mappings or to fetch a remote component.

closes #1226


matentzn (approved):
LGTM

PR #1253, opened 2025-05-23
Remove code of conduct from the template.

### LINKED ISSUE #1228: Missing markdown file, `contacts.md`, in link
Thank you for all your useful work in ontologies and for providing this repo to the scientific and engineering community.

There is a minor issue in the file, "[CODE_OF_CONDUCT.md](https://github.com/INCATools/ontology-development-kit/blob/master/template/CODE_OF_CONDUCT.md)," while clicking on the link, "[contacting the project team](https://github.com/INCATools/ontology-development-kit/blob/master/template/contact.md)" -- the file to which this link connects, "contact.md," is "404 - page not found"

Thank you again for the time you spend on this work and the fabulous results.

gouttegd:
An automatically generated code of conduct is of little value without any indication that it is in fact endorsed by the project, instead of having simply been left there because the project owner may not even have realised the ODK generated it.

Project owner who wants a CoC should take the necessary steps to add one themselves – if they truly care about the values that a CoC is intended to convey, they should not mind having to do that. If they say, “I want a CoC but I don’t want to do anything to get one”, then they don’t care about a CoC, they care about being seen as having a CoC – in other words, it’s virtue signalling, and it’s arguably worse than not having a CoC.

closes #1228


matentzn (approved):
Its a bit sad but ultimately you are right. Having dozens of COCs out there with no one having explicitly supported them for a reason is not helpful.

PR #1259, opened 2025-05-26
Preserve non-ODK-managed imports by default.

gouttegd:
When updating a repository and updating the import declarations in the edit file, by default do _not_ call the `odk:import` command in “exclusive” mode, so that import declarations that do not correspond to an ODK-declared import or component are preserved.


matentzn (approved):
Thank you..

PR #1264, opened 2025-05-26
Make `dicer-cli` more helpful in case of an invalid policy file.

### LINKED ISSUE #1262: Weaving validate_idranges into test will cause a lot of community support requests
ID ranges files are hardly ever used computationally - they are more like a glorified TXT file to document who can add what term ids.

I have already fixed two broken id range files with the two ontologies I have tested :P 

Not a problem, but we need to communicate this clearly in the release notes, as no one really know how to _fix_ id ranges files, because the validator does not tell us how. 

Idea: instead of using dicer, why not.. load the id ranges file into ROBOT instead? At least you will get a proper error message? :P Not sure.

```
if [ -f hp-idranges.owl ]; then dicer-cli policy hp-idranges.owl ; fi
dicer-cli: Cannot read policy file: Cannot load ID range policy
```

gouttegd:
This PR upgrades Dicer-CLI to version 2.1 and uses the new options available in that version to report a somewhat helpful error message from the OWLAPI in case the policy cannot be loaded.

closes #1262


matentzn (approved):
Excellent!

PR #1269, opened 2025-05-27
Add dynamic branch names to github actions

matentzn:
This is so that the two actions do not overwrite each others branches.

@gouttegd not entirely sure this will work until we test it.

### DIFF
--- a/.github/workflows/constraints.yml
+++ b/.github/workflows/constraints.yml
@@ -22,6 +22,7 @@ jobs:
         uses: peter-evans/create-pull-request@v3
         if: ${{ success() }}
         with:
+          branch: update-obo-epm-${{ github.run_id }}
           token: ${{ secrets.WORKFLOW_TOKEN }}
           commit-message: Update constraints.txt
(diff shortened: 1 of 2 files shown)


.github/workflows/constraints.yml
```diff
@@ -22,6 +22,7 @@ jobs:
         uses: peter-evans/create-pull-request@v3
         if: ${{ success() }}
         with:
+          branch: update-obo-epm-${{ github.run_id }}
```
gouttegd, line 25:
```suggestion
          branch: update-constraints-${{ github.run_id }}
```

gouttegd: Otherwise I don’t see the point of the PR! :D

gouttegd (changes requested):

.github/workflows/constraints.yml
matentzn: Actually the run_id was not even necessary - this way it does not update the existing PR, which it should, when running again.

PR #1275, opened 2025-05-28
Allow disabling automatic management of import declarations

gouttegd:
This PR adds a new project-level option in the ODK configuration schema to control whether the ODK should manage the import declarations in the `-edit` file and the redirections in the XML catalog.

When enabled (it is by default), users should never have to manually touch the import declarations or the redirections (though in the case of the redirections, they _can_ insert additional redirections if they want; they will not be removed as long as they are not in the `odk-managed-catalog` group).
    
If disabled, then it will be the same as it was before: the ODK will not touch the import declarations and the redirections, it will be up to the users to ensure they are correctly updated when needed.

The option is honoured both when a repository is updated and when it is seeded for the first time. That is, if `manage_import_declarations: false` is already set in the initial configuration file at seeding, then the newly created `-edit` and catalog files will _not_ contain automatically generated import declarations and redirections – from the beginning, it will be up to the users to add them.

### DIFF
--- a/template/_dynamic_files.jinja2
+++ b/template/_dynamic_files.jinja2
@@ -22,19 +22,21 @@
 ^^^ src/ontology/{{ project.id }}-edit.{{ project.edit_format }}
 {% if project.edit_format == "obo" -%}
 format-version: 1.2
-{%   if project.import_group is defined -%}
-{%     if project.import_group.use_base_merging -%}
+{%   if project.manage_import_declarations -%}
+{%     if project.import_group is defined -%}
+{%       if project.import_group.use_base_merging -%}
 import: {{ project.uribase }}/{% if project.uribase_suffix is not none %}{{ project.uribase_suffix }}{% else %}{{ project.id }}{% endif %}/imports/merged_import.owl
-{%     else -%}
-{%       for imp in project.import_group.products -%}
+{%       else -%}
+{%         for imp in project.import_group.products -%}
 import: {{ project.uribase }}/{% if project.uribase_suffix is not none %}{{ project.uribase_suffix }}{% else %}{{ project.id }}{% endif %}/imports/{{ imp.id }}_import.owl
-{%       endfor -%}
+{%         endfor -%}
+{%       endif -%}
 {%     endif -%}
-{%   endif -%}
-{%   if project.use_dosdps -%}
+{%     if project.use_dosdps -%}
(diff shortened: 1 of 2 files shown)


odk/odk.py
```diff
@@ -697,8 +697,8 @@ class OntologyProject(JsonSchemaMixin):
     use_custom_import_module : bool = False
     """if true add a custom import module which is managed through a robot template. This can also be used to manage your module seed."""
 
-    preserve_non_odk_managed_imports : bool = True
```
matentzn, line 700: I just want to bounce one thought about removing this option:

For the catalog we are good:

> When enabled (it is by default), users should never have to manually touch the import declarations or the redirections (though in the case of the redirections, they can insert additional redirections if they want; they will not be removed as long as they are not in the odk-managed-catalog group).

Perfect.

For the edit file, removing `preserve_non_odk_managed_imports` as an option is probably the right thing to do, but it does at least not allow adding a random imports statement along with the ODK managed ones; I cant see a use case for this right now, but I know that some _non_ ODK ontologies still use a handful of direct (URL, un-catalogued) imports.

In any case, I think this is the right thing to do. If people want to use this feature _while_ having random imports in their edit file they can file a feature request and we just add that option in a future release, but I doubt this will be requested.

matentzn (approved):
Looks great! THANK YOU for having the patience to do this! Much appreciated!

odk/odk.py
gouttegd: > it does at least not allow adding a random imports statement along with the ODK managed ones;

Yes. It is all or nothing: either you let the ODK manage the import declarations, or you do it yourself.

Knowing that the “do-it-yourself” has always been the normal behaviour (since until now the ODK did not offer anything else), I see very little point in having yet another option to allow people to say both (1) please manage my imports (`manage_import_declarations: true`) and (2) please let me do my own things at the same time (`preserve_non_odk_managed_imports: true`).

> I know that some non ODK ontologies still use a handful of direct imports

I do not doubt that. I also know of some very ODK-managed ontologies that also have additional imports on top of those declared in the ODK. But again, in those cases the ontology owners will just have to set `manage_import_declarations: false` and keep managing the imports themselves.

gouttegd: > If people want to use this feature while having random imports in their edit file they can file a feature request and we just add that option in a future release

If this is indeed a request from users, then I’d have no objection.

Issue #1286, opened 2025-08-06
Make it possible to change the configuration schema transparently for existing users


gouttegd:
The schema for the ODK configuration file is a bit of a mess.

In particular, the top-level (“project”) dictionary contains a lot of options about wildly diverse stuff ranging from enabling validation of RDF/XML artefacts to setting the amount of memory allocated to ROBOT to setting the emails to which Travis should send failure reports, etc. That schema would benefit from some reorganisation to make things clearer, e.g. grouping all options related to the upstream repository in a `repository` section,  grouping all options related to ROBOT in a `robot` section, etc.

That is, instead of

```yaml
title: "Cell Ontology"
github_org: obophenotype
git_main_branch: master
repo: uberon
obo_format_options: --clean-obo 'simple merge-comments'
robot_plugins:
  plugins:
    - name: flybase
      mirror_from: https://github.com/FlyBase/flybase-robot-plugin/releases/download/flybase-robot-plugin-0.2.2/flybase.jar
    - name: uberon
      mirror_from: https://github.com/obophenotype/uberon-robot-plugin/releases/download/uberon-robot-plugin-0.4.0/uberon.jar
reasoner: ELK
robot_relax_options: --include-subclass-of true
allow_equivalents: asserted-only
robot_report:
  release_reports: False
  fail_on : ERROR
  use_labels : False
  custom_profile : False
  report_on :
    - edit
...
```

we could have something like:

```yaml
title: "Cell Ontology"
repository:
  name: cl
  organisation: obophenotype
  main_branch_name: master
robot:
  obo_format_options: --clean-obo 'simple merge-comments'
  relax_options: --include-subclass-of true
  reasoner: ELK
  allow_equivalents: asserted-only
  plugins:
    - name: flybase
      mirror_from: https://github.com/FlyBase/flybase-robot-plugin/releases/download/flybase-robot-plugin-0.2.2/flybase.jar
    - name: uberon
      mirror_from: https://github.com/obophenotype/uberon-robot-plugin/releases/download/uberon-robot-plugin-0.4.0/uberon.jar
  report:
    release_reports: False
    fail_on : ERROR
    use_labels : False
    custom_profile : False
    report_on :
      - edit
...
```

[...]

PR #1287, opened 2025-08-06
Allow changing the configuration schema transparently for users

gouttegd:
This PR implements the system envisioned in #1286 to allow us (ODK developers) to reorganise the ODK configuration schema to (hopefully) make it clearer, all the while allowing users to keep using their existing configuration files.

The first commit implements the actual configuration update mechanism (the `update_config_dict` method in `odk/odk.py`).

The second commit is a proof-of-concept in which we regroup some of the ROBOT-related options, that are currently dispersed in the top-level dictionary, into a new `robot` section.

### DIFF
--- a/odk/odk.py
+++ b/odk/odk.py
@@ -913,6 +913,7 @@ class Generator(object):
                 config_hash = h.hexdigest()
                 stream.seek(0)
                 obj = yaml.load(stream, Loader=yaml.FullLoader)
+            update_config_dict(obj)
             project = from_dict(data_class=OntologyProject, data=obj)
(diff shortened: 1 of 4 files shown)


matentzn (approved):
I went through this line by line and it looks lovely. 

Right now I cant test anything, but let that not block you, I agree

1. with the need of making the config schema more sane
2. the gracious backwards compatible approach to handle this

Very nice!

matentzn:
> The changes in that commit (grouping several ROBOT options into a single robot section) are probably good 

They are great!

PR #1297, opened 2025-09-30
Always use GitHub-CLI for public release to GitHub.

### LINKED ISSUE #1296: Remove the `github_curl` and `github_python` public release methods
The ODK currently offers three methods of doing a public release to GitHub:

* through a custom Python script (`make-release-assets.py`) based on the `pygithub` package (`public_release: github_python`);
* through curl-based direct calls to GitHub’s API (`public_release: github_curl`);
* through GitHub’s own command-line tool (`gh`) – this is the default if `public_release` is set to anything else than `github_python` or `github_curl`).

I don’t see the point of keeping (i.e. maintaining) the first two methods while we have the third one.

I propose ① moving `gh` from ODKFull to ODKLite, and ② always use `gh` whenever doing a public release to GitHub.

gouttegd:
This PR removes the `github_curl` and `github_python` methods of pushing a release to GitHub – these should no longer be needed now that GitHub-CLI is available.

The `public_release` ODK option should now be set to the name of the hosting service for releases. For now, the only supported option is `github` (which, in effect, is exactly as before). The older values `github_curl` and `github_python` are silently transformed into `github`.

To accommodate for that change:

(1) GitHub-CLI (`gh`) is moved from ODKFull to ODKLite;

(2) conversely, PyGithub (which was only used for the `github_python` method) is moved from ODKLite to ODKFull (it is not completely *removed*, since people may have custom workflows that depend on it).

closes #1296


matentzn (approved):
Wonderful removal of tech debt

PR #1338, opened 2026-04-28
Update the base image to Ubuntu 26.04.

### LINKED ISSUE #1320: Update base system to Ubuntu 26.04 “Resolute Raccoon”
It’s an even year, meaning it will be time to update the base system of the ODK images.

Nothing to do immediately as Ubuntu 26.04 is not released yet (it is [scheduled](https://discourse.ubuntu.com/t/resolute-raccoon-release-schedule/47198) to be released on 23 April), but this issue is here to remind us of this update.

Will come way too late for the upcoming ODK 1.6.1, but should definitely be part of ODK 1.7.

gouttegd:
Use the new Ubuntu 26.04 as the base image for:

* computing the new Python constraints;
* building the `builder` image;
* building the `odklite` image.

closes #1320


gouttegd:
Draft PR as (1) I have not tested it locally yet (and I would be surprised if it works without any tweaks) and (2) I have not updated the Python constraints yet.

In both cases I won't do that until I have better Internet connectivity than what is available to me right now.

gouttegd:
> I would be surprised if it works without any tweaks

… And indeed it failed before I even finished writing this comment. Fully expected: the new base image comes with a different version of setuptools, so the `pip-constraints.txt` file needs to be updated to reflect what the new base image provides.

constraints.txt
```diff
@@ -34,301 +33,292 @@ comm==0.2.3
...
-keyring==25.6.0
+jupyterlab_server==2.28.0
+jupyterlab_widgets==3.0.16
+keyring==25.7.0
 kgcl-rdflib==0.5.0
 kgcl_schema==0.6.9
-kgx==2.6.0
-kiwisolver==1.4.9
-lark==1.3.0
+kgx==2.3.2
```
matentzn, line 137: While we dont track kgx as core, it is quite an important backage, and that downgrade is concerning

constraints.txt
```diff
@@ -34,301 +33,292 @@ comm==0.2.3
...
-narwhals==2.7.0
-nbclient==0.10.2
-nbconvert==7.16.6
+narwhals==2.20.0
+nbclient==0.10.4
+nbconvert==7.17.1
 nbformat==5.10.4
 ndex2==3.11.0
-neo4j==6.0.2
+neo4j==4.4.13
```
matentzn, line 173: I think that downgrade is too much..

constraints.txt
```diff
@@ -34,301 +33,292 @@ comm==0.2.3
...
-openai==2.1.0
+ontoportal_client==0.0.9
+openai==2.33.0
 openpyxl==3.1.5
 ordered-set==4.1.0
-packaging==25.0
+packaging==26.0
 paginate==0.5.7
-pandas==2.3.3
+pandas==3.0.2
```
matentzn, line 192: hui.. Thats.. scary! But why not. @cthoyt?

constraints.txt
```diff
@@ -34,301 +33,292 @@ comm==0.2.3
...
-kgx==2.6.0
-kiwisolver==1.4.9
-lark==1.3.0
+kgx==2.3.2
+kiwisolver==1.5.0
+lark==1.3.1
+librt==0.9.0
 lightrdf==0.4.0
-linkml==1.9.4
-linkml-map==0.3.8
```
matentzn, line 145: I depend on this, so I'd have to add this back to the core list.

constraints.txt
```diff
@@ -34,301 +33,292 @@ comm==0.2.3
...
+keyring==25.7.0
 kgcl-rdflib==0.5.0
 kgcl_schema==0.6.9
-kgx==2.6.0
-kiwisolver==1.4.9
-lark==1.3.0
+kgx==2.3.2
+kiwisolver==1.5.0
+lark==1.3.1
+librt==0.9.0
```
matentzn, line 140: never heard of this

matentzn (commented):
All looks great. The typical concerns on the python front - nothing new. Please feel free to override me on all counts, I dont feel super strongly and none of it is related to the base image migration.

constraints.txt
gouttegd: Given that `kgx` is a serial offender when it comes to “causing the dependency solver to fail by needlessly and excessively punning its dependency with overly conservative version constraints”, I am not keen on going to any great length for its sake. Its developers can sort their own mess.

constraints.txt
gouttegd: Feel free to spend time investigating why the dependency solver picked that version (i.e., which Python package(s) again caused it by excessively conservative version constraints).

I am not going to do it unless I have a precise bug report telling us that a workflow is broken by that downgrade.

constraints.txt
gouttegd: Not core. Core (which contains everything that ends up in ODK-Lite) is for what is needed by the _standard_ ODK workflows. I am pretty sure we have no standard workflows that need LinkML-Map.

If you need LinkML-Map for non-standard workflows, that’s perfectly fine, but then explicitly add `linkml-map` to the requirements for **ODKFull**, not **ODKLite**.

matentzn: Of course

constraints.txt
matentzn: Sure! I dont really care. resolving

matentzn (approved):
Looks great!

PR #1341, opened 2026-05-02
Add compatibility wrappers for pre-1.7 repositories.

### LINKED ISSUE #1339: ODK 1.7 will need compatibility wrappers for pre-1.7 repositories
Among other things, the new ODK-Core project has replaced a few scripts in the ODK image by a single consolidated script called `odk-helper`.

For example, the `check-rdfxml` script (used to check RDF/XML artefacts for correctness) has been replaced by a `odk-helper check-rdfxml` subcommand.

The new consolidated script will be easier to maintain, but one consequence is that a Makefile generated with a pre-1.7 image will _not_ work with a 1.7+ image, because the Makefile will try to run the old scripts that are no longer available in the new image.

This will instantly break, for example, all workflows that use the `obolibrary/odkfull:latest` image, the moment the ODK 1.7 image is released.

This is not acceptable, so we'll need to provide wrapper scripts so that pre-1.7 Makefile can still work with the 1.7 image. At least for ODK 1.7, and then we may remove the wrappers in ODK 1.8.

gouttegd:
ODK versions up to 1.6 had included two small helper scripts:

* check-rdfxml, to check RDF/XML artefacts for correctness;
* context2csv, to convert a JSON-based context file to a CSV file.

In ODK 1.7, as part of the migration to the ODK-Core system, those scripts were consolidated into a single `odk-helper` script. Makefiles generated with ODK 1.7 will automatically call this new `odk-helper` script.

But of course, Makefiles that were generated with a pre-1.7 ODK will still attempt to call the original individual scripts. To make sure that standard workflows from a pre-1.7 ODK instance can still be run with a ODK 1.7 image, we must provide wrapper scripts, which is what we do here.

Similarly, we must also provide symlinks linking the old locations of the ROBOT plugins and of the templates to their new respective locations.

closes #1339

### DIFF
--- a/docker/odklite/Dockerfile
+++ b/docker/odklite/Dockerfile
@@ -126,6 +126,10 @@ RUN mkdir /tools && ln -s /usr/local/bin/odk /tools/odk.py && \
 # Install the OBO extended prefix map
 COPY resources/obo.epm.json /odk/resources
 
+# Install wrapper scripts for compatibility with pre-1.7 Makefiles
+COPY --chmod=755 scripts/compat-check-rdfxml.sh /odk/bin/check-rdfxml
+COPY --chmod=755 scripts/compat-context2csv.sh /odk/bin/context2csv
+
 # Install anything that may be missing from ODK-Core
 RUN odk install /odk
 
--- a/docker/odklite/Dockerfile
+++ b/docker/odklite/Dockerfile
@@ -116,22 +116,29 @@ RUN sed -i s/@@ODK_IMAGE_VERSION@@/$ODK_VERSION/ /odk/bin/odk-info
 # Install a helper script to launch a repository update
 COPY --chmod=755 scripts/update_repo.sh /odk/bin/update_repo
 
-# The odk script used to be named `odk.py` and to be located at
-# /tools/odk.py, and both the name (with the `.py` extension) and the
-# full path have been hardcoded at several places, so we need some
-# symlinks for compatibility.
-RUN mkdir /tools && ln -s /usr/local/bin/odk /tools/odk.py && \
-    ln -s /usr/local/bin/odk /odk/bin/odk.py
-
 # Install the OBO extended prefix map
 COPY resources/obo.epm.json /odk/resources
 
-# Install wrapper scripts for compatibility with pre-1.7 Makefiles
-COPY --chmod=755 scripts/compat-check-rdfxml.sh /odk/bin/check-rdfxml
-COPY --chmod=755 scripts/compat-context2csv.sh /odk/bin/context2csv
-
 # Install anything that may be missing from ODK-Core
 RUN odk install /odk
+
+# Compatibility layer for work with pre-1.7 repositories.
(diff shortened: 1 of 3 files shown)


docker/odklite/Dockerfile
```diff
@@ -116,18 +116,29 @@ RUN sed -i s/@@ODK_IMAGE_VERSION@@/$ODK_VERSION/ /odk/bin/odk-info
...
+#     changed in 1.7 but which have been hardcoded in several places in
+#     pre-1.7 repos:
+#     (a) the `odk` script;
+#     (b) the directory containing the ROBOT plugins;
+#     (c) the directory containing the templates.
+RUN mkdir -p /tools && \
+    ln -s /usr/local/bin/odk /tools/odk.py && \
+    ln -s /usr/local/bin/odk /odk/bin/odk.py && \
+    ln -s /odk/resources/robot/plugins /tools/robot-plugins && \
+    ln -s /usr/local/lib/python3.12/dist-packages/incatools/odk/templates /tools/templates
```
matentzn, line 136: Note you hardcoding the python version in the symlink is fine for now but creates a maintenance overhead.

scripts/compat-context2csv.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+cat > $$.tmp
+odk-helper context2csv $$.tmp
+rm $$.tmp
```
matentzn, line 5: Double check if this will absorb the exit code thrown by context2csv (I forgot what context2csv, maybe exit codes are not so relevant, just mentioning here for your consideration).

matentzn (commented):
Looks great, two minor considerations for yourself.

docker/odklite/Dockerfile
gouttegd: It’s a good point, but:

(1) There’s not much we can do about that. We did not choose to put the Python packages into a directory that includes the Python version number, that’s what Python always does.

(2) We will only need to change that whenever we bump the Python version, which only happens when we upgrade the base image, so once every two years.

And I do not plan to keep this compatibility layer indefinitely. At some point in the future, we will say, “pre-1.7 repositories will no longer work with ODK 1.X, you will need to either run `update_repo` or make sure you don’t use `odkfull:latest` in your workflows”. Not sure yet what “1.X” will be (and happy to discuss it), but probably 1.9 or 1.10.

matentzn: Yes. folks can always use the OLD ODK containers and workflows so I am not concerned. We wont break anything in running systems, just backwards compatibility for those that want to upgrade!

scripts/compat-context2csv.sh
gouttegd: The only `context2csv` could error out is if the JSON context file is bogus – if that was the case, I’d expect many other things to go wrong whenever a workflow is run, since we now feed the context to all ROBOT invocations.

But still, it’s a good point. The wrapper script should return the same exit code as the command it is wrapping.

matentzn (approved):
great!

PR #1342, opened 2026-05-03
Update ODK-Core to version 0.1.4.

gouttegd:
This PR updates the version of ODK-Core that is bundled with the ODK image to version 0.1.4, containing all the changes brought to ODK-Core since early January.


matentzn (dismissed):
If you would be soooo kind, can we redo to capture the latest sssom-py which would avoid me having to build workaround for a nasty "split dataframes by prefix bug"? if not, I am fine with this 100%!

gouttegd:
You do understand that the version of `sssom-py` in this PR is _not_ necessarily the one that will end up in ODK 1.7, right?

We are still far from ODK 1.7, so it is certain that we will update the Python constraints at least one more time before we get there.

matentzn (approved):
I am sure there will be at least two minor releases on the way :) Much appreciated!

PR #1344, opened 2026-05-05
Clarify comments inside `python-requirements.txt`.

### LINKED ISSUE #1340: Revise and simplify `python-requirements.txt`
This is our current `python-requirements.txt` file:

```
# Core dependencies
./core[workflows]
linkml
oaklib
semsimian

# Secondary dependencies
PyGithub
curies
funowl
jsonschema2md
jupyter
kgx
matplotlib
matplotlib_venn
networkx
numpy
ontobio
ontodev-cogs
ontodev-gizmos
pandas
plotly
pyspellchecker
rdflib
requests
seaborn
tabulate
upsetplot
jinjanator
```

It has a few problems:

(1) The comments are misleading. The only “core” dependency is `core[workflows]` – all Python packages that should be part of the core (i.e. part of ODKLite) MUST be explicitly listed as dependencies within the `core` package, so that `core` is the _only_ Python module that we should have to install when building ODKLite. If I recall correctly (@matentzn you can correct me if I’m wrong here), what was meant with those comments is that the “core dependencies” are those for which we always want the very latest version (and thus, for which we should pay close attention when updating the constraints file, to check that they are not downgraded), whereas the “secondary dependencies” are those for which we don’t really care whether we got the last version or not.

(2) Some of those dependencies explicitly listed here are _already_ dependencies (direct or indirect) of the `core` package (e.g. `linkml`, `oaklib`, `pandas`) and therefore _could_ be removed here (not strictly necessary of course, but could maybe avoid confusion).

(3) Some Python packages that we actually want are not explicitly listed, e.g. `linkml-map` (which is currently included as a result of some dependency chains, but we should not rely on that).

gouttegd:
The comments within the `python-requirements.txt` file were misleading, as they were suggesting that the file listed both packages needed for ODK-Lite ("core dependencies") and for ODK-Full ("secondary dependencies").

This PR clarifies the contents of the python-requirements.txt by breaking it down in three sections:

(A) The section with the dependencies needed for ODK-Lite -- which, by definition, only contain the ODK-Core package.

(B) The section containing all the Python packages that should be, as much as possible, always kept to their most recent versions ("first-tier dependencies").

(C) The section containing all the other packages wanted for ODK-Full.

This commit also adds to the (C) section the `linkml-map` package. For now, that package is already included in ODK-Full because it is an indirect dependency of something else, but that may change (as we observed when trying to update the base image to Ubunty 26.04, where `linkml-map` is no longer included), so we make `linkml-map` an _explicit_ dependency.

closes #1340

### DIFF
--- a/python-requirements.txt
+++ b/python-requirements.txt
@@ -1,16 +1,40 @@
-# Core dependencies
+# Core dependency
+# This is the ONLY dependency that should be required for ODK-Lite. If
+# you want to add a Python package to ODK-Lite (which should only happen
+# if said package is required by a standard ODK workflow), do NOT add it
+# here, but instead add it to the "workflows" dependency group in the
+# ODK-Core project.
 ./core[workflows]
+
+# First-tier dependencies
+# The following dependencies are those for which we should make sure to
+# always provide the most recent version, unless we have a compelling
+# reason not to do so (e.g. if the latest version of a package is known
+# to break a workflow).
+# This list MAY include dependencies that are already listed in the
+# dependencies of the ODK-Core package.
 linkml
 oaklib
 semsimian
 
-# Secondary dependencies
-PyGithub
+# Second-tier dependencies
+# Those are all the other Python packages that should be provided in the
+# ODK-Full image, but for which we do not necessarily always try to
+# provide the most recent version.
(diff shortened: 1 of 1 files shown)


matentzn (approved):
Awesome

PR #1347, opened 2026-05-20
Upgrade DOSDP Tools to 0.20

matentzn:
Still testing this.


gouttegd:
Note that, with the switch to ODK-Core, the test suite in this repo only runs a small fraction of the existing tests.

If you want to run the full test suite with DOSDP Tools 0.20, you must add it to ODK-Core ([here](https://github.com/INCATools/odkcore/blob/3a20e7a609a12c4205b99f2d9bdcf0687b6f421a/src/incatools/odk/setup.py#L25)), and then run ODK-Core’s test suite (either locally, or on GitHub after creating a PR).

matentzn:
> Note that, with the switch to ODK-Core, the test suite in this repo only runs a small fraction of the existing tests.

I didnt realise there were ANY tests for DOSDP tools? I have already tested it with 4 repos (OBA, Mondo, ECTO, DPO) - so I am not too concerned. But a single place to maintain the DOSDP version - which is it?

gouttegd:
> I didnt realise there were ANY tests for DOSDP tools?

I think there are a couple of test configs that do make use of DOSDP. But of course, as for all our tests, we merely check that we can generate the repository and run a `make_release` workflow – we do not actually check that the tools behave as they should.

> But a single place to maintain the DOSDP version - which is it?

The way things work for now is that ODK-Core can install the tools itself, _but_ the build process for the ODK images will override ODK-Core.

So there are _two places_ where the version of DOSDP-Tools can be specified, but the one place that gives the version that will end up in the actual ODK image is the one that you modified in this repository.

gouttegd (approved):
Preemptive approval. @matentzn Feel free to merge whenever you are happy that you’ve done enough testing.

Clare72:
Looks good from testing dpo and FBbt

aleixpuigb:
Looks good from testing in OBA

gouttegd (approved):

Issue #1354, opened 2026-07-14
[bot] Bump bundled oaklib to 0.7.2 so sqlite:obo CI downloads use the semanticsql CDN


kltm:
[bot] Opened by a Claude Code agent on behalf of @kltm.

## Problem

The ODK image (`obolibrary/odkfull` / `odklite`) is the shared runtime for OBO ontology CI, and it bundles OAK. `constraints.txt` currently pins `oaklib==0.6.23`, which resolves `sqlite:obo:<onto>` databases by downloading the raw bucket URL `https://s3.amazonaws.com/bbop-sqlite/<onto>.db.gz` via Python's stdlib `urllib`, with **no HTTP conditional request** — so every ODK-based ontology repo re-downloads full `.db.gz` files from raw S3 on each CI run.

OAK **v0.7.2** (released 2026-07-09, via INCATools/ontology-access-kit#897 / #898) repoints the default download to the CDN **https://semanticsql.berkeleybop.io** (edge-cached) and uses a non-`urllib` User-Agent (the CDN returns 403 to `Python-urllib`). That fix is released but **stranded behind the ODK pin** — none of the downstream ODK ontology CI benefits from it yet.

## Ask

Bump `constraints.txt` `oaklib==0.6.23` → `oaklib==0.7.2` and rebuild/publish the ODK images, so the entire `sqlite:obo:` CI fan-out repoints to the CDN in a single image release (largest blast radius by far). Compatibility across the 0.6.23 → 0.7.2 jump needs maintainer verification — a draft PR is attached.

Related: INCATools/ontology-access-kit#880 (egress umbrella), #889 (epic), INCATools/semantic-sql#112 (bucket lockdown). Part of the monarch-initiative "Minimize semsql hits" effort.

_— Posted by Claude Code agent on behalf of @kltm._

gouttegd:
Closing as this is _not_ an ODK issue.

That the ODK currently provides OAK 0.6.23 is not an explicit pinning decision on the ODK part, it is merely the consequence of OAK 0.7.2 being not compatible with Python 3.14. Once this will have been fixed on the OAK side, the ODK will automatically pick up the earliest OAK release compatible with Python 3.14.

Issue #897, opened 2023-07-15
Decide how to deal with 8 downregulated dependencies


matentzn:
Discussed on slack, but adding here because I have to make some immediate progress with this. 

Because of the currently wrong way we are dealing with python dependencies in ODK (one mega environment rather than many focussed ones) we had to do some prioritisation of dependencies, leading to 8 transitive dependencies being down-regulated (funowl, kgx, docutils, neo4j, ontobio, Sphinx, tomli, tox, see #894).

- funowl
- kgx (@hrshdhgd working on this)
- docutils
- neo4j
- ontobio (@cmungall says remove, and I agree, so consider this a flag for removal)
- Sphinx
- tomli
- tox

matentzn:
```
Because kgx (2.2.1) depends on networkx (>=2.8,<3.0)
 and no versions of kgx match >2.2.1,<3.0.0, kgx (>=2.2.1,<3.0.0) requires networkx (>=2.8,<3.0).
So, because test depends on both networkx (^3.1) and kgx (^2.2.1), version solving failed.
```

anitacaron:
We have had two other minor releases and have not encountered any issues yet.

matentzn:
We are going to close this because for a few months, versions where only ever upgraded. If a user comes that needs a newer kgx version, or network x, they can make a separate issue.

PR #1355, opened 2026-07-14
[bot] constraints: bump oaklib 0.6.23 -> 0.7.2 (repoint sqlite:obo downloads to the semanticsql CDN)

kltm:
[bot] Opened by a Claude Code agent on behalf of @kltm.

Bumps the bundled `oaklib` pin so ODK-based ontology CI downloads `sqlite:obo:` databases from the edge-cached CDN (`https://semanticsql.berkeleybop.io`) instead of the raw S3 bucket.

OAK v0.7.2 changed the default semsql download base to the CDN and switched off the stdlib `urllib` backend (the CDN returns 403 to `Python-urllib`). ODK currently pins `oaklib==0.6.23` (raw S3 + urllib), so the released fix does not reach any downstream ODK ontology CI until this bump + an image rebuild — the single largest-blast-radius repoint available.

**Please verify** 0.6.23 → 0.7.2 compatibility before merge (that is why this is a one-line proposal for review). `semsql==0.4.0` (line 270) may also need a companion bump once a semsql release carrying the CDN default is cut (INCATools/semantic-sql#114).

Refs INCATools/ontology-development-kit#1354, INCATools/ontology-access-kit#880 / #897.

_— Posted by Claude Code agent on behalf of @kltm._

### DIFF
--- a/constraints.txt
+++ b/constraints.txt
@@ -177,7 +177,7 @@ nh3==0.3.5
 notebook==7.5.5
 notebook_shim==0.2.4
 numpy==2.4.4
-oaklib==0.6.23
+oaklib==0.7.2
 oauthlib==3.3.1
 ols-client==0.2.1
 ontobio==2.9.12


matentzn (changes requested):
I will do this - this is not the right way

gouttegd:
> this is not the right way

It’s not as is the right way was not [documented](https://github.com/INCATools/ontology-development-kit/blob/master/CONTRIBUTING.md#python-constraints), but hey, what else can you expect from a probabilistic bullshit generator…

gouttegd:
> I will do this

Don’t waste your time : `oaklib` 0.7.2 is explicitly _not_ compatible with Python 3.14, which is the Python version used in the ODK since the upgrade to Ubuntu 26.04. This needs to be fixed in OAK first.

matentzn:
> Don’t waste your time : oaklib 0.7.2 is explicitly not compatible with Python 3.14, which is the Python version used in the ODK since the upgrade to Ubuntu 26.04. This needs to be fixed in OAK first.

There are also some other conflicts - for example oaklib and curies.. Anyways this is part of a huge fallout from egress costs incurred by oaks caching of sqlite connectors. So I need to help figure out how to fix it fast..

gouttegd:
> So I need to help figure out how to fix it fast..

I’ve been looking at that for the past hour. I don’t know about the `curies` package, but the issue with `oaklib` is unlikely to be an easy fix.

As far as I can tell, it is ultimately caused by Python 3.14 [introducing a breaking change](https://github.com/python/cpython/issues/141829) that is breaking `pyhornedowl`. Any attempt to load anything from `pyhornedowl` under Python 3.14 results in

```
AttributeError: 'typing.Union' object attribute '__doc__' is read-only
```

My understanding is that the 3.14 change is a problem for `pyhornedowl` because part of the process of generating the Python bindings from the Rust Horned-OWL library is the automatic generation of docstrings.

PyHornedOWL defines at least one type as a `typing.Union`: [IRIParam](https://github.com/ontology-tools/py-horned-owl/blob/8658cf32ad482ab87e21838f3dc1de7fd2f4dffd/src/model.rs#L52). I believe the wrapper generation process automatically tries to assign a docstring to that type, which is no longer allowed under Python 3.14.

I am not even sure this can be fixed in PyHornedOWL itself, it looks like an issue with the PyO3 bindings library.

matentzn:
😓 
I didnt even get to pyhornedowl yet.. 

Do you think it would work if we made pyhornedowl and "extra" dependency in oak?

gouttegd:
> Do you think it would work if we made pyhornedowl an "extra" dependency in oak?

Certainly not. `pyhornedowl` is _unconditionally_ imported by the `oaklib.selector` module (which I believe is one of the most “core” components of OAK). Declaring it as an extra dependency may allow you to _install_ the `oaklib` package, but any attempt to _use_ it (or to run the `runoak` command) would result in a runtime import error due to the missing `pyhornedowl` package.

You would also need to make sure that `pyhornedowl` is _really_ treated as an optional dependency (imported only when present, and with the OAK still working reasonably well when absent).

Aside:

> Anyways this is part of a huge fallout from egress costs incurred by oaks caching of sqlite connectors

Given that

(1) I am pretty sure that no standard ODK workflow makes use of OAK, so this only affects people users with custom workflows;

(2) the ODK does have a mechanism for sharing the host’s OAK cache with the container, precisely to prevent the OAK from constantly redownloading the same databases upon each invocation;

I am sorry but I am quite inclined to consider this “huge fallout” as a not-my-problem issue.

matentzn:
> I am sorry but I am quite inclined to consider this “huge fallout” as a not-my-problem issue.

You should! Leave it to me :) Thanks for checking this helped me a lot!

gouttegd:
> I am not even sure this can be fixed in PyHornedOWL itself, it looks like an issue with the PyO3 bindings library.

Scratch that, the generation of the docstrings seems to be a custom thing done in PyHornedOWL, _not_ a built-in feature of PyO3. So it _should_ be fixable entirely from within PyHornedOWL.

matentzn:
They said they already did - just waiting for another release: 

https://github.com/ontology-tools/py-horned-owl/issues/58
https://github.com/INCATools/ontology-access-kit/issues/901
https://github.com/INCATools/ontology-access-kit/issues/900

gouttegd:
:facepalm: Note to self: Next time, don’t forget to check out the tip of the main branch, just in case the issue has already been fixed since the last release…

matentzn:
Lol :P I would not do that EVEN if I knew it might be necessary.. But thats why people like you are needed 😬

PR #1356, opened 2026-07-14
Fix update constraints

gouttegd:
This PR fixes a handful of issues with the mechanism to update constraints:

(1) It updates the `constraints.yml` GitHub Action to make sure it uses the same base image as the ODK itself (this should really have been done when the base image was updated).

(2) It updates the `update-constraints.sh` script to make sure that the container in which the constraints are computed has all the tools required to build Python packages from source, in case some of our packages would need to be built from source (whether this is the case or not may vary over time, because it depends on whether all of our dependencies provide pre-compiled wheels arm for Python 3.14 on x86_64 and arm64, something that may change whenever a new version of a dependency is available).

(3) It includes “reverse pins” to force PIP’s dependency solver to use recent versions of `kgx` and `ontobio`. For some reasons, those packages tend to be regularly downgraded, and this can have cascading effect (in particular, as soon as `ontobio` is downgraded to version 2.8.5, this results in a systematic build failure).

### DIFF
--- a/.github/workflows/constraints.yml
+++ b/.github/workflows/constraints.yml
@@ -18,7 +18,7 @@ jobs:
           submodules: recursive
       - name: Build constraints.txt
         run: |
-          docker run --rm -v ${{ github.workspace }}:/workspace -w /workspace ubuntu:24.04 \
+          docker run --rm -v ${{ github.workspace }}:/workspace -w /workspace ubuntu:26.04 \
           sh scripts/update-constraints.sh --in-docker
       - name: Create Pull Request
         uses: peter-evans/create-pull-request@v3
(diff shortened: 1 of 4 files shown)


matentzn (approved):
With the oaklib caveat - which we know about and classified together as an upstream issue we will fix (maybe I am closer then what I thought, there was a new py-horned-owl release that supposedly fixed the HARD problem of 3.14 incompatibility, I am waiting for reports for other incompatibilities coming in).

gouttegd:
> With the oaklib caveat

Yeah, this issue is not about `oaklib` at all. It just fixes some problems that I realized were there while I had been looking at the `oaklib` issue yesterday. :)

PR #1358, opened 2026-07-16
Enforce make constraints.txt to use linux/amd64 platform

matentzn:
The `make constraints.txt` target runs inside an `x86_64` (`amd64`) container, because that is the architecture of the ODK builder image and therefore the one the constraints must reflect.

On native `arm64` some packages (e.g. `py-horned-owl`, a Rust extension) have no pre-built wheel, and pip would try to compile them from source and fail (`error: linker cc not found`) — besides which the resulting constraints would not match the `amd64` builder image.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -157,7 +157,7 @@ publish-multiarch-dev:
 
 # This should use the same base image as the one used to build the ODK itself.
 constraints.txt: python-requirements.txt
-	docker run -v $$PWD:/work -w /work --rm -ti ubuntu:26.04 /work/scripts/update-constraints.sh --in-docker
+	docker run --platform linux/amd64 -v $$PWD:/work -w /work --rm -ti ubuntu:26.04 /work/scripts/update-constraints.sh --in-docker
 
 clean-tests:
 	rm -rf target/*


gouttegd:
> The make constraints.txt target runs inside an x86_64 (amd64) container, because that is the architecture of the ODK builder image and therefore the one the constraints must reflect.

Err, no. Like, nothing is correct in that sentence.

> The `make constraints.txt` target runs inside an x86_64 (amd64) container

No. It runs on whatever architecture the host is using. Run it on a x86_64 platform (like I do when I run it on my personal GNU/Linux box), it runs inside a x86_64 container. Run it on a arm64 platform (like I do when I run on it on my professional Mac box), it runs inside a arm64 container.

> [x86_64] is the architecture of the ODK builder image

No. Same principle as above. When you build the ODK on a x86_64 host, you are building a x86_64 builder image. When you build the ODK on a arm64 host, you are building a arm64 builder image. And when you build the multiarch images (when building for release), you are building a multiarch builder image.

If the builder image was only available for x86_64, the ODK would never have worked on arm64.

> the one the constraints must reflect.

No. The constraint system assumes that, even if some packages might need to be built from source on one architecture and not on the other, the dependencies should be the same.

If you run `update-constraints.txt` on x86_64 and find that a given set of Python packages meet all the dependency requirements, then it is assumed that the same set of Python packages would meet all the dependency requirements on arm64 as well (and the other way around), regardless of whether some of those packages might need to be built from source on arm64.

It should not be an unreasonable assumption for a language that is supposed to be portable.

> On native arm64 some packages (e.g. py-horned-owl, a Rust extension) have no pre-built wheel, and pip would try to compile them from source and fail (error: linker cc not found)

If `pip` fails to build `py-horned-owl` when running `make constraints.txt` under arm64, then it will _also_ fail to build it when building the builder image. _That_ is the real problem that needs fixing here: that `pip` fails to build `py-horned-owl` under arm64.

> `(error: linker cc not found)`

Sigh… Another unhelpful message from a Python tool. Must be a day ending with `y`. The linker _is_ installed in the image that we use to compute the Python constraints. That was precisely one of the point of the PR I made two days ago (#1356).

gouttegd:
> That is the real problem that needs fixing here: that pip fails to build py-horned-owl under arm64.

Problem that I cannot reproduce here, by the way: both `make constraints` and `make builder` work just fine on my arm64 Mac after adding a dependency to `py-horned-owl`.

matentzn:
I will try again.. :P was worth a shot - this way it worked in any case.

gouttegd:
> this way it worked in any case.

I have no doubt that your “fix” would have allowed `make constraints.txt` to work. But it would only have kicked the problem down the road: if the problem is that `py-horned-owl` cannot be built on arm64, computing the constraints on amd64 only _masks_ the problem, which would resurface again the moment you try actually building the ODK.

matentzn:
The problem only happened yesterday - today it was fixed.. Maybe the new py-horned-owl version.

gouttegd:
FWIW, the only way I could reproduce a similar error was by reverting my my local checked out copy to commit https://github.com/INCATools/ontology-development-kit/commit/bbaee1eebc6ce2d5ac915d8827ad8c31ffaf0ed4, _before_ PR #1356 had been merged.

Which is completely expected: before that PR, development tools were _not_ installed in the image that we use to compute constraints, so any attempt to install a Python package that needs to be built from source was bound to fail. That’s precisely one of the issues that PR #1356 had been intended to fix (and did fix).

> The problem only happened yesterday - today it was fixed.. Maybe the new py-horned-owl version.

Or you just updated your master branch between yesterday and today?

matentzn:
> FWIW, the only way I could reproduce a similar error was by reverting my my local checked out copy to commit https://github.com/INCATools/ontology-development-kit/commit/bbaee1eebc6ce2d5ac915d8827ad8c31ffaf0ed4, before PR https://github.com/INCATools/ontology-development-kit/pull/1356 had been merged.

THAT explains it! I just reflexively updated from master upstream into the branch I had already had that change! Detective @gouttegd :P

gouttegd:
Funny, it’s almost as if I had a point when [I said](https://github.com/INCATools/ontology-development-kit/pull/1355#issuecomment-4970085919), just two days ago, that you should always check out the tip of the master/main branch before trying to fix an issue, just in case the issue had already been fixed there…

PR #1362, opened 2026-07-27
(Hopefully) fix the update-constraints workflow.

gouttegd:
The `constraints.yml` workflow has been constantly running into permission errors. Not sure what the problem is exactly, but the `update-epm.yml` workflow is working fine, so in this commit we just align the `constraints.yml` workflow on the `update-epm.yml` workflow.


matentzn (approved):
Fingers crossed

PR #1373, opened 2026-09-10
Fix broken `publish-no-build` target

TankLin23:
> `make publish` cannot push any image at all; this fixes a leftover reference to a Make variable that was removed in 2022.

## Summary

The `publish-no-build` target (used by `make publish`) is currently broken: it still refers to the `DEV` variable, which no longer exists. Because of that, the target fails on its very first command and never pushes anything.

## The problem

The target still contains:

```make
publish-no-build:
	docker push $(DEV):$(VERSION)
	docker push $(IM):latest
	docker push $(IM):$(VERSION)
	$(MAKE) -C docker/odklite publish-no-build
```

`DEV` was removed from the top-level `Makefile` in 94bc1aa7 *"Allow to only push :dev-tagged images to Docker Hub"*, which dropped the then-redundant `obolibrary/odkdev` image. That commit removed the other `$(DEV)` lines but kept this one by mistake.

`DEV` is therefore undefined and expands to the empty string, so the first command run by the target is:

```
$ make -n publish-no-build
docker push :"v1.7"
docker push obolibrary/odkfull:latest
docker push obolibrary/odkfull:"v1.7"
make -C docker/odklite publish-no-build
docker push obolibrary/odklite:"v1.2.30"
docker push obolibrary/odklite:latest
```

(the `"` characters are part of the value of `VERSION` and are removed by the shell; the effective command is `docker push :v1.7`, which is not a valid image reference.)

Docker rejects that first command, and GNU Make stops executing a recipe at the first failing command — so neither `odkfull` nor `odklite` is ever pushed. `make publish` fails before pushing any image.

## The fix

Remove the leftover line. `publish-no-build` then pushes exactly the images that are built, which is what `publish-multiarch` already does:

```
$ make -n publish-no-build
docker push obolibrary/odkfull:latest
docker push obolibrary/odkfull:"v1.7"
make -C docker/odklite publish-no-build
docker push obolibrary/odklite:"v1.2.30"
docker push obolibrary/odklite:latest
```

## Notes

- This only affects the non-multiarch publishing path. `publish-multiarch` and `publish-multiarch-dev` (the targets used for actual releases) never referenced `$(DEV)` and are unaffected.
- I did not build or push any image, as that requires Docker Hub credentials. The change only removes a reference to a variable that is demonstrably undefined, and has been checked with `make -n`.


TankLin23:
Just a gentle nudge — the `Build the ODK images and run the tests` workflow on this PR is still waiting on approval, so it has not run yet. The change is a single-line removal and I have verified the expansion with `make -n publish-no-build`. Happy to adjust anything if it helps.

gouttegd (approved):

PR #1374, opened 2026-09-10
Fix documentation issues

TankLin23:
Three small fixes in the documentation area, one commit each.

## 1. Two pages are built but unreachable from the navigation

`mkdocs.yml` does not list `docs/UsingRobotPlugins.md` and `docs/Synonyms.md` in its `nav` section. Both pages are nevertheless built and listed in the sitemap, so they are served (and indexed by search engines) but cannot be reached by browsing the site:

```
$ mkdocs build --strict
INFO    -  The following pages exist in the docs directory, but are not included in the "nav" configuration:
  - Synonyms.md
  - UsingRobotPlugins.md
```

This matters in particular for `UsingRobotPlugins.md`, which is the only page documenting the `robot_plugins` option of the ODK configuration.

Both pages are added to the navigation. The ROBOT plugins page was never added to the navigation when it was created (292d5ee4); the Synonyms page was part of the navigation until 796a84cd trimmed it down to a single entry, and was never re-added when the file itself was later restored (4bc363b3).

After the change, `mkdocs build --strict` no longer reports anything, and both pages appear in the navigation of the generated site:

```
$ grep -oE 'href="(UsingRobotPlugins|Synonyms)/"' site/index.html
href="UsingRobotPlugins/"
href="Synonyms/"
```

## 2. A dead Sphinx extension is published on the docs site

`docs/exts/markdowntransform.py` is referenced nowhere in the repository; it belongs to the Sphinx/ReadTheDocs documentation setup that was replaced by MkDocs in e07a1b89 (which deleted `docs/conf.py` and `docs/index.rst`). It implements the Sphinx `source-read` event, which is meaningless for MkDocs.

Because MkDocs copies all non-Markdown files from the documentation directory into the generated site, that dead file currently ends up published at:

https://incatools.github.io/ontology-development-kit/exts/markdowntransform.py

It is removed together with the now-empty `docs/exts` directory.

## 3. Wording errors in the ROBOT plugins page

`docs/UsingRobotPlugins.md` contains a wrongly-worded sentence at the top ("allows to use plugins"), a paragraph announcing a KGCL example that actually shows the Uberon one, and a section title missing a word ("Using a plugin a custom workflow").

## Notes

- Tested with `mkdocs build --strict` (MkDocs 1.6.1 / Material for MkDocs 9.7.7): the build is clean and both pages appear in the navigation.
- No file outside the documentation area is touched.


TankLin23:
Just a gentle nudge on this one. It is documentation-only — all three files fall under the existing `paths-ignore` patterns, so no image build is triggered. Happy to adjust anything if it helps.

gouttegd (approved):
Thanks!

PR #1376, opened 2026-09-12
Force `pip` to accept installing packages in the global environment.

gouttegd:
Recent versions of `pip` in Ubuntu categorically refuses to install anything in the global environment (be it the system-wide environment or the user-specific environment) unless it is called with `--break-system-packages`.

This guardrail is completely useless in the ODK, where the "global environment" is already a Docker container completely isolated from the rest of the system.

So we force `pip` to always behave as if `--break-system-packages` was always active, so that users who need to install their own Python packages don't have to remember to use that option.


matentzn (approved):
Yes!

PR #1377, opened 2026-09-13
Add linkml-owl to python requirements

matentzn:
`linkml-owl` is needed so LinkML instances can be translated into OWL, see also https://linkml.io/linkml-owl/tutorial/.


gouttegd (approved):

gouttegd:
Though upon seeing [this](https://linkml.io/linkml-owl/comparison/):

> Although LinkML is robust and stable, LinkML-OWL is alpha software and incomplete. For now, to convert from TSV to OWL you should for now use a dedicated environment: 
>
> * dosdp-tools
> * robot-templates
> * ottr

Given that both DOSDP-Tools and ROBOT templates **are** usable out-of-the-box with the ODK, I wonder if it’s really a good idea to provide in the ODK a tool whose own developers recommend against using… Seems to me like the kind of tools about which I’d say, “you want it, you install it”…

matentzn:
> LinkML-OWL is alpha software and incomplete.

I am working on a reconciling a series of rogue repos which started to pick LinkML as the curation lawyer for ontologies. What I ultimately want to achieve is this (odk 1.8) - a way to keep the curation layer (mondo-edit.yaml) in LinkML land, while still being able to leverage the entire standardisation layer (release artefacts / qc). A lot of this is not thought through, like how the import layer should work is entirely a massive question mark, but I feel this is an important task for me for mid 2027. I am already dealing with some repos that are using a linkml source - but they are entirely auto generated from some database so they do not need the ODK repo layout, just the toolbox.

gouttegd:
OK, thanks for the background info.

Just one thing to consider (this is not an objection: if you really do want `linkml-owl` in ODKFull, go ahead :) ): We are making at most two ODK releases per year (one major release and possibly one minor release). If `linkml-owl` is still in alpha and thus presumably under development, are you sure you want to be stuck with a frozen version for several months? And having to ask for a new minor ODK release whenever a new development version of `linkml-owl` is available?

I believe this is exactly the kind of things for which you would be better off installing `linkml-owl` yourself (possibly using https://github.com/INCATools/odkcore/pull/30) so that you can easily use the latest version whenever you need – at least until `linkml-owl` is deemed stable enough.

Again, not a strong objection. If you thought this through and really want `linkml-owl`, it’s your call.

gouttegd:
Aside:

> a way to keep the curation layer (mondo-edit.yaml) in LinkML land, while still being able to leverage the entire standardisation layer (release artefacts / qc)

Sounds like my LinkML-Java project could at some point include a ROBOT plugin for easy LinkML-to-OWL injection, similarly to the `sssom:inject` command for easy SSSOM-to-OWL injection…

matentzn:
Thanks for giving me the power of choice. My decision to merge this is heavily influenced by two factors:

1. I would like to avoid to having to run a pip install every time i run the pipeline (especially now that this is under development, where we run this many times an hour)
2. Its not THAT Alpha - it works ok and it is used in some production systems like [CHEMROF](https://github.com/chemkg/chemrof/blob/main/src/chemrof/converter/owl_output.py) - So I would argue its more like beta now. :P Good enough for production work.

> Sounds like my LinkML-Java project could at some point include a ROBOT plugin for easy LinkML-to-OWL injection, similarly to the sssom:inject command for easy SSSOM-to-OWL injection…

Dont burn out :P But thanks for the offer :) 2027!

gouttegd:
> I would like to avoid to having to run a pip install every time i run the pipeline (especially now that this is under development, where we run this many times an hour)

Then you still have the option of building your own custom image. All it takes is the simplest Dockerfile:

```Dockerfile
FROM obolibrary/odkfull:latest
RUN python3 -m pip install linkml-owl
```

I’ve said many times that building a custom image was by far the best option for anyone with custom needs. I still believe it is.

matentzn:
> I’ve said many times that building a custom image was by far the best option for anyone with custom needs. I still believe it is.

This is true, but a maintenance layer; Since I am not the only person working on this process (there are three already, and more soon, including @ptgolden and @Reasat) I dont really want to add the necessity to build a new docker image every time I run a github action, or host an extended docker image just for adding one library. We have 6 or so repos which need this already.

I see your point though. If I see the scope of ODK burgeoning with this effort, I will fork and host another docker image.

gouttegd:
> I dont really want to add […] host an extended docker image just for adding one library

Yeah, you find it more convenient to clutter ODKFull with stuff that is only useful for you, one or two other people, and a grand total of 6 repos.

Be advised that I might decide to put a stop to that at some point.

> If I see the scope of ODK burgeoning with this effort, I will fork and host another docker image.

Why forking? You can build your own custom image on top of `obolibrary/odkfull` without having to fork.

And forking could be a terribly bad idea, because you could easily end up publishing your own image under the canonical `obolibrary/odkfull` name.

matentzn:
I didn't mean forking I meant extending; of course I would not fork! 😬

gouttegd:
Makes me wonder if we should consider having a `ODKExp` image (“Experimental”) on top of ODKFull. With much relaxed rules about what can go there, but also no guarantees that whatever is in that image will remain forever.

gouttegd:
> Yeah, you find it more convenient to clutter ODKFull with stuff that is only useful for you, one or two other people, and a grand total of 6 repos.

Sorry, this comment was totally uncalled for.

matentzn:
> Sorry, this comment was totally uncalled for.

I work with you for many years man I only hear the professional concerns underneath the prose - same I do with everyone else! No worries from me 🙏

PR #1381, opened 2026-09-16
Fix the broken link to the OBO Academy synonyms documentation

TankLin23:
`docs/Synonyms.md` is a one-line pointer to the OBO Academy book, but the page it
points to no longer exists:

```
$ curl -s -o /dev/null -w '%{http_code}\n' -L https://oboacademy.github.io/obook/reference/synonyms-obo/
404
```

OBO Academy has reorganised its content; the synonyms material now lives in the
`lesson/` section, under a page whose title is still *"An Introduction to
Synonyms in OBO Ontologies"*:

```
$ curl -s -o /dev/null -w '%{http_code}\n' -L https://oboacademy.github.io/obook/lesson/synonyms/
200
```

This updates the stub to point to that page.

Documentation-only change; no build output is affected.


gouttegd (approved):
The entire ODK documentation will need to be overhauled at some point, but in the meantime, there’s no reason to keep broken links. Thanks!

